PR #1, opened 2019-07-22
Initial commit for plugin

vthapar:
Adds skeletal plugin
README on how to compile


go.mod
```diff
@@ -3,9 +3,46 @@ module github.com/submariner-io/coastguard
...
+	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
+	go.etcd.io/bbolt v1.3.3 // indirect
+	go.uber.org/atomic v1.4.0 // indirect
+	go.uber.org/multierr v1.1.0 // indirect
+	go.uber.org/zap v1.10.0 // indirect
+	google.golang.org/grpc v1.22.0 // indirect
+	gopkg.in/jcmturner/goidentity.v3 v3.0.0 // indirect
 )
 
 replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190223031456-9a5ae4453bd
```
tpantelis: Is this needed, ie lighthouse doesn't use strongSwan?

go.mod
```diff
@@ -3,9 +3,46 @@ module github.com/submariner-io/coastguard
 go 1.12
```
tpantelis: The module above should reference lighthouse.

plugin/lighthouse.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Use a non-writer so we don't write NXDOMAIN to client
+	nw := nonwriter.New(w)
+
+	// Call next plugin (if any).
+	return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, nw, r)
+}
+
+// Name implements the Handler interface.
+func (lh Lighthouse) Name() string { return "lighthouse" }
```
tpantelis: The return and end brace should be on separate lines.

plugin/setup.go
```diff
@@ -0,0 +1,37 @@
...
+	if c.NextArg() {
+		// If there was another token, return an error, because we currently don't have any configuration.
+		// Any errors returned from this setup function should be wrapped with plugin.Error, so we
+		// can present a slightly nicer error message to the user.
+		return plugin.Error("lighthouse", c.ArgErr())
+	}
+
+	// Add the Plugin to CoreDNS, so Servers can use it in their plugin chain.
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		return Lighthouse{Next: next}
```
tpantelis: I think you want to return `& Lighthouse(...)`

go.mod
mangelajo: True, this can be removed.

plugin/README.md
```diff
@@ -0,0 +1,51 @@
+# lighthouse
+
+## Name
+
+*Lighthouse* - serve services across clusters.
```
mangelajo: serve services? :-)
 
what about DNS discovery for services across clusters?

plugin/README.md
```diff
@@ -0,0 +1,51 @@
...
+kubernetes:kubernetes
+lighthouse:github.com/submariner-io/lighthouse/plugin/lighthouse
+~~~
+* `make -f Makefile.release DOCKER=your-docker-repo release`
+* `make -f Makefile.release DOCKER=your-docker-repo docker`
+* `make -f Makefile.release DOCKER=your-docker-repo docker-push`
+
+## Examples
+
+~~~ txt
```
mangelajo: use "`" instead of ~    ?

plugin/README.md
```diff
@@ -0,0 +1,51 @@
...
+. {
+    errors
+    log
+    kubernetes cluster.local {
+      fallthrough
+    }
+    lighthouse
+}
+~~~
```
mangelajo: same here

plugin/setup_test.go
```diff
@@ -0,0 +1,21 @@
+package lighthouse
```
mangelajo: you need to adapt to use ginkgo/gomega.

plugin/setup.go
vthapar: Used this as reference:
https://github.com/coredns/example/blob/master/setup.go#L41

plugin/README.md
vthapar: Changed to discover, wording was from template README I picked up.

plugin/README.md
vthapar: It will become a multiline block as we figure out and add more options.

go.mod
vthapar: done

plugin/lighthouse.go
vthapar: Done

plugin/README.md
vthapar: Done

plugin/setup_test.go
vthapar: Will do in subsequent patches. Still figuring out how to write tests with ginkgo

plugin/setup.go
mangelajo: Yes but we agreed to use ginkgo, which is just a different framework. I believe you can map the same thing to ginkgo.

vthapar: This reply looks out of place here. did you mean this for the setup_test.go? If yes, I'll be making the switch to ginkgo in subsequent patch, if okay with you.

PR #5, opened 2019-07-25
Added the resource files for the lighthouse CRD. The CRD will hold

aswinsuryan:
the information about the services created in all the clusters.


mangelajo:
Please check the golint errors: https://travis-ci.com/submariner-io/lighthouse/builds/120624720#L1402 (seems more like compilation error)

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
+package v1
+
+import (
+	"net"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
tpantelis: These special comments are directives for the k8s code generator. Do you plan to generate all the Clientset code as well?

aswinsuryan: Yes.  I think that will be required to use this CRD, is that correct? Also one of the file zz_generated.deepcopy.go  is auto-generated.

tpantelis: It's not required but the generated code makes it much easier to use. You can use [this script wrapper](https://github.com/submariner-io/submariner/blob/master/scripts/codegen) to generate the code.

mangelajo:
@tpantelis do you think it's ok to merge ?

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              MultiClusterServiceSpec `json:"spec"`
+}
+
+//MultiClusterServiceSpec provides the specification of a MultiClusterService
+type MultiClusterServiceSpec struct {
+	//The name of the service. When a DNS request is received this field is used to
+	//retrieve the information about the clusters in which the service is present.
+	//This field is not optional.
+	ServiceName string `json:"service_name"`
```
tpantelis: How is this different than the standard k8s metadata `Name`?

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+
+	//The DNS zone in which the service is a part of. This field is not optional.
+	ClusterDomain string `json:"cluster_domain"`
+
+	//The cluster IP of the service running in the cluster that is identified by the
+	//ClusterID field. This field is not optional.
+	ServiceIP net.IP `json:"service_ip"`
+
+	// +optional
+	//The port in which the service listens at.This field is an optional field.
```
tpantelis: space after period, also the term _field_ is mentioned twice: _This is an optional field._

tpantelis:
> @tpantelis do you think it's ok to merge ?

Just a couple comments. Also what about generating the clientset code? Will that be in a separate patch?

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: I will push the generated client set code as a separate patch.

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: will push these changes soon in another patch

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: Added the client generated code.

pkg/apis/multiclusterservice/v1/types.go
tpantelis: > How is this different than the standard k8s metadata `Name`?

bump...

aswinsuryan: Not really clear about this comment. This will act as the key for the CRD and it is the Kubernetes service name.

tpantelis: `metav1.ObjectMeta` has a `Name` field. Do you need a separate `ServiceName` field here or just use `Name` for that purpose? If the former, how would `Name` and `ServiceName` differ?

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	//List of service information across clusters. This field is not optional. Each
+	//service name should have at least one service info associated with it.
+	Items []MultiClusterServiceInfo `json:"items"`
+}
+
+//MultiClusterServiceInfo provides information about service in each cluster.
+type MultiClusterServiceInfo struct {
+	//The unique identity of the cluster where the service is available. This field is
+	//not optional.
+	ClusterID string `json:"cluster_id"`
```
tpantelis: It seems the convention in k8s is to use camel case for JSON field names and not use underscores, ie `cluster_id` -> `clusterID` etc.  (Sorry for not commenting on this earlier - I just noticed it)

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+//MultiClusterServiceSpec provides the specification of a MultiClusterService
+type MultiClusterServiceSpec struct {
+	//The name of the service. When a DNS request is received this field is used to
+	//retrieve the information about the clusters in which the service is present.
+	//This field is not optional.
+	ServiceName string `json:"service_name"`
+
+	//List of service information across clusters. This field is not optional. Each
+	//service name should have at least one service info associated with it.
+	Items []MultiClusterServiceInfo `json:"items"`
```
tpantelis: The name `Items` seems a little vague to me - how about `clusterServiceInfo`?

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: Ya, name in ObjectMeta will do the purpose, removed the service name.

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: Changed

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: Done

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,57 @@
...
+	//List of service information across clusters. This field is not optional. Each
+	//service name should have at least one service info associated with it.
+	Items []ClusterServiceInfo `json:"items"`
+}
+
+//ClusterServiceInfo provides information about service in each cluster.
+type ClusterServiceInfo struct {
+	//The unique identity of the cluster where the service is available. This field is
+	//not optional.
+	ClusterID string `json:"clusterId"`
```
tpantelis: Not trying to nit but the recommended convention for naming "ids" is uppercase, ie `clusterID` ...

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,57 @@
...
+
+	//The DNS zone in which the service is a part of. This field is not optional.
+	ClusterDomain string `json:"clusterDomain"`
+
+	//The cluster IP of the service running in the cluster that is identified by the
+	//ClusterID field. This field is not optional.
+	ServiceIP net.IP `json:"serviceIp"`
```
tpantelis: same with "ip", ie `serviceIP`.

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,57 @@
...
+type MultiClusterService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              MultiClusterServiceSpec `json:"spec"`
+}
+
+//MultiClusterServiceSpec provides the specification of a MultiClusterService
+type MultiClusterServiceSpec struct {
+	//List of service information across clusters. This field is not optional. Each
+	//service name should have at least one service info associated with it.
```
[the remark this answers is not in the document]
aswinsuryan: I mean to say it is not optional, and since we have removed the service name I feel this is comment is redundant and hence removing it.

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,57 @@
...
+	Items []ClusterServiceInfo `json:"items"`
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
...
+type MultiClusterService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              MultiClusterServiceSpec `json:"spec"`
+}
+
+//MultiClusterServiceSpec provides the specification of a MultiClusterService
+type MultiClusterServiceSpec struct {
+	//List of service information across clusters. This field is not optional.
+	Items []ClusterServiceInfo `json:"clusterServiceInfo"`
```
tpantelis: I also meant to name the struct field `ClusterServiceInfo` but I guess `Items` is OK, at least the external JSON representation indicates _clusterServiceInfo_.

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+//+genclient
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+//MultiClusterService describes the information of service that is available across
```
mpeterson: Put spaces at the beginning of the comments

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
...
+
+	//The DNS zone in which the service is a part of. This field is not optional.
+	ClusterDomain string `json:"clusterDomain"`
+
+	//The cluster IP of the service running in the cluster that is identified by the
+	//ClusterID field. This field is not optional.
+	ServiceIP net.IP `json:"serviceIP"`
+
+	//+optional
+	//The port in which the service listens at.This is an optional field.
```
mpeterson: s/t.T/t. T/

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+//+genclient
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+//MultiClusterService describes the information of service that is available across
+//clusters. A shared informer running in the lighthouse control plane will listen for
+//service creation in every cluster that is part of the control plane and whenever a
+//service is created a MultiClusterService will be created. Lighthouse will
+//distribute the MultiClusterService to every other cluster that is the part of the
```
mpeterson: you'll distribute a MultiClusterService or a MultiClusterServiceList CRD?

pkg/apis/multiclusterservice/v1/types.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+//MultiClusterServiceList is a list of multiclusterservice objects.
+type MultiClusterServiceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []MultiClusterService `json:"items"`
+}
```
mpeterson: General comment:

This CRD is too verbose. Can't you already have a list of a dictionary where key is the Domain and the value is the IP/PORT?

Anyways the controller is the one that takes those decisions, there is no need to provide such a structured CRD IMHO

aswinsuryan: I was under the impression this is the general practice as I modeled  it based on the other CRD's. Doesn't it make it easier to test create /delete the CRD's using kubectl?

pkg/apis/multiclusterservice/v1/types.go
aswinsuryan: If there are existing services when a cluster joins , I think it will be easier to distribute MultiClusterServiceList. In the case of new service MultiClusterService would make more sense. But this needs to be tested to confirm it can work in this way.

pkg/apis/multiclusterservice/v1/types.go
mpeterson: I'm not sure I understand the advantage you say this gives. Could you explain me with an example or differently?

pkg/apis/multiclusterservice/v1/types.go
mpeterson: I think it's best to have less CRD objects in the instance and for sure not a mix of them. It makes the controllers job much harder.

PR #8, opened 2019-08-05
Fix broken CoreDNS compile

vthapar:


vthapar:
Build is failing coz unable to find the new package that is being added as part of this patch. 

https://travis-ci.com/submariner-io/lighthouse/builds/121924665

Looking for packages to test
1258Running tests in . ./plugin/...
1259Failed to compile lighthouse:
1260
1261can't load package: package github.com/submariner-io/lighthouse/plugin/lighthouse: cannot find package "." in:
1262	/go/src/github.com/submariner-io/submariner/vendor/github.com/submariner-io/lighthouse/plugin/lighthouse

I get the same error locally too. Would this be an expected failure till this patch gets in?

PR #9, opened 2019-08-07
Lighthouse ci

dimaunx:
Lighthouse ci proposal. Currently includes coredns with external lighthouse plugin build and deployment.


Dockerfile.dapper
```diff
@@ -1,51 +1,61 @@
...
-# helm           | e2e tests
-# kubefedctl     | e2e tests
-# kind           | e2e tests
-# ginkgo         | tests
-# goimports      | code formatting
+# gcc             | ginkgo
+# git             | find the workspace root
+# curl            | download other tools
+# docker.io       | Dapper
+# build-essential | coredns make
```
skitt: Do we need `build-essential` here, or only `make`? `build-essential` pulls in `g++` and `dpkg-dev` along with `make`, which adds quite a lot to the container.

dimaunx: Good point.

dimaunx: Added just make and tested.

vthapar:
Tested and Verified!

scripts/build-coredns
```diff
@@ -0,0 +1,28 @@
...
+DOCKER_TAG="lighthouse-coredns:dev"
+BUILD_DIR="build/linux/amd64"
+
+cd ${TOPDIR}
+git clone -b v${COREDNS_VERSION} ${COREDNS_GIT}
+cd ${TOPDIR}/coredns
+sed -i "\$alighthouse:github.com/submariner-io/lighthouse/plugin/lighthouse" plugin.cfg
+go mod vendor
+mkdir -p ${TOPDIR}/coredns/vendor/github.com/submariner-io/lighthouse/plugin/ ${BUILD_DIR}
+cp -R ${TOPDIR}/plugin ${TOPDIR}/coredns/vendor/github.com/submariner-io/lighthouse/
```
mangelajo: (not sure), do you need -f too to make sure it overwrites existing old files? 

another option would be a link?

scripts/kind-e2e/config/cluster1-config.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Cluster
```
mangelajo: Wouldn't it be less maintenance burden if you clone submariner somewhere and use those files from there?, you'd need to maintain only once.

scripts/kind-e2e/config/elasticsearch.yaml
```diff
@@ -0,0 +1,139 @@
+---
```
mangelajo: same

scripts/kind-e2e/config/filebeat.yaml
```diff
@@ -0,0 +1,159 @@
+---
```
mangelajo: same ':-D

scripts/kind-e2e/config/netshoot.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
mangelajo: ok, I won't repeat myself more ':-D

scripts/kind-e2e/e2e.sh
```diff
@@ -0,0 +1,319 @@
+#!/usr/bin/env bash
```
mangelajo: This one to some extent, if we make the one in subm extensible somehow.

scripts/build-coredns
mangelajo: cp -R seems to overwrite,

but I would delete the directory, and then copy again, some files could have been deleted in the new sources of lighthouse plugin, and .. if present make compilation fail.

A link would also solve it (but not sure if that's container-happy)

dimaunx: The coredns folder and its vendor will be removed each run , there a trap exit that deleted the folder, this ensures every run is a new code.

scripts/kind-e2e/config/cluster1-config.yaml
dimaunx: Yes you are correct, we already spoke about internally. The idea is to create a tool based on kind that will just accept parameters and all the projects will use it. This will also remove the dependency on dapper. Some suggestions for names are armada and [futtock](https://www.google.com/search?q=futtock&oq=futtock&aqs=chrome..69i57j69i61.1933j0j7&sourceid=chrome&ie=UTF-8)  
:)

dimaunx: The problem is that if we make changes to submariner e2e and break its e2e we are braking all the projects together.

scripts/build-coredns
mangelajo: Oh righty!, I missed that part thanks :)

scripts/kind-e2e/config/cluster1-config.yaml
mangelajo: Well, but shouldn't CI prevent such breakage in first instance? :) I'd be fine with that.

mangelajo: Why another tool?, 

Remember that we need to maintain those tools, and compile for other platforms, for example, I use dapper in OSX....

dimaunx: Yes you are right but there could be always an edge case, its lame argument :) i know, another point is that each project has its own custom flow, so pulling all submariner stuff is not needed and each project can have its own upgrade path and validate and test logic.

skitt: The issue is that we’re sharing a bunch of code, which means we need to revisit the situation (and one possible conclusion is that it’s fine where it is, but we do need to think about it). If we pull the e2e code out as a separate project, then we’d have project management on it, _e.g._ releases, so other projects could depend on a version they know works, and we wouldn’t force unintentional breakage on them...

dimaunx: Yes to have  a smart provisioner in separate repository and all the projects could utilize it is seems the way to go.

PR #11, opened 2019-08-14
Dns handler

vthapar:
Basic DNS handler that responds on service name requests of typeA with dns record.

### DIFF
--- a/plugin/lighthouse/handler.go
+++ b/plugin/lighthouse/handler.go
@@ -1,0 +1,62 @@
+package lighthouse
+
+import (
+	"context"
+	"errors"
+	"net"
+	"strings"
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+)
+
+// ServeDNS implements the plugin.Handler interface.
+func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+
+	a := new(dns.Msg)
+	a.SetReply(r)
+	a.Authoritative = true
+
+	services := lh.SvcsMap
+	svcName := strings.Split(state.QName(), ".")[0]
+	ip := services[svcName]
+	if ip == "" {
+		// We can't handle this,let another plugin take an attempt
+		// NOTE: Once we have options enabled, this will only be done if
+		//       fallthrough is enabled.
+		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+	}
+	rr := new(dns.A)
+
+	if state.Family() == 1 {
+		// IPv4 query
+		rr.Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
+		rr.A = net.ParseIP(ip).To4()
+	} else {
+		// We don't support IPv6, let another plugin take an attempt
+		log.Debugf("IPv6 queries not supported yet")
+		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+	}
+
+	a.Answer = []dns.RR{rr}
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	wErr := w.WriteMsg(a)
+	if wErr != nil {
+		log.Errorf("Failed to write message %#v: %v", a, wErr)
+		return dns.RcodeServerFailure, lh.Error("failed to write response")
+	}
+
+	return dns.RcodeSuccess, nil
+}
+
+func (lh *Lighthouse) Error(str string) error {
+	return plugin.Error(lh.Name(), errors.New(str))
+}
+
+// Name implements the Handler interface.
+func (lh *Lighthouse) Name() string {
+	return "lighthouse"
+}
--- a/plugin/lighthouse/handler_test.go
+++ b/plugin/lighthouse/handler_test.go
@@ -1,0 +1,92 @@
+package lighthouse
+
+import (
+	"context"
+	"os"
+
+	"github.com/coredns/coredns/plugin/pkg/dnstest"
+	"github.com/coredns/coredns/plugin/test"
+	"github.com/miekg/dns"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("[setup] Test basic DNS Handler", func() {
+	Describe("function serveDNS", testDnsServer)
+})
+
+func testDnsServer() {
+	var (
+		lh  *Lighthouse
+		ctx context.Context
+		tc  test.Case
+	)
+	tests := []test.Case{
+		{
+			// service is present.
+			Qname: "mysvc.default.svc.cluster.local.", Qtype: dns.TypeA,
+			Rcode: dns.RcodeSuccess,
+			Answer: []dns.RR{
+				test.A("mysvc.default.svc.cluster.local.    0    IN    A    100.96.156.175"),
+			},
+		},
+		{
+			// service does not exist, expect error.
+			Qname: "mysvc2.default.svc.cluster.local.", Qtype: dns.TypeA,
+			Rcode: dns.RcodeServerFailure,
+		},
+		{
+			// service does not exist, expect error.
+			Qname: "mysvc2.default.svc.cluster.local.", Qtype: dns.TypeAAAA,
+			Rcode: dns.RcodeServerFailure,
+		},
+	}
+
+	os.Setenv("LIGHTHOUSE_SVCS", "mysvc=100.96.156.175,dummy=100.96.156.175,none=100.96.156.100")
+	lh = &Lighthouse{}
+	lh.SvcsMap = setupServicesMap()
+
+	ctx = context.TODO()
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[0]
+		})
+		It("Should return A record for existent query", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+			respErr := test.SortAndCheck(rec.Msg, tc)
+			Expect(respErr).NotTo(HaveOccurred())
+		})
+	})
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[1]
+		})
+		It("Should return an error for non existent query", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+
+	When("DNS query for typeAAAA record", func() {
+		BeforeEach(func() {
+			tc = tests[2]
+		})
+		It("Should return an error", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+}
--- a/plugin/lighthouse/lighthouse.go
+++ b/plugin/lighthouse/lighthouse.go
@@ -1,38 +1,16 @@
 package lighthouse
 
 import (
-	"context"
-
 	"github.com/coredns/coredns/plugin"
 	clog "github.com/coredns/coredns/plugin/pkg/log"
-	"github.com/coredns/coredns/plugin/pkg/nonwriter"
-	"github.com/coredns/coredns/request"
-
-	"github.com/miekg/dns"
 )
 
 // Define log to be a logger with the plugin name in it. This way we can just use log.Info and
 // friends to log.
 var log = clog.NewWithPlugin("lighthouse")
 
+type ServicesMap map[string]string
 type Lighthouse struct {
-	Next plugin.Handler
-}
-
-// ServeDNS implements the plugin.Handler interface.
-func (lh Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
-	state := request.Request{W: w, Req: r}
-	// Debug log that we've have seen the query. This will only be shown when the debug plugin is loaded.
-	log.Debugf("Received query from '%s'", state.Name())
-
-	// Use a non-writer so we don't write NXDOMAIN to client
-	nw := nonwriter.New(w)
-
-	// Call next plugin (if any).
-	return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, nw, r)
-}
-
-// Name implements the Handler interface.
-func (lh Lighthouse) Name() string {
-	return "lighthouse"
+	Next    plugin.Handler
+	SvcsMap ServicesMap
 }
--- a/plugin/lighthouse/setup.go
+++ b/plugin/lighthouse/setup.go
@@ -1,9 +1,13 @@
 package lighthouse
 
 import (
-	"github.com/caddyserver/caddy"
+	"os"
+	"strings"
+
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
+
+	"github.com/caddyserver/caddy"
 )
 
 // init registers this plugin within the Caddy plugin framework. It uses "example" as the
@@ -26,11 +30,29 @@ func setupLighthouse(c *caddy.Controller) error {
 		return plugin.Error("lighthouse", c.ArgErr())
 	}
 
+	svcsMap := setupServicesMap()
+	log.Debugf("Services Map for lighthouse is %s", svcsMap)
 	// Add the Plugin to CoreDNS, so Servers can use it in their plugin chain.
 	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
-		return Lighthouse{Next: next}
+		return &Lighthouse{Next: next, SvcsMap: svcsMap}
 	})
 
 	// All OK, return a nil error.
 	return nil
+}
+
+func setupServicesMap() ServicesMap {
+	svcs := make(ServicesMap)
+	svcArray := os.Getenv("LIGHTHOUSE_SVCS")
+	if svcArray == "" {
+		return svcs
+	}
+	strArray := strings.Split(svcArray, ",")
+	for _, str := range strArray {
+		svc := strings.Split(str, "=")
+		if len(svc) == 2 {
+			svcs[svc[0]] = svc[1]
+		}
+	}
+	return svcs
 }
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,8 @@ require (
 	github.com/onsi/ginkgo v1.8.0
 	github.com/onsi/gomega v1.5.0
 	github.com/submariner-io/submariner v0.0.0-20190708095718-350482d85dd4
+	k8s.io/apimachinery v0.0.0-20190629003722-e20a3a656cff
+	k8s.io/client-go v11.0.0+incompatible
 )
 
 replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190223031456-9a5ae4453bd
--- a/go.sum
+++ b/go.sum
@@ -38,6 +38,7 @@ github.com/eapache/go-resiliency v1.1.0/go.mod h1:kFI+JgMyC7bLPUVY133qvEBtVayf5m
 github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21/go.mod h1:+020luEh2TKB4/GOp8oxxtq0Daoen/Cii55CzbTV6DU=
 github.com/eapache/queue v1.1.0/go.mod h1:6eCeP0CKFpHLu8blIFXhExK/dRa7WDZfr6jVFPTqq+I=
 github.com/evanphx/json-patch v0.0.0-20180908160633-36442dbdb585/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/evanphx/json-patch v4.1.0+incompatible h1:K1MDoo4AZ4wU0GIU/fPmtZg7VpzLjCxu+UwBD1FvwOc=
 github.com/evanphx/json-patch v4.1.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
 github.com/farsightsec/golang-framestream v0.0.0-20181102145529-8a0cb8ba8710/go.mod h1:eNde4IQyEiA5br02AouhEHCu3p3UzrCdFR4LuQHklMI=
 github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568 h1:BHsljHzVlRcyQhjrss6TZTdY2VfCqZPbv5k3iBFa2ZQ=
@@ -59,6 +60,7 @@ github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zV
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/groupcache v0.0.0-20160516000752-02826c3e7903/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
+github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef h1:veQD95Isof8w9/WXiA+pa3tz3fJXkt5B7QaRBrM62gk=
 github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
 github.com/golang/mock v1.1.1/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfbm0A=
 github.com/golang/mock v1.2.0/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfbm0A=
@@ -100,6 +102,7 @@ github.com/hashicorp/go-syslog v1.0.0/go.mod h1:qPfqrKkXGihmCqbJM2mZgkZGvKG1dFdv
 github.com/hashicorp/golang-lru v0.0.0-20160207214719-a0d98a5f2880/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hashicorp/golang-lru v0.0.0-20180201235237-0fb14efe8c47/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hashicorp/golang-lru v0.5.0/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
+github.com/hashicorp/golang-lru v0.5.1 h1:0hERBMJE1eitiLkihrMvRVBYAkpHzc/J3QdDN+dAcgU=
 github.com/hashicorp/golang-lru v0.5.1/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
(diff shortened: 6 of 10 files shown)


plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+		rr.(*dns.A).Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
+		rr.(*dns.A).A = net.ParseIP(ip).To4()
+	default:
+		return onFailure(lh, ctx, w, r)
+	}
+
+	a.Answer = []dns.RR{rr}
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	if w.WriteMsg(a) != nil {
```
tpantelis: Same here re: logging...

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+		return onFailure(lh, ctx, w, r)
+	}
+	var rr dns.RR
+
+	switch state.Family() {
+	case 1:
+		rr = new(dns.A)
+		rr.(*dns.A).Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
+		rr.(*dns.A).A = net.ParseIP(ip).To4()
+	default:
```
tpantelis: I think it would be good to at least log here, at least Debug but maybe Warning or Error if we expect this case to be infrequent.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/coredns/coredns/plugin/pkg/nonwriter"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+// ServeDNS implements the plugin.Handler interface.
+func (lh Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+	state := request.Request{W: w, Req: r}
+	// Debug log that we've have seen the query. This will only be shown when the debug plugin is loaded.
```
tpantelis: Is this comment misplaced or did you intend to add logging here?

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	if ip == "" {
+		return onFailure(lh, ctx, w, r)
+	}
+	var rr dns.RR
+
+	switch state.Family() {
+	case 1:
+		rr = new(dns.A)
+		rr.(*dns.A).Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
+		rr.(*dns.A).A = net.ParseIP(ip).To4()
```
tpantelis: Instead of casting to `dns.A` on each line, I think it would be cleaner to use a temp var of type `dns.A` and then assign to `rr`.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+			tc = tests[2]
+		})
+		It("Should return error for AAAA", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
```
tpantelis: Sorry - I'm not familiar with _typeA_ vs _typeAAAA_ but there's no test for when a  _typeAAAA_ record exists. Is this by design?

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/coredns/coredns/plugin"
+	"github.com/coredns/coredns/plugin/pkg/nonwriter"
+	"github.com/coredns/coredns/request"
+
+	"github.com/miekg/dns"
+)
+
+// ServeDNS implements the plugin.Handler interface.
+func (lh Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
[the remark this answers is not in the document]
vthapar: Took this code as is from reference plugin. All in-tree plugins also use same naming convention so went with it.

Changes done.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	state := request.Request{W: w, Req: r}
+	// Debug log that we've have seen the query. This will only be shown when the debug plugin is loaded.
+
+	a := new(dns.Msg)
```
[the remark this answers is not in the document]
vthapar: Same as above.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	if w.WriteMsg(a) != nil {
+		return onFailure(lh, ctx, w, r)
+	}
+
+	return dns.RcodeSuccess, nil
+}
+
+func onFailure(lh Lighthouse, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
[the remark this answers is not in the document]
vthapar: Still can't think in Go very well, so went with what I am used to. Advantages of using receiver over arg? In this particular case it is just a wrapper return method which can be called from multiple places within this file.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	services := lh.SvcsMap
+	svcName := strings.Split(state.QName(), ".")[0]
+	ip := services[svcName]
+	if ip == "" {
+		return onFailure(lh, ctx, w, r)
+	}
+	var rr dns.RR
+
+	switch state.Family() {
+	case 1:
```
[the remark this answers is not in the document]
vthapar: Done.

plugin/lighthouse/handler.go
vthapar: Misplaced. This was from before I had DNS handler and debug log was to check if plugin loaded correctly or not.

plugin/lighthouse/handler_test.go
vthapar: A is ipv4 and AAAA is ipv6. There is test for AAAA. I wanted to figure out how to define testscases in one place and use a loop to run them all. tests[2] is the AAAA test cases.

plugin/lighthouse/setup.go
```diff
@@ -26,11 +30,28 @@ func setupLighthouse(c *caddy.Controller) error {
 		return plugin.Error("lighthouse", c.ArgErr())
 	}
 
+	svcsMap := setupServicesMap()
+	log.Debugf("Services Map for lighthouse is %s", svcsMap)
 	// Add the Plugin to CoreDNS, so Servers can use it in their plugin chain.
 	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
-		return Lighthouse{Next: next}
+		return Lighthouse{Next: next, SvcsMap: svcsMap}
```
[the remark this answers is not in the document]
vthapar: Done.

plugin/lighthouse/setup.go
```diff
@@ -26,11 +30,28 @@ func setupLighthouse(c *caddy.Controller) error {
...
+	var strArray []string
+	svcs := make(map[string]string)
+	svcArray := os.Getenv("LIGHTHOUSE_SVCS")
+	if svcArray == "" {
+		return svcs
+	}
+	strArray = strings.Split(svcArray, ",")
+	for _, str := range strArray {
+		svc := strings.Split(str, "=")
+		svcs[svc[0]] = svc[1]
```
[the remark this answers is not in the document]
vthapar: Done.

plugin/lighthouse/setup.go
```diff
@@ -26,11 +30,28 @@ func setupLighthouse(c *caddy.Controller) error {
...
-		return Lighthouse{Next: next}
+		return Lighthouse{Next: next, SvcsMap: svcsMap}
 	})
 
 	// All OK, return a nil error.
 	return nil
 }
+
+func setupServicesMap() map[string]string {
+	var strArray []string
```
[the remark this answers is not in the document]
vthapar: LIne 47 is svcArray. Changed Line 51 to use := and removed this.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+	ctx = context.TODO()
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[0]
+		})
+		It("Should return A record for query type A", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			_, err := lh.ServeDNS(ctx, rec, m)
```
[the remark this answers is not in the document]
vthapar: That is done by test.SortAndCheck on Line61. 'test' is the dns test package provided by coredns and includes built in checks for these things. It not only matches error code with reply, but also format of reply message etc.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			resp := rec.Msg
+			respErr := test.SortAndCheck(resp, tc)
```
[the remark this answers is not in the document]
vthapar: Done. I had an extra check in here which I removed, forgot to remove redundant resp.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+	lh = Lighthouse{}
+	lh.SvcsMap = setupServicesMap()
+
+	ctx = context.TODO()
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[0]
+		})
+		It("Should return A record for query type A", func() {
```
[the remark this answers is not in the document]
vthapar: Done. I tweaked the wording when writing tests, missed few cleanups.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+			respErr := test.SortAndCheck(resp, tc)
+			Expect(respErr).NotTo(HaveOccurred())
+		})
+	})
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[1]
+		})
+		It("Should return error for non existent query", func() {
```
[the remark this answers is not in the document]
vthapar: Done.

plugin/lighthouse/lighthouse.go
```diff
@@ -1,38 +1,15 @@
...
-
-	// Call next plugin (if any).
-	return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, nw, r)
-}
-
-// Name implements the Handler interface.
-func (lh Lighthouse) Name() string {
-	return "lighthouse"
+	Next    plugin.Handler
+	SvcsMap map[string]string
```
mpeterson: Just want to add this, as it was not commented and it's something I'd comment in the f-up patch. Since you are using this map as a return type in other places it makes sense to create a type for  it and make it more explicit.

plugin/lighthouse/handler.go
vthapar: Another convention followed all over CoreDNS code.

plugin/lighthouse/lighthouse.go
vthapar: This is just placeholder map. Will be replaced eventually by cache+reader for our CRD. this is just to allow some e2e testing of DNS code while CRD cache+reader is done. But good idea, will get more familiar with how to create types.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+	os.Setenv("LIGHTHOUSE_SVCS", "mysvc=100.96.156.175,dummy=100.96.156.175,none=100.96.156.100")
+	lh = Lighthouse{}
+	lh.SvcsMap = setupServicesMap()
+
+	ctx = context.TODO()
+
+	When("DNS query for typeA record", func() {
+		BeforeEach(func() {
+			tc = tests[0]
+		})
```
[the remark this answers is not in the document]
vthapar: Thanks, this is what I intended to do but couldn't figure out good way to do it.

plugin/lighthouse/handler.go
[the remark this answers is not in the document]
tpantelis: Functionally it doesn't really matter - either way `lh` is passed in. To me it seems more natural for call sites to do `lh.onFailure(...)` then `onFailure(lh, ...)`.

plugin/lighthouse/handler_test.go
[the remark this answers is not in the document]
tpantelis: I meant the return code from `lh.ServeDNS `itself  - you're ignoring it here  with `_`.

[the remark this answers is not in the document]
vthapar: Got it, done. It doesn't add much value, but agree, good to check rcode and error both.

plugin/lighthouse/handler_test.go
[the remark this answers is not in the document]
vthapar: Am still looking into how to do this better with context and scenarios. What I really want to do here is something like scenarios. Will revisit this once understand it better.

plugin/lighthouse/handler.go
vthapar: Not really an error, just something we don't support yet and in which case we just pass on to next plugin if present. Added a debug log.

plugin/lighthouse/handler.go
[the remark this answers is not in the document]
vthapar: This will not be used from outside of this file. Maybe change if do change it to be called from other places?

plugin/lighthouse/handler.go
vthapar: Done

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("[setup] Test basic DNS Handler", func() {
+	Describe("function serveDNS", testDnsServer)
+})
+
+func testDnsServer() {
+	var (
+		lh  Lighthouse
```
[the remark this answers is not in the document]
vthapar: Done

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,63 @@
...
+	a.SetReply(r)
+	a.Authoritative = true
+
+	services := lh.SvcsMap
+	svcName := strings.Split(state.QName(), ".")[0]
+	ip := services[svcName]
+	if ip == "" {
+		return onFailure(lh, ctx, w, r)
+	}
+	var rr dns.RR
```
mpeterson: no need to do this, as only case IPv4 is being implemented here. Thus you could just put `:=` on L34

plugin/lighthouse/handler.go
mpeterson: They do because they support A and AAAA but in this case there is no need. Also, it's not a nice pattern IMHO. I'd change it.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,63 @@
...
+
+	switch state.Family() {
+	case 1:
+		rr = new(dns.A)
+		rr.(*dns.A).Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
+		rr.(*dns.A).A = net.ParseIP(ip).To4()
+	default:
+		log.Debugf("IPv6 queries not supported yet")
```
mpeterson:
```suggestion
		log.Debugf("Only A type queries supported")
```

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+			Qname: "mysvc2.default.svc.cluster.local.", Qtype: dns.TypeAAAA,
+			Rcode: dns.RcodeServerFailure,
+		},
+	}
+
+	os.Setenv("LIGHTHOUSE_SVCS", "mysvc=100.96.156.175,dummy=100.96.156.175,none=100.96.156.100")
+	lh = Lighthouse{}
+	lh.SvcsMap = setupServicesMap()
+
+	ctx = context.TODO()
```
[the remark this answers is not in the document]
mpeterson: I'd prefer if they had `BeforeEach` and `AfterEach` though, as it is safer.

plugin/lighthouse/handler.go
vthapar: Query family is ipv4 or Ipv6, so makes better sense. A and AAAA are type of record that are returned in reply.

plugin/lighthouse/handler.go
vthapar: Done. scope of variable declared in switch/case is within switch case, so declared it here.

plugin/lighthouse/handler.go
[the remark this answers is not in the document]
tpantelis: yeah it's fine either way.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+
+	services := lh.SvcsMap
+	svcName := strings.Split(state.QName(), ".")[0]
+	ip := services[svcName]
+	if ip == "" {
+		return onFailure(lh, ctx, w, r)
+	}
+	rr := new(dns.A)
+
+	switch state.Family() {
```
mpeterson: actually it doesn't make sense for the code as is to have a switch and now having the `rr` outside makes this a bit weird.

Just do an `if` on `state.Family() != 1` and clean this code.

plugin/lighthouse/handler_test.go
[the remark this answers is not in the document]
tpantelis: By scenario I assume you mean the table driven pattern used with straight go testing. I don't think this pattern is really conducive for ginkgo as each test case is intended to have it's own `When`/`It` structure.  I guess you could "shoehorn" it by using a single general `It` for the test case loop and `By` but that doesn't really seem to fit the spirit of ginkgo. I think the compromise is to have a `When`/`It` for each case and put most or all the common test code into a function.

plugin/lighthouse/handler.go
tpantelis: I'm wondering if this should be a Warning instead...

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	default:
+		log.Debugf("IPv6 queries not supported yet")
+		return onFailure(lh, ctx, w, r)
+	}
+
+	a.Answer = []dns.RR{rr}
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	if w.WriteMsg(a) != nil {
+		log.Debugf("Failed to writeMsg")
```
tpantelis: `WriteMsg` returns an error - we should log it here along with the context, eg

 `log.Debugf("Failed to write message %#v: %v", a, err)`

Also maybe this should be Error or Warning?

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	state := request.Request{W: w, Req: r}
+
+	a := new(dns.Msg)
+	a.SetReply(r)
+	a.Authoritative = true
+
+	services := lh.SvcsMap
+	svcName := strings.Split(state.QName(), ".")[0]
+	ip := services[svcName]
+	if ip == "" {
```
tpantelis: I think we should log to Debug here, eg

`log.Debugf("No entry for %s was found", svcName)`

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+		log.Debugf("IPv6 queries not supported yet")
+		return onFailure(lh, ctx, w, r)
+	}
+
+	a.Answer = []dns.RR{rr}
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	if w.WriteMsg(a) != nil {
+		log.Debugf("Failed to writeMsg")
+		return onFailure(lh, ctx, w, r)
```
tpantelis: In this case we're actually handling the query but failed to write the message. On success, we return and of course don't invoke the next plugin so should we defer to the next plugin on failure? It seems we should return a failure code along with the error here.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+		})
+		It("Should return an error for non existent query", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+
```
tpantelis: We should add a test case for when `WriteMsg` fails. For that, you can pass a custom `ResponseWriter` to `dnstest.NewRecorder`, eg

```
type FailingResponseWriter struct {
    test.ResponseWriter
    errorMsg string
}

func (w *FailingResponseWriter) WriteMsg(m *dns.Msg) error {
    return errors.New(w.errorMsg)
}
```

plugin/lighthouse/handler.go
mpeterson: When and if we'll support IPv6 we'll refactor but as of now it doesn't make sense.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+		})
+		It("Should return an error", func() {
+			m := tc.Msg()
+			rec := dnstest.NewRecorder(&test.ResponseWriter{})
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+}
```
tpantelis: I think it would be useful in each test to verify that the `Next plugin.Handler`  was or was not invoked where appropriate and not assume that a `RcodeServerFailure` return code means that `NextOrFailure` was called. You could create a mock `plugin.Handler` that simply records that `ServeDNS` was called or use [gomock](https://github.com/golang/mock).

plugin/lighthouse/handler.go
vthapar: Agree, I realized this too right after pushing the patch.

plugin/lighthouse/handler.go
vthapar: Way plugin model works, it is like a chain and if we can't handle, we pass on to next one. Just coz we couldn't handle, doesn't mean coreDNS will send a failure as other plugin may still respond. Also, the response msg NXDOMAIN serves as good enough indicator that no one could respond to it. Adding a warning may blow up the logs. I haven't seen *any* plugin log info/warning for these kind of errors.

plugin/lighthouse/handler.go
vthapar: Hmm... most other plugins just ignore coz this failing is like printf() failing. Added a check just to satisfy golint. In case of failure, we will go to next plugin and if all fails return NXDOMAIN to user.

plugin/lighthouse/handler.go
vthapar: This code won't last long. svcs map is placeholder code till we have lighthouse CRD available. This will not make it to MVP even.

plugin/lighthouse/handler.go
vthapar: See above. User will know of error as NXDOMAIN response. WriteMsg failing is like basic io failure. Other plugins don't even check for return value and assume success. This was added just to satisfy GoLint.

plugin/lighthouse/handler_test.go
vthapar: In our UTs, we are the only plugin loaded, so it will always be failure. Our code's coverage doesn't care which of the two it is, so we're not really getting any value from mocking `plugin.Handler` That would be more like testing `NextOrFailure`, if it returns next plugin or failure based on another plugin being present or not.

plugin/lighthouse/handler.go
skitt: Shouldn’t the plugin return `SERVFAIL` in this case?

plugin/lighthouse/handler.go
skitt: Wouldn’t `NOTIMP` be more appropriate here?

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	return dns.RcodeSuccess, nil
+}
+
+func onFailure(lh *Lighthouse, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
```
skitt: One source of incomprehension for me, initially, is the name of this function. I think it would be easier to understand if it was named after what it ultimately does, `nextOrFailure`.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	// Use a non-writer so we don't write NXDOMAIN to client
+	nw := nonwriter.New(w)
```
skitt: Is this a pattern copied from other plugins? It seems weird to chain to another plugin but not give it a usable writer!

skitt:
My review is more about learning stuff, not suggesting you did things incorrectly @vthapar!

plugin/lighthouse/handler.go
tpantelis: yeah I don't know what circumstance(s) would cause this to fail or why other plugins ignore the return error - maybe it's oversight or seen as an error that would never occur or just not following best practices.  But if we're going to log it then I think it's useful to include as much context as we can. Since this is an error condition and I assume would be uncommon, it feels like an Error or Warning log level, ie useful to emit by default.

plugin/lighthouse/handler.go
skitt: The full sequence in this scenario is that we fail to write the message, try to pass the query on to the next plugin, and if that fails to find anything, the whole process will end up returning `NXDOMAIN`. We’re not indicating anything as a result of the error, and that bothers me — imagine you’re trying to debug this, as an admin; if you see “no such domain” (or some such error, “no such service” or whatever), will you suspect that there *is* really a matching service but that a technical error prevented the appropriate CoreDNS plugin from finding it?

vthapar: Got it, agree it will help troubleshoot if we do run into this error. And if it is that rare, warning/error shouldn't really show up at all. Will add a warn.

plugin/lighthouse/handler.go
vthapar: I think this is something we may revisit later once we get better handle on how our plugin will be used. For now am erring on side of caution, i.e. assuming there might be some other plugin that might be able to handle a query we can't and that can return response. Note that `NOTIMP` response from CoreDNS doesn't mean "our" plugin doesn't implement it, but rather CoreDNS doesn't implement it. For services that are in-cluster, Kubernetes plugin comes first and it may be able to return IPv6 queries. Won't `NOTIMP` be misleading in that case?

plugin/lighthouse/handler.go
vthapar: Done.

plugin/lighthouse/handler.go
vthapar: This one's from reference plugin. I'll probably visit it later once get a better handle on DNS protocol and how lighthouse fits with other plugins. Other plugins will use their own writer [all do in their ServeDNS, I think this is just to avoid *us* writing NXDOMAIN to client.

vthapar:
> > My review is more about learning stuff, not suggesting you did things incorrectly @vthapar!
> 
> yeah me too  I'm not really familiar with the low-level DNS protocol. And I know you're mainly patterning after existing plugins.

I am learning too. Came up with this code after looking at other plugins, reference plugin etc. and still trying out things. This code will likely change as code gets more complex and use pattern from one of more complex plugins.

plugin/lighthouse/handler.go
vthapar: This is more about what `lighthouse` will do with a failure. If we later decide to have some failure scenario where we just return without checking for next we could revisit this.

plugin/lighthouse/handler.go
vthapar: Based on other comments, figured out a way that I think works. Done.

plugin/lighthouse/handler.go
vthapar: Tried out a few things and you're correct. The use case for non writer is different from what we're doing over here. Corrected this.

plugin/lighthouse/handler_test.go
```diff
@@ -0,0 +1,92 @@
...
+func testDnsServer() {
+	var (
+		lh  *Lighthouse
+		ctx context.Context
+		tc  test.Case
+	)
+	tests := []test.Case{
+		{
+			// service is present.
+			Qname: "mysvc.default.svc.cluster.local.", Qtype: dns.TypeA,
```
skitt: What happens when the cluster’s suffix isn’t `cluster.local`? See our e2e deployments for an example.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,62 @@
...
+		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+	}
+
+	a.Answer = []dns.RR{rr}
+
+	log.Debugf("Responding to query with '%s'", a.Answer)
+	wErr := w.WriteMsg(a)
+	if wErr != nil {
+		log.Errorf("Failed to write message %#v: %v", a, wErr)
+		return dns.RcodeServerFailure, lh.Error("failed to write response")
```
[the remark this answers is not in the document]
vthapar: Will do in the next patch where some of this code will be modified.

plugin/lighthouse/handler.go
```diff
@@ -0,0 +1,62 @@
...
+	ip := services[svcName]
+	if ip == "" {
+		// We can't handle this,let another plugin take an attempt
+		// NOTE: Once we have options enabled, this will only be done if
+		//       fallthrough is enabled.
+		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+	}
+	rr := new(dns.A)
+
+	if state.Family() == 1 {
```
mpeterson: nit: I would do it != 1 and avoid the else

plugin/lighthouse/lighthouse.go
```diff
@@ -1,38 +1,16 @@
...
-
-	"github.com/miekg/dns"
 )
 
 // Define log to be a logger with the plugin name in it. This way we can just use log.Info and
 // friends to log.
 var log = clog.NewWithPlugin("lighthouse")
 
+type ServicesMap map[string]string
 type Lighthouse struct {
```
mpeterson: I would add explicit compiler-time tests that make sure you have the correct interfaces implemented (e.g. error and Handler)

This is done by doing:

```
var _ error = &Lighthouse{}
var _ plugin.Handler = &Lighthouse{}
```

It also has the advantage to show whoever is reading your code that you intend and in fact implement those interfaces (and everyone knows that explicit is better than implicit ;))

plugin/lighthouse/handler_test.go
vthapar: It doesn't matter, at least not yet. I have another patch in works that will require configuration of domains we support and that is when we will add more checks.

PR #15, opened 2019-08-26
Initial lighthouse discovery commit.

aswinsuryan:
This patch registers for the cluster added event and updates the map it maintains

It creates the lighthouse CRD when a service is created and keeps it in the memory.


pkg/controller/contorller.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// clusterEvents is the channel to receive events from all the
+	// existing remote clusters
+	clusterEvents chan *remotecluster.Event
+
+	// processingMutex is used to avoid synchronization issues when handling
+	// objects inside the controller, it's a generalistic lock, although
+	// later in time we can come up with a more granular implementation.
+	processingMutex *sync.Mutex
+}
```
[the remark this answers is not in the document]
aswinsuryan: The code is changed to use workqueues.

pkg/controller/contorller.go
```diff
@@ -0,0 +1,56 @@
+package controller
```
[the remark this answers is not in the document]
aswinsuryan: Updated the file name.

main.go
```diff
@@ -0,0 +1,62 @@
+package main
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,259 @@
...
+		controller   *LightHouseController
+		watchReactor *watchReactor
+	)
+
+	BeforeEach(func() {
+		controller = New(mocks.NewMockFederator(gomock.NewController(GinkgoT())))
+		watchReactor = newWatchReactor(controller)
+	})
+
+	AfterEach(func() {
```
mpeterson: perhaps niling again like you did in the previous test?

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,259 @@
...
+		fakeClientset = fakeCoreClientSet.NewSimpleClientset()
+
+		controller.newClientset = func(c *rest.Config) (kubernetes.Interface, error) {
+			return fakeClientset, nil
+		}
+
+		controller.OnAdd(clusterID, &rest.Config{})
+	})
+
+	AfterEach(func() {
```
mpeterson: same. might be worth to reset the vars.

go.mod
```diff
@@ -5,12 +5,22 @@ go 1.12
...
 	k8s.io/apimachinery v0.0.0-20190629003722-e20a3a656cff
 	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/klog v0.3.3
+	sigs.k8s.io/controller-runtime v0.1.12
+	sigs.k8s.io/kubefed v0.1.0-rc3
 )
 
 replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190223031456-9a5ae4453bd
+
+replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190521190702-177766529176
```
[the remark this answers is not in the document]
aswinsuryan: Since the default dependencies added based on core-dns version does not work client-go version required for submariner components.

go.mod
```diff
@@ -5,12 +5,22 @@ go 1.12
...
+
+replace k8s.io/api => k8s.io/api v0.0.0-20190222213804-5cb15d344471
```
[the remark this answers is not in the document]
aswinsuryan: same as above.

go.mod
```diff
@@ -5,12 +5,22 @@ go 1.12
...
 	github.com/onsi/gomega v1.5.0
-	github.com/submariner-io/submariner v0.0.0-20190708095718-350482d85dd4
+	github.com/submariner-io/admiral v0.0.0-20190829090417-2e381e854f60
+	github.com/submariner-io/submariner v0.0.2-0.20190828132721-a11a9a84c90d
+	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
 	k8s.io/apimachinery v0.0.0-20190629003722-e20a3a656cff
 	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/klog v0.3.3
+	sigs.k8s.io/controller-runtime v0.1.12
+	sigs.k8s.io/kubefed v0.1.0-rc3
```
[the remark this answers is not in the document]
aswinsuryan: This was auto-generated, should we be removing this?

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/submariner-io/admiral/pkg/federate"
+	multiclusterservice "github.com/submariner-io/lighthouse/pkg/apis/lighthouse.submariner.io/v1"
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+type RemoteCluster struct {
+	stopCh chan struct{}
+
+	ClusterID       string
+	serviceInformer cache.SharedIndexInformer
+	queue           workqueue.RateLimitingInterface
+}
+
+func New(federator federate.Federator) *LightHouseController {
+
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// ObjectDeleted is called when an object is deleted
+func (c *LightHouseController) ObjectDeleted(obj interface{}) {
+	service := obj.(*core_v1.Service)
+	serviceName := service.ObjectMeta.Name
+	if _, exists := c.remoteServices[serviceName]; exists {
+		err := c.federator.Delete(c.remoteServices[serviceName])
+		if err != nil {
+			fmt.Println("  Delete failed for service.", err)
```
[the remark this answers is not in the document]
aswinsuryan: done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+		delete(c.remoteServices, serviceName)
+		klog.Infof("Service is deleted from the map: %s", serviceName)
+	}
+	fmt.Println("ServiceHandler.ObjectDeleted")
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// ObjectUpdated is called when an object is updated
+func (c *LightHouseController) ObjectUpdated(objOld, objNew interface{}) {
+	//TODO asuryana
+	fmt.Println("ServiceHandler.ObjectUpdated")
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller_test.go
aswinsuryan: Done.

pkg/controller/controller_test.go
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+func (rc *RemoteCluster) close() {
+	close(rc.stopCh)
+	rc.queue.ShutDown()
+	klog.Infof(" watcher queue shutdown for cluster: %s", rc.ClusterID)
+}
+
+func (c *LightHouseController) Run(remoteCluster *RemoteCluster) {
```
tpantelis: t seems this function (and `runWorker `) should take `*RemoteCluster` as the receiver.

pkg/controller/controller.go
```diff
@@ -0,0 +1,272 @@
...
+			remoteCluster.queue.Forget(key)
+		}
+	}
+}
+
+func (c *LightHouseController) ObjectCreated(obj interface{}, clusterID string) {
+	klog.Info("ServiceHandler.ObjectCreated")
+	service := obj.(*core_v1.Service)
+	serviceName := service.ObjectMeta.Name
+	if rs, exists := c.remoteServiceMap.remoteServices[serviceName]; !exists {
```
tpantelis: This function is called by a per-cluster worker so `remoteServices` is being accessed by multiple threads and thus needs to be synchronized. However this can be avoided by having a `remoteServices` per `RemoteCluster`. In fact you could even make `remoteServices` local to the `runWorker` function instead of storing as a struct field.

pkg/controller/controller.go
```diff
@@ -0,0 +1,272 @@
...
+	// Indirection hook for unit tests to supply fake client sets
+	newClientset func(kubeConfig *rest.Config) (kubernetes.Interface, error)
+}
+
+type RemoteCluster struct {
+	stopCh          chan struct{}
+	ClusterID       string
+	serviceInformer cache.SharedIndexInformer
+	queue           workqueue.RateLimitingInterface
+}
```
tpantelis: Does `RemoteCluster` and `ClusterID` need to be exported?

pkg/main.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func buildKubeFedFederator(stopCh <-chan struct{}) federate.Federator {
+	kubeConfig, err := clientcmd.BuildConfigFromFlags(masterURL, kubeConfig)
+	if err != nil {
+		klog.Fatalf("Error attempting to load kubeconfig: %s", err.Error())
+	}
+	federator, err := kubefed.New(kubeConfig,
+
+		stopCh)
```
tpantelis: nit: you can append this line to 59 to save some lines.

pkg/controller/controller.go
```diff
@@ -0,0 +1,272 @@
...
+
+		obj, exists, err := remoteCluster.serviceInformer.GetIndexer().GetByKey(key.(string))
+
+		if err != nil {
+			klog.Errorf("Error processing %s (will retry): %v", key, err)
+			// requeue the item to work on later
+			remoteCluster.queue.AddRateLimited(key)
+		}
+		if !exists {
+			c.ObjectDeleted(obj, remoteCluster.ClusterID)
```
tpantelis: `obj` will be nil here so `ObjectDeleted` will cause a panic. It looks like you only need the service `Name` which you can extract from the `key`.

pkg/controller/controller.go
tpantelis: Also shouldn't `remoteServices` be keyed by service `Name` and `Namespace`?

pkg/controller/controller.go
```diff
@@ -0,0 +1,272 @@
...
+	stopCh          chan struct{}
+	ClusterID       string
+	serviceInformer cache.SharedIndexInformer
+	queue           workqueue.RateLimitingInterface
+}
+
+func New(federator federate.Federator) *LightHouseController {
+	return &LightHouseController{
+		federator:        federator,
+		remoteServiceMap: remoteServiceMap{make(map[string]*mcservice.MultiClusterService)},
```
mpeterson: create a type for `map[string]*mcservice.MultiClusterService` to make the code more maintainable.

mpeterson: same consideration goes for `RemoteCluster`

pkg/controller/controller.go
```diff
@@ -0,0 +1,263 @@
...
+	c.remoteClusters[clusterID] = remoteCluster
+	c.processingMutex.Unlock()
+	serviceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			if err == nil {
+				remoteCluster.queue.Add(key)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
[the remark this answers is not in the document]
mpeterson: ah cool, my sourcegraph extension wasn't working at the time and couldn't check easily. Thanks for the insight, @tpantelis

pkg/controller/controller.go
mpeterson: nice catch

pkg/controller/controller.go
tpantelis: Some more thoughts ... 

If the federate API had a `CreateOrUpdate` method that gets the existing resource and allows the caller to merge its changes then that would alleviate the need to maintain the `remoteServices` map here. We can consider that later.

The way this code is structured now, a controller restart would basically erase and recreate each `MultiClusterService` piecemeal as it receives each existing service.  The benefit to this approach is that it inherently handles services that were deleted while the controller was down (partially anyway). The potential downside is that it produces unnecessary churn with both federating and for clients observing updates. Not sure if this will be a problem but it's something to keep in mind.  The `CreateOrUpdate` method described above would alleviate it although we'd still have to handle services that no longer exist on restart.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+func (c *remoteCluster) serviceCreated(obj interface{}, clusterID string) error {
+	klog.V(2).Infof("ServiceHandler.ObjectCreated")
+	service := obj.(*core_v1.Service)
+	key := service.ObjectMeta.Name + ":" + service.ObjectMeta.Namespace
+	if rs, exists := c.remoteServiceMap[key]; !exists {
+		c.processingMutex.Lock()
+		defer c.processingMutex.Unlock()
```
tpantelis: You need to synchronize all accesses to `remoteServiceMap`, ie move these 2 lines above line 224. 

Also I think you should use a different `Mutex` here as `processingMutex` is also used for `remoteClusters`. I'd suggest `remoteClustersMutex` and `remoteServicesMutex`.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		ServiceIP: service.Spec.ClusterIP,
+	}
+}
+
+// ObjectDeleted is called when an object is deleted
+func (c *remoteCluster) serviceDeleted(key string) error {
+	if rs, exists := c.remoteServiceMap[key]; exists {
+		if len(rs.Spec.Items) == 1 {
+			c.processingMutex.Lock()
+			defer c.processingMutex.Unlock()
```
tpantelis: Similar as in `serviceCreated`, move these 2 lines above line 264.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		newClientset: func(c *rest.Config) (kubernetes.Interface, error) {
+			return kubernetes.NewForConfig(c)
+		},
+	}
+}
+
+func (c *LightHouseController) Start() error {
+	err := c.federator.WatchClusters(c)
+	if err != nil {
+		return fmt.Errorf("Could not register cluster watch: %v", err)
```
tpantelis: As recommended [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), error strings should not be capitalized

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	c.processingMutex.Lock()
+	defer c.processingMutex.Unlock()
+
+	for _, remoteClusters := range c.remoteClusters {
+		remoteClusters.close()
+	}
+}
+
+func (c *LightHouseController) OnAdd(clusterID string, kubeConfig *rest.Config) {
+	klog.Infof("adding cluster: %s", clusterID)
```
tpantelis: nit - message should be capitalized.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		return
+	}
+
+	//startNewServiceWatcher acquires the lock before accessing the map.
+	c.startNewServiceWatcher(clusterID, clientSet)
+}
+
+func (c *LightHouseController) OnUpdate(clusterID string, kubeConfig *rest.Config) {
+	//TODO asuryana need to implement
+	klog.Infof("updating cluster: %s", clusterID)
```
tpantelis: nit - message should be capitalized.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+// HasSynced is required for the cache.Controller interface.
+func (c *remoteCluster) HasSynced() bool {
+	return c.serviceInformer.HasSynced()
+}
+
+func (c *remoteCluster) runWorker() {
+	for {
+		key, shutdown := c.queue.Get()
+		if shutdown {
+			klog.Infof("Lighthouse watcher for cluster \"%s\" stopped", c.clusterID)
```
tpantelis: Instead of \"%s\" use %q

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+
+func (c *remoteCluster) serviceCreated(obj interface{}, clusterID string) error {
+	klog.V(2).Infof("ServiceHandler.ObjectCreated")
+	service := obj.(*core_v1.Service)
+	key := service.ObjectMeta.Name + ":" + service.ObjectMeta.Namespace
+	if rs, exists := c.remoteServiceMap[key]; !exists {
+		c.processingMutex.Lock()
+		defer c.processingMutex.Unlock()
+		c.remoteServiceMap[key] = c.createLightHouseCRD(service, clusterID)
+		klog.Infof("A new service is created and it is added to the map: %s", key)
```
tpantelis: How about `"Creating a new MultiClusterService with service %q from cluster %q", key, clusterID`?

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	} else {
+		klog.Infof("An addService event was received for a Service already in our cache: %s, updating instead", key)
```
tpantelis: How about `"Adding service from cluster %q to existing MultiClusterService %q", clusterID, key`?

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		if len(rs.Spec.Items) == 1 {
+			c.processingMutex.Lock()
+			defer c.processingMutex.Unlock()
+			err := c.federator.Delete(c.remoteServiceMap[key])
+			if err != nil {
+				klog.Errorf("Delete failed for service : %v", err)
+				return err
+			}
+			delete(c.remoteServiceMap, key)
+			klog.Infof("Service is deleted from the map: %s", key)
```
tpantelis: How about `"Service %q has been deleted from cluster %q  - deleted MultiClusterService", key, c.clusterID`?

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		} else {
+			//TODO asuryana the remote service list needs to be updated.
+			klog.Infof("Service list is updated.")
```
tpantelis: How about `"Service %q has been deleted from cluster %q - removing from MultiClusterService", key, c.clusterID`?

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (c *remoteCluster) createClusterServiceInfo(service *v1.Service, clusterID string) mcservice.ClusterServiceInfo {
+	return mcservice.ClusterServiceInfo{
+		ClusterID: clusterID,
+		ServiceIP: service.Spec.ClusterIP,
+	}
+}
+
+// ObjectDeleted is called when an object is deleted
```
tpantelis: _ObjectDeleted_ -> _serviceDeleted_

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (c *remoteCluster) run() {
+	defer utilruntime.HandleCrash()
+
+	go c.serviceInformer.Run(c.stopCh)
+
+	if !cache.WaitForCacheSync(c.stopCh, c.HasSynced) {
+		utilruntime.HandleError(fmt.Errorf("Timed out waiting for caches to sync"))
+		return
```
tpantelis: I don't think you should return here as it will elide the call to `c.runWorker()` below with no way to recover. Most likely I think this could fail if the remote API server isn't available in which case I believe it could connect later and start the watch. That said, I don't think you really need to call `WaitForCacheSync` at all for your use case.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	go c.runWorker()
+}
+
+// HasSynced is required for the cache.Controller interface.
```
tpantelis: Just curious - why do you need/want to implement the `cache.Controller` interface? Also `remoteCluster` doesn't actually implement `cache.Controller` b/c it's missing the `Run` and `LastSyncResourceVersion` methods.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+func (rc *remoteCluster) close() {
+	close(rc.stopCh)
+	rc.queue.ShutDown()
+	klog.Infof(" watcher queue shutdown for cluster: %s", rc.clusterID)
+}
+
+func (c *remoteCluster) run() {
```
tpantelis: I'd prefer the receiver variable name be `r` for improved clarity to match the first letter of the struct name.

pkg/controller/controller.go
aswinsuryan: Removed it as has sync is no more used.

pkg/controller/controller.go
aswinsuryan: Removed.

pkg/controller/controller.go
aswinsuryan: Done

pkg/controller/controller.go
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		if shutdown {
+			klog.Infof("Lighthouse watcher for cluster \"%s\" stopped", c.clusterID)
+			return
+		}
+		defer c.queue.Done(key)
+
+		obj, exists, err := c.serviceInformer.GetIndexer().GetByKey(key.(string))
+
+		if err != nil {
+			klog.Errorf("Error processing %s (will retry): %v", key, err)
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		klog.V(2).Infof("stopping watcher for %s", clusterID)
+		remoteCluster.close()
+		delete(c.remoteClusters, clusterID)
+	}
+}
+
+func (rc *remoteCluster) close() {
+	close(rc.stopCh)
+	rc.queue.ShutDown()
+	klog.Infof(" watcher queue shutdown for cluster: %s", rc.clusterID)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+		},
+	})
+
+	go remoteCluster.run()
+}
+
+func (c *LightHouseController) removeServiceWatcher(clusterID string) {
+	remoteCluster, ok := c.remoteClusters[clusterID]
+	if ok {
+		klog.V(2).Infof("stopping watcher for %s", clusterID)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+func (c *LightHouseController) OnAdd(clusterID string, kubeConfig *rest.Config) {
+	klog.Infof("adding cluster: %s", clusterID)
+
+	clientSet, err := c.newClientset(kubeConfig)
+	if err != nil {
+		klog.Errorf("error creating clientset for cluster %s: %s", clusterID, err.Error())
+		return
+	}
+
+	//startNewServiceWatcher acquires the lock before accessing the map.
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,288 @@
...
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+)
+
+type remoteServiceMap map[string]*mcservice.MultiClusterService
+type remoteClustersMap map[string]*remoteCluster
+
+type LightHouseController struct {
+	//remoteServices is a map that holds the remote services that are discovered.
+	remoteServiceMap remoteServiceMap
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,262 @@
...
+		It("should distribute the resource", func() {
+			mockFederator.EXPECT().Distribute(EqMultiClusterService(multiClusterService)).Return(nil).Do(func(c *lighthousev1.MultiClusterService) {
+				distributeCalled <- true
+			})
+
+			Expect(createService()).To(Succeed())
+
+			Eventually(distributeCalled, 5).Should(Receive(), "Distribute was not called")
+		})
+	})
```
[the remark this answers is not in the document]
tpantelis: We can enhance the unit tests in follow-ip PR(s). I can help out with that as well (I know Vishal is on PTO for awhile).

pkg/controller/controller.go
```diff
@@ -0,0 +1,275 @@
...
+				klog.Errorf("Error retrieving service with key %q from the cache: %v", key, err)
+				// requeue the item to work on later
+				r.queue.AddRateLimited(key)
+				return
+			}
+			if !exists {
+				err := r.serviceDeleted(key)
+				if err != nil {
+					klog.Errorf("Error deleting service %q in cluster %q: %v", key, err, r.clusterID)
+					r.queue.AddRateLimited(key)
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller.go
```diff
@@ -0,0 +1,275 @@
...
+			if err != nil {
+				klog.Errorf("Error retrieving service with key %q from the cache: %v", key, err)
+				// requeue the item to work on later
+				r.queue.AddRateLimited(key)
+				return
+			}
+			if !exists {
+				err := r.serviceDeleted(key)
+				if err != nil {
+					klog.Errorf("Error deleting service %q in cluster %q: %v", key, err, r.clusterID)
```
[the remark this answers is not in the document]
aswinsuryan: Done

pkg/controller/controller.go
```diff
@@ -0,0 +1,275 @@
...
+					r.queue.AddRateLimited(key)
+				}
+			} else {
+				err := r.serviceCreated(obj, key)
+				if err != nil {
+					klog.Errorf("Error creating service %q in cluster %q: %v", key, err, r.clusterID)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// startNewServiceWatcher acquires the lock before accessing the map.
+	r.startNewServiceWatcher(clusterID, clientSet)
+}
+
+func (r *LightHouseController) OnUpdate(clusterID string, kubeConfig *rest.Config) {
+	// TODO asuryana need to implement
+	klog.Infof("Updating cluster: %s", clusterID)
+	klog.Fatalf("Not implemented yet")
```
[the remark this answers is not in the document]
aswinsuryan: Implemented Update.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+	createService := func() error {
+		_, err := fakeClientset.CoreV1().Services(nameSpace).Create(service)
+		return err
+	}
+
+	When("a Cluster resource is added", func() {
+		It("should distribute the resource", func() {
+			var captured **lighthousev1.MultiClusterService = new(*lighthousev1.MultiClusterService)
```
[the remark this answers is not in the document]
aswinsuryan: removed.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,262 @@
...
+func EqMultiClusterService(expected *lighthousev1.MultiClusterService) *eqMultiClusterService {
+	return &eqMultiClusterService{expected}
+}
+
+func (m *eqMultiClusterService) Matches(x interface{}) bool {
+	actual, ok := x.(*lighthousev1.MultiClusterService)
+	if !ok {
+		return false
+	}
+	return m.expected.Name == actual.Name && reflect.DeepEqual(m.expected.Spec, actual.Spec)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

pkg/controller/controller_test.go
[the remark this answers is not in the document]
aswinsuryan: That will be great! we can do this in a new PR.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,262 @@
...
+			mockFederator.EXPECT().Distribute(EqMultiClusterService(multiClusterService)).Return(nil).Do(func(c *lighthousev1.MultiClusterService) {
+				*captured = c
+				distributeCalled <- true
+			})
+
+			Expect(createService()).To(Succeed())
+
+			Eventually(distributeCalled, 5).Should(Receive(), "Distribute was not called")
+		})
+	})
```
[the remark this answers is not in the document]
aswinsuryan: Can be done in a new PR.

pkg/controller/controller_test.go
```diff
@@ -0,0 +1,262 @@
...
+			mockFederator.EXPECT().Distribute(EqMultiClusterService(multiClusterService)).Return(nil).Do(func(c *lighthousev1.MultiClusterService) {
+				distributeCalled <- true
+			})
+
+			Expect(createService()).To(Succeed())
+
+			Eventually(distributeCalled, 5).Should(Receive(), "Distribute was not called")
+		})
+	})
+}
```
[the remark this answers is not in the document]
aswinsuryan: Can be done in a new PR.

pkg/controller/controller.go
```diff
@@ -0,0 +1,275 @@
...
+	if rs, exists := r.remoteServices[key]; !exists {
+		r.remoteServices[key] = r.createLightHouseCRD(service, r.clusterID)
+		klog.Infof("Creating a new MultiClusterService with service %q from cluster %q", key, r.clusterID)
+	} else {
+		klog.Infof("Adding service from cluster %q to existing MultiClusterService %q", r.clusterID, key)
+		rs.Spec.Items = append(rs.Spec.Items, r.createClusterServiceInfo(service, r.clusterID))
+	}
+	err := r.federator.Distribute(r.remoteServices[key])
+	if err != nil {
+		klog.Errorf("Distribute failed : %v", err)
```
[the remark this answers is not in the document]
aswinsuryan: Done.

tpantelis:
@aswinsuryan I see that you resolved my comments and looks like you pushed your branch again but I don't see any new changes...

aswinsuryan:
@tpantelis Oh forgot to do git add, resubmitted it now.

pkg/controller/controller.go
```diff
@@ -0,0 +1,286 @@
...
+			if err != nil {
+				return err
+			}
+			delete(r.remoteServices, key)
+			klog.Infof("Service %q has been deleted from cluster %q - deleted MultiClusterService", key, r.clusterID)
+		} else {
+			serviceList := rs.Spec.Items
+			for i, service := range serviceList {
+				if service.ClusterID == r.clusterID {
+					serviceList = append(serviceList[:i], serviceList[i+1:]...)
```
tpantelis: You should break from the loop here.

Also using `append` will move elements over which can be inefficient. It's faster to swap the element to delete with the one at the end:

```
serviceList[i] = serviceList[len(serviceList)-1]
serviceList = serviceList[:len(serviceList)-1] 
```
This will not maintain the order but that shouldn't matter.

pkg/controller/controller.go
```diff
@@ -0,0 +1,286 @@
...
+				}
+			}
```
tpantelis: In this loop you reassign the local `serviceList` var but not `rs.Spec.Items` so the change will not be reflected in `rs.Spec.Items`. Either eliminate `serviceList` and use `rs.Spec.Items` directly or set `rs.Spec.Items = serviceList` here.  I suggest the former.

pkg/controller/controller.go
```diff
@@ -0,0 +1,286 @@
...
+	klog.V(2).Infof("In remoteCluster serviceCreated for cluster %q, service %#v", r.clusterID, obj)
+	service := obj.(*core_v1.Service)
+	r.remoteServicesMutex.Lock()
+	defer r.remoteServicesMutex.Unlock()
+	if rs, exists := r.remoteServices[key]; !exists {
+		r.remoteServices[key] = r.createLightHouseCRD(service, r.clusterID)
+		klog.Infof("Creating a new MultiClusterService with service %q from cluster %q", key, r.clusterID)
+	} else {
+		klog.Infof("Adding service from cluster %q to existing MultiClusterService %q", r.clusterID, key)
+		rs.Spec.Items = append(rs.Spec.Items, r.createClusterServiceInfo(service, r.clusterID))
```
tpantelis: Now that you've implemented `OnUpdate`, an element for `r.clusterID` may already exist so you don't want to blindly append. First search for the element, ie:

```
for _, info := range rs.Spec.Items {
    if info.ClusterID == r.clusterID {
        return nil
    }
}

rs.Spec.Items = append(...)
```

pkg/controller/controller.go
```diff
@@ -0,0 +1,285 @@
...
+
+	for _, remoteClusters := range r.remoteClusters {
+		remoteClusters.close()
+	}
+}
+
+func (r *LightHouseController) OnAdd(clusterID string, kubeConfig *rest.Config) {
+	klog.Infof("Adding cluster: %s", clusterID)
+	// startNewServiceWatcher acquires the lock before accessing the map.
+	r.startNewServiceWatcher(clusterID, kubeConfig)
```
tpantelis: Need to `Lock/Unlock` here as `startNewServiceWatcher` no longer does it.

pkg/controller/controller.go
```diff
@@ -0,0 +1,285 @@
...
+		stopCh:              make(chan struct{}),
+		clusterID:           clusterID,
+		serviceInformer:     serviceInformer,
+		queue:               workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		remoteServices:      r.remoteServices,
+		federator:           r.federator,
+		remoteServicesMutex: &sync.Mutex{},
+	}
+	r.remoteClustersMutex.Lock()
+	defer r.remoteClustersMutex.Unlock()
```
tpantelis: This needs to be removed as the callers now `Lock/Unlock` -  note `Lock` is not re-entrant.

tpantelis:
I pushed the changes to address the latest 2 comments and some other minor changes [here](https://github.com/submariner-io/lighthouse/pull/20/commits/11c45f8dacddf70fdbc4def1450b4a3777c26863).  I didn't have permissions to push the commit to this PR. You can either incorporate the commit into this PR or we can merge both PRs.

aswinsuryan:
> I pushed the changes to address the latest 2 comments and some other minor changes [here](https://github.com/submariner-io/lighthouse/pull/20/commits/11c45f8dacddf70fdbc4def1450b4a3777c26863). I didn't have permissions to push the commit to this PR. You can either incorporate the commit into this PR or we can merge both PRs.

@tpantelis Thanks,  we can merge both the PR's .

pkg/controller/controller.go
```diff
@@ -0,0 +1,285 @@
...
+
+func (r *remoteCluster) runWorker() {
+	for {
+		keyObj, shutdown := r.queue.Get()
+		if shutdown {
+			klog.Infof("Lighthouse watcher for cluster \"%q\" stopped", r.clusterID)
+			return
+		}
+		key := keyObj.(string)
+		func() {
```
mangelajo: why do you create a lambda function, and just call it down? L208

This can be remove.

pkg/controller/controller.go
mangelajo: See comment by tpantelis ^

pkg/controller/controller.go
tpantelis: This is done so `r.queue.Done(key)` can be deferred as it needs to be called on every path. Of course the code could be re-structured w/o the func as well. I can push another PR to change that - is that OK?

mangelajo: I think it'd work without the func also, right?

I thought defer was handled at the end of a { .... }  block, Is it function level?

I'll merge.

tpantelis: > I thought defer was handled at the end of a { .... } block, Is it function level?
> 
yes it is

PR #17, opened 2019-09-05
Add config options to plugin

vthapar:
Adds following coredns options to Lighthouse plugin:
 1. fallthrough
 2. zones
 3. clusters


plugin/lighthouse/setup.go
```diff
@@ -22,25 +21,51 @@ func init() {
...
+		}
+		for i, str := range lh.Zones {
+			lh.Zones[i] = plugin.Host(str).Normalize()
+		}
+
+		for c.NextBlock() {
+			switch c.Val() {
+			case "fallthrough":
+				lh.Fall.SetZonesFromArgs(c.RemainingArgs())
+			case "clusters":
```
mangelajo: Why do you add clusters config?, that may be automatically discovered, may be I'm confusing where does this go?

vthapar: This will be used to specify proximity order of clusters. If service is available in multiple clusters, this list will be used to determine which is the closest one, and should return the IP from that one.

plugin/lighthouse/parse.go
```diff
@@ -0,0 +1,117 @@
+package lighthouse
```
mpeterson: might we worth looking at importing this to be used as an external dependency. As these methods and functions are not an exported part of the package this basically means you'll have to create a PR or raise an Issue and make your case for them to export it. I guess it won't happen though but might worth the try.

NOT REVIEWED

plugin/lighthouse/parse_test.go
```diff
@@ -0,0 +1,109 @@
+package lighthouse
```
mpeterson: NOT REVIEWED because I assume it's a copy-paste

vthapar: Test contents are copy paste but they're not using ginkgo, so changed accordingly.

plugin/lighthouse/lighthouse.go
```diff
@@ -1,16 +1,33 @@
...
 	clog "github.com/coredns/coredns/plugin/pkg/log"
 )
 
+const (
+	Svc = "svc"
+	Pod = "pod"
+)
+
+var (
+	errInvalidRequest = errors.New("invalid query name")
```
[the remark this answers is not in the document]
vthapar: This is added to keep parse.go code as-is, so using same name and type etc. It is not used anywhere else.

plugin/lighthouse/handler.go
```diff
@@ -15,31 +15,50 @@ import (
...
+	if state.QType() != dns.TypeA {
+		// We only support TypeA
+		log.Debugf("Only TypeA queries are supported yet")
+		if lh.Fall.Through(state.Name()) {
+			return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+		} else {
+			return dns.RcodeNotImplemented, lh.Error("Only TypeA supported")
+		}
+	}
+	zone = qname[len(qname)-len(zone):] // maintain case of original query
```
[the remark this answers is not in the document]
vthapar: Taken as is. Call in line 19 converts zone to all lowercase for comparison. here we are re-extracting the original case zone from incoming query. qname is fqdn and zone is the domain.

plugin/lighthouse/handler.go
```diff
@@ -15,31 +15,50 @@ import (
...
-		rr.Hdr = dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}
-		rr.A = net.ParseIP(ip).To4()
-	} else {
-		// We don't support IPv6, let another plugin take an attempt
-		log.Debugf("IPv6 queries not supported yet")
-		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+	pReq, pErr := parseRequest(state)
+	if pErr != nil || pReq.podOrSvc != Svc {
+		// We can't handle this
+		if lh.Fall.Through(state.Name()) {
```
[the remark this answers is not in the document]
vthapar: Will do.

plugin/lighthouse/handler.go
```diff
@@ -15,31 +15,50 @@ import (
...
+			return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+		} else {
+			return dns.RcodeNameError, lh.Error("Only services supported")
+		}
+	}
+
+	ip := lh.Lookup(pReq.service)
```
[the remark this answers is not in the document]
vthapar: done.

plugin/lighthouse/handler_test.go
```diff
@@ -64,29 +64,177 @@ func testDnsServer() {
 		})
 	})
 
-	When("DNS query for typeA record", func() {
+	When("TypeA DNS query for service that doesn't exist", func() {
 		BeforeEach(func() {
-			tc = tests[1]
+			tc = test.Case{
+				// service does not exist, expect error.
+				Qname: "mysvc2.default.svc.cluster.local.", Qtype: dns.TypeA,
```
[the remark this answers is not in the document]
vthapar: Done

plugin/lighthouse/handler_test.go
```diff
@@ -64,29 +64,177 @@ func testDnsServer() {
...
+		})
+		It("Should call next plugin", func() {
+			m := tc.Msg()
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+
+	When("TypeAAAA DNS query", func() {
```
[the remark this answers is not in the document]
vthapar: Done

plugin/lighthouse/handler_test.go
```diff
@@ -64,29 +64,177 @@ func testDnsServer() {
...
+		It("Should return RcodeNameError", func() {
 			m := tc.Msg()
-			rec := dnstest.NewRecorder(&test.ResponseWriter{})
 			code, err := lh.ServeDNS(ctx, rec, m)
 			Expect(err).To(HaveOccurred())
 			Expect(code).Should(Equal(tc.Rcode))
 		})
 	})
+
+	When("TypeAAAA DNS query", func() {
```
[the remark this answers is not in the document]
vthapar: Done

plugin/lighthouse/handler.go
```diff
@@ -60,3 +79,9 @@ func (lh *Lighthouse) Error(str string) error {
 func (lh *Lighthouse) Name() string {
 	return "lighthouse"
 }
+
+func (lh *Lighthouse) Lookup(service string) string {
+	services := lh.SvcsMap
+	svcName := strings.Split(service, ".")[0]
```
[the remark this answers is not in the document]
vthapar: In fact it doesn't coz we have test cases covering it without `.` This line is not needed, missed changing it on refactoring.

Done.

plugin/lighthouse/handler_test.go
```diff
@@ -5,57 +5,57 @@ import (
...
+				Qname: "mysvc.default.svc.cluster.local.", Qtype: dns.TypeA,
+				Rcode: dns.RcodeSuccess,
+				Answer: []dns.RR{
+					test.A("mysvc.default.svc.cluster.local.    0    IN    A    100.96.156.175"),
+				},
+			}
 		})
-		It("Should return A record for existent query", func() {
+		It("Should return A record", func() {
 			m := tc.Msg()
```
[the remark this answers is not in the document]
vthapar: Will likely do it later, still getting hang of writing ginkgo and figuring out what are good patterns.

plugin/lighthouse/handler.go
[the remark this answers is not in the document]
tpantelis: I agree with @mpeterson. Can you incorporate your comments above into code comments with maybe a simple example illustrating the qname/zone structure for clarity?

plugin/lighthouse/handler.go
```diff
@@ -15,48 +14,69 @@ import (
...
 	return "lighthouse"
 }
+
+func (lh *Lighthouse) error(str string) error {
+	return plugin.Error(lh.Name(), errors.New(str))
+}
+
+func (lh *Lighthouse) lookup(service string) string {
+	services := lh.SvcsMap
+	return services[service]
```
tpantelis: Any particular reason for using a temp var instead of `return lh.SvcsMap[service]`?

plugin/lighthouse/handler_test.go
```diff
@@ -64,29 +65,181 @@ func testDnsServer() {
...
+		})
+		It("Should return error RcodeNotImplemented", func() {
+			m := tc.Msg()
+			code, err := lh.ServeDNS(ctx, rec, m)
+			Expect(err).To(HaveOccurred())
+			Expect(code).Should(Equal(tc.Rcode))
+		})
+	})
+
+	When("Write msg fails", func() {
```
tpantelis: _"Write msg fails"_ does not match what it actually tests/verifies below.

plugin/lighthouse/setup.go
```diff
@@ -22,25 +21,51 @@ func init() {
...
+				lh.Clusters = c.RemainingArgs()
+			default:
+				if c.Val() != "}" {
+					return &Lighthouse{}, c.Errf("unknown property '%s'", c.Val())
```
tpantelis: nit - this should return  nil instead of `&Lighthouse{}`.

plugin/lighthouse/setup_test.go
```diff
@@ -2,23 +2,87 @@ package lighthouse
...
 	})
 
-	It("Should give error on setup", func() {
-		c := caddy.NewTestController("dns", `lighthouse more`)
-		err := setupLighthouse(c)
-		Expect(err).To(HaveOccurred())
+	It("Should populate fields correctly", func() {
+		config := `lighthouse cluster2.local cluster3.local {
+			clusters cluster2 cluster3
+            fallthrough cluster2.local
```
tpantelis: Indentation looks off here...

plugin/lighthouse/setup_test.go
```diff
@@ -2,23 +2,87 @@ package lighthouse
...
+		Expect(lh.Fall).Should(Equal(fallZones))
+		Expect(lh.Zones).Should(Equal(zones))
+		Expect(lh.Clusters).Should(Equal(clusters))
 	})
 
-})
+	It("Should handle empty zones", func() {
+		config := `lighthouse {
+			clusters cluster2 cluster3
+            fallthrough
```
tpantelis: Indentation looks off here...

plugin/lighthouse/handler_test.go
vthapar: Oops! Copy paste error, forgot to change msg.

plugin/lighthouse/handler.go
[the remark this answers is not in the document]
vthapar: Done.

plugin/lighthouse/handler.go
vthapar: Missed in refactoring. done.

plugin/lighthouse/setup.go
vthapar: Done.

FYI, used this as reference:
https://github.com/coredns/coredns/blob/master/plugin/etcd/setup.go#L98

plugin/lighthouse/setup_test.go
vthapar: Looked correct in my IDE. tab vs spaces thing. Are we supposed to use tab or spaces?

plugin/lighthouse/setup_test.go
vthapar: done. Tabs vs spaces issue.

plugin/lighthouse/setup.go
vthapar: Removed this option. Need a different way to get cluster order from admin, will be a separate feature.

plugin/lighthouse/lighthouse.go
```diff
@@ -1,16 +1,32 @@
...
+
 // Define log to be a logger with the plugin name in it. This way we can just use log.Info and
 // friends to log.
 var log = clog.NewWithPlugin("lighthouse")
 
 type ServicesMap map[string]string
 type Lighthouse struct {
 	Next    plugin.Handler
+	Fall    fall.F
+	Zones   []string
```
tpantelis: These don't need to be exported.

plugin/lighthouse/setup_test.go
```diff
@@ -2,23 +2,76 @@ package lighthouse
...
+		c := caddy.NewTestController("dns", config)
+		lh, err := lighthouseParse(c)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("dummy"))
+		setupErr := setupLighthouse(c) // For coverage
+		Expect(setupErr).To(HaveOccurred())
+		Expect(lh).Should(BeNil())
+
+	})
+}
```
tpantelis: We'll have to figure out how best to resolve the conflicts with https://github.com/submariner-io/lighthouse/pull/19/. Not sure which PR should get merged first...

tpantelis:
LGTM. There's some minor things that can be addressed in a followup.

PR #18, opened 2019-09-12
Create docker image for controller

vthapar:


scripts/kind-e2e/e2e.sh
```diff
@@ -337,6 +344,9 @@ setup_broker
 setup_cluster2_gateway
 setup_cluster3_gateway
 test_connection
+if [[ $4 = true ]]; then
```
skitt: Shouldn’t Lighthouse’s e2e always set the controller up?

PR #19, opened 2019-09-24
Added DNS controller for lighthouse.

aswinsuryan:
1)The DNS controller listens for the MultiClusterServiceEvents
2)It servies the DNS queires using the MultiClusterServiceInfo it received


tpantelis:
I pushed https://github.com/submariner-io/lighthouse/pull/22 to address my comments above.

plugin/lighthouse/setup_test.go
```diff
@@ -22,3 +23,12 @@ var _ = Describe("[setup] Test basic bring up", func() {
...
 })
+
+func testStart() {
+	BuildConfigFromFlags = func(masterUrl, kubeconfigPath string) (*rest.Config, error) {
+		if kubeconfigPath == "" {
+			return &rest.Config{}, nil
+		}
+		return nil, nil
+	}
+}
```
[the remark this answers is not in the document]
tpantelis: I see now that this is done to override `clientcmd.BuildConfigFromFlags` to avoid errors. However this should really be part of test setup (ie `BeforeEach`) rather than nested in a separate `When` block (which looks a bit odd). I made modifications and added more tests/assertions in this [commit](https://github.com/submariner-io/lighthouse/pull/22/commits/8aa36e88c99f3e6cf61712a7d3518885f9d35a64).

aswinsuryan:
> I pushed #22 to address my comments above.

The build seems to be failing in Travis, but it is passing in my local.

tpantelis:
> > I pushed #22 to address my comments above.
> 
> The build seems to be failing in Travis, but it is passing in my local.

yeah for me too

tpantelis:
> > > I pushed #22 to address my comments above.
> > 
> > 
> > The build seems to be failing in Travis, but it is passing in my local.
> 
> yeah for me too

So the failure is

`build github.com/coredns/coredns: cannot load k8s.io/client-go/util/workqueue: open /go/src/github.com/submariner-io/lighthouse/coredns/vendor/k8s.io/client-go/util/workqueue: no such file or directory`

Locally I see that _coredns/vendor/k8s.io/client-go/util/workqueue_ does not exist either but the build still succeeds. This seems to be b/c `go mod` is able to go out and find it and store under _~/go/pkg/mod_. However this doesn't happen on the travis build. I wonder if this is b/c travis is using Go 1.11.x instead of 1.12.x - that's the only possible difference I can see. But either way, we should ensure `k8s.io/client-go/util/workqueue` is present under vendor.  I'll look into it more tomorrow.

scripts/build-coredns
```diff
@@ -20,9 +20,15 @@ cd ${TOPDIR}
 git clone -b v${COREDNS_VERSION} ${COREDNS_GIT}
 cd ${TOPDIR}/coredns
 sed -i '/^kubernetes:kubernetes/a lighthouse:github.com/submariner-io/lighthouse/plugin/lighthouse' plugin.cfg
+sed -i '/^\tgithub.com\/prometheus\/common\ v0.6.0/a \\tgithub.com\/submariner-io\/lighthouse\ v0.0.0-20190903104548-989676a12c1d' go.mod
+sed -i '/^replace github.com\/miekg\/dns\sv1.1.3 => github.com\/miekg\/dns\sv1.1.14/a replace\ k8s.io\/apimachinery\ v0.0.0-20190629003722-e20a3a656cff\ =>\ k8s.io\/apimachinery\ v0.0.0-20190313205120-d7deff9243b1' go.mod
+sed -i '/^replace github.com\/miekg\/dns\sv1.1.3 => github.com\/miekg\/dns\sv1.1.14/a replace\ github.com\/openzipkin-contrib\/zipkin-go-opentracing\ =>\ github.com\/openzipkin-contrib\/zipkin-go-opentracing\ v0.3.5' go.mod
+sed -i '/^replace github.com\/miekg\/dns\sv1.1.3 => github.com\/miekg\/dns\sv1.1.14/a replace\ github.com/bronze1man/goStrongswanVici\ =>\ github.com/mangelajo/goStrongswanVici\ v0.0.0-20190223031456-9a5ae4453bd' go.mod
```
tpantelis: For the 3 lines above, we really want to just append the _replace..._ line to the end of the file which can be done via `sed -i '$a <text to append>' rather than searching for a particular line of text.

aswinsuryan:
> workqueue



> > > > I pushed #22 to address my comments above.
> > > 
> > > 
> > > The build seems to be failing in Travis, but it is passing in my local.
> > 
> > 
> > yeah for me too
> 
> So the failure is
> 
> `build github.com/coredns/coredns: cannot load k8s.io/client-go/util/workqueue: open /go/src/github.com/submariner-io/lighthouse/coredns/vendor/k8s.io/client-go/util/workqueue: no such file or directory`
> 
> Locally I see that _coredns/vendor/k8s.io/client-go/util/workqueue_ does not exist either but the build still succeeds. This seems to be b/c `go mod` is able to go out and find it and store under _~/go/pkg/mod_. However this doesn't happen on the travis build. I wonder if this is b/c travis is using Go 1.11.x instead of 1.12.x - that's the only possible difference I can see. But either way, we should ensure `k8s.io/client-go/util/workqueue` is present under vendor. I'll look into it more tomorrow.

The version k8s.io/client-go v11.0.0+incompatible have it but in the vendor folder seems to be a different version.

tpantelis:
> > So the failure is
> > `build github.com/coredns/coredns: cannot load k8s.io/client-go/util/workqueue: open /go/src/github.com/submariner-io/lighthouse/coredns/vendor/k8s.io/client-go/util/workqueue: no such file or directory`
> > Locally I see that _coredns/vendor/k8s.io/client-go/util/workqueue_ does not exist either but the build still succeeds. This seems to be b/c `go mod` is able to go out and find it and store under _~/go/pkg/mod_. However this doesn't happen on the travis build. I wonder if this is b/c travis is using Go 1.11.x instead of 1.12.x - that's the only possible difference I can see. But either way, we should ensure `k8s.io/client-go/util/workqueue` is present under vendor. I'll look into it more tomorrow.
> 
> The version k8s.io/client-go v11.0.0+incompatible have it but in the vendor folder seems to be a different version.

The _build-coredns_ script adds lighthouse version v0.0.0-20190903104548-989676a12c1d to _go.mod_.  I can't tell exactly what commit that is but it's not any of the commits from this patch so it's pulling in a previously merged version which doesn't have the code in this PR that uses `workqueue`. `go mod vendor` only pulls in packages needed by the code to build so, since nothing needs `workqueue`, it doesn't include it. The _build-coredns_ script then copies the local lighthouse source from this PR into _vendor_ but it doesn't match the original source that was used to build the dependencies under  _vendor_ .

What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_  file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary.  If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces. 

What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.

@aswinsuryan can you grant me permission to push a patch to this PR?

aswinsuryan:
> > > So the failure is
> > > `build github.com/coredns/coredns: cannot load k8s.io/client-go/util/workqueue: open /go/src/github.com/submariner-io/lighthouse/coredns/vendor/k8s.io/client-go/util/workqueue: no such file or directory`
> > > Locally I see that _coredns/vendor/k8s.io/client-go/util/workqueue_ does not exist either but the build still succeeds. This seems to be b/c `go mod` is able to go out and find it and store under _~/go/pkg/mod_. However this doesn't happen on the travis build. I wonder if this is b/c travis is using Go 1.11.x instead of 1.12.x - that's the only possible difference I can see. But either way, we should ensure `k8s.io/client-go/util/workqueue` is present under vendor. I'll look into it more tomorrow.
> > 
> > 
> > The version k8s.io/client-go v11.0.0+incompatible have it but in the vendor folder seems to be a different version.
> 
> The _build-coredns_ script adds lighthouse version v0.0.0-20190903104548-989676a12c1d to _go.mod_. I can't tell exactly what commit that is but it's not any of the commits from this patch so it's pulling in a previously merged version which doesn't have the code in this PR that uses `workqueue`. `go mod vendor` only pulls in packages needed by the code to build so, since nothing needs `workqueue`, it doesn't include it. The _build-coredns_ script then copies the local lighthouse source from this PR into _vendor_ but it doesn't match the original source that was used to build the dependencies under _vendor_ .

The lighthouse copy was required since it was not honouring the dependency added in the go.mod , ideally that step can be skipped as the code is now available in the remote repo.
> 
> What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_ file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary. If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces.
> 
> What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.
> 
> @aswinsuryan can you grant me permission to push a patch to this PR?

I have made you an assignee(no sure if that helps), and  since you are a maintainer you should have access,

tpantelis:
> What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_ file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary. If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces.
> 
> What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.

I finally found out why travis was using vendoring - the Dapper configuration set `-mod=vendor` in `GOFLAGS`. I pushed [commit](https://github.com/submariner-io/lighthouse/pull/19/commits/7f6cbbd922400fa6675a043acb9c662cce330586) that removes this and `go mod vendor` and adds a `replace` line in _go.mod_ so the local version of lighthouse is used.  So now _build-coredns_ succeeds!

I also included the commits from https://github.com/submariner-io/lighthouse/pull/22 in this PR to fix race condition failures in the unit tests.

Unfortunately now the e2e tests fail in the travis build...

aswinsuryan:
> > What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_ file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary. If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces.
> > What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.
> 
> I finally found out why travis was using vendoring - the Dapper configuration set `-mod=vendor` in `GOFLAGS`. I pushed [commit](https://github.com/submariner-io/lighthouse/pull/19/commits/7f6cbbd922400fa6675a043acb9c662cce330586) that removes this and `go mod vendor` and adds a `replace` line in _go.mod_ so the local version of lighthouse is used. So now _build-coredns_ succeeds!
> 
> I also included the commits from #22 in this PR to fix race condition failures in the unit tests.
> 
> Unfortunately now the e2e tests fail in the travis build...

Thanks, that's awesome.
The new failure seems to related to the updated core dns image failing to deploy.  I will have a look at it.

aswinsuryan:
> > What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_ file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary. If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces.
> > What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.
> 
> I finally found out why travis was using vendoring - the Dapper configuration set `-mod=vendor` in `GOFLAGS`. I pushed commit [Fix go mod issue in build-coredns](https://github.com/submariner-io/lighthouse/pull/19/commits/7f6cbbd922400fa6675a043acb9c662cce330586) that removes this and the use of `go mod vendor` and adds a `replace` line in _go.mod_ so the local version of lighthouse is used. So now _build-coredns_ succeeds!
> 
> I also included the commits from #22 in this PR to fix a race condition failures in the unit tests.
> 
> Unfortunately now the e2e tests fail in the travis build...

The e2e was passing before since a dummy value was returned for service discovery. Now when it is replaced with real implementation it is failing. I rebased the code and now lighthouse controller is deployed as a part of e2e. I added lighthouse crd in the clusters and the missing permissions in the e2e script. The coredns is now up and does not have any error logs. Now I am trying to resolve the below error, where the lighthouse service account is lacking permission for creating the federated crd this is required for the service discovery to be successful.

E1015 09:25:59.101296       1 controller.go:201] Error creating service "default/nginx-demo" in cluster "cluster3": error creating federated resoource &unstructured.Unstructured{Object:map[string]interface {}{"apiVersion":"types.kubefed.k8s.io/v1beta1", "kind":"FederatedMultiClusterService", "metadata":map[string]interface {}{"name":"nginx-demo", "namespace":"default"}, "spec":map[string]interface {}{"placement":map[string]interface {}{"clusterSelector":map[string]interface {}{"matchLabels":map[string]interface {}{}}}, "template":map[string]interface {}{"metadata":map[string]interface {}{}, "spec":map[string]interface {}{"clusterServiceInfo":[]interface {}{map[string]interface {}{"clusterDomain":"", "clusterID":"cluster3", "serviceIP":"100.96.117.143"}}}}}}}: federatedmulticlusterservices.types.kubefed.k8s.io is forbidden: User "system:serviceaccount:kube-federation-system:kubefed-controller" cannot create resource "federatedmulticlusterservices" in API group "types.kubefed.k8s.io" in the namespace "default"

package/lighthouse-crd.yaml
```diff
@@ -0,0 +1,32 @@
...
+  names:
+        kind: MultiClusterService
+        plural: multiclusterservices
+        singular: multiclusterservice
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          required: ["clusterServiceInfo"]
```
tpantelis: Does this mean `clusterServiceInfo` must have at least one element? If so I don't think we should impose this restriction in case we don't actually delete `MultiClusterService` instances when the last `ClusterServiceInfo` element is removed.

package/lighthouse-crd.yaml
```diff
@@ -0,0 +1,32 @@
...
+          properties:
+            clusterServiceInfo:
+              properties:
+                clusterID:
+                  type: "string"
+                  minimum: 1
+                clusterDomain:
+                  type: "string"
+                  minimum: 1
```
tpantelis: I'm not clear on what `clusterDomain` is for.  You're forcing it to be specified but it is not used in code, at least yet. Perhaps we should remove this restriction until we have a use case for `clusterDomain`?

aswinsuryan:
> > > What's puzzling though is why it is even using _vendor_ on travis since a _go.mod_ file exists and `-mod=vendor` isn't set for `go build` in the coredns _Makefile_ to force it to use vendoring. Locally it uses go modules and not _vendor_ and thus the error isn't seen however that also means it's not building the local lighthouse source into the coredns binary. If you manually set `-mod=vendor` in the _Makefile_ then it uses _vendor_ and the error reproduces.
> > > What we want to do is tell `go mod` to use the version of lighthouse on the local filesystem - it looks like this can be done with `replace`. We shouldn't even need to use vendoring either. I'll play around with this and put together a patch.
> > 
> > 
> > I finally found out why travis was using vendoring - the Dapper configuration set `-mod=vendor` in `GOFLAGS`. I pushed commit [Fix go mod issue in build-coredns](https://github.com/submariner-io/lighthouse/pull/19/commits/7f6cbbd922400fa6675a043acb9c662cce330586) that removes this and the use of `go mod vendor` and adds a `replace` line in _go.mod_ so the local version of lighthouse is used. So now _build-coredns_ succeeds!
> > I also included the commits from #22 in this PR to fix a race condition failures in the unit tests.
> > Unfortunately now the e2e tests fail in the travis build...
> 
> The e2e was passing before since a dummy value was returned for service discovery. Now when it is replaced with real implementation it is failing. I rebased the code and now lighthouse controller is deployed as a part of e2e. I added lighthouse crd in the clusters and the missing permissions in the e2e script. The coredns is now up and does not have any error logs. Now I am trying to resolve the below error, where the lighthouse service account is lacking permission for creating the federated crd this is required for the service discovery to be successful.
> 
> E1015 09:25:59.101296 1 controller.go:201] Error creating service "default/nginx-demo" in cluster "cluster3": error creating federated resoource &unstructured.Unstructured{Object:map[string]interface {}{"apiVersion":"types.kubefed.k8s.io/v1beta1", "kind":"FederatedMultiClusterService", "metadata":map[string]interface {}{"name":"nginx-demo", "namespace":"default"}, "spec":map[string]interface {}{"placement":map[string]interface {}{"clusterSelector":map[string]interface {}{"matchLabels":map[string]interface {}{}}}, "template":map[string]interface {}{"metadata":map[string]interface {}{}, "spec":map[string]interface {}{"clusterServiceInfo":[]interface {}{map[string]interface {}{"clusterDomain":"", "clusterID":"cluster3", "serviceIP":"100.96.117.143"}}}}}}}: federatedmulticlusterservices.types.kubefed.k8s.io is forbidden: User "system:serviceaccount:kube-federation-system:kubefed-controller" cannot create resource "federatedmulticlusterservices" in API group "types.kubefed.k8s.io" in the namespace "default"

The e2e passed with the changes, we can enhance it further though.

tpantelis:
> Would prefer this to go in after #17
> 
> pull/17 refactors this code, adds better error handling and logging. almost all the code in handler.go will be much simpler after pull/17 changes are factored in.

I'm fine with that.  I did some refactoring in the setup/setup test in my patch so I can adjust/rebase on your changes after #17 is merged.

package/lighthouse-crd.yaml
aswinsuryan: Removed the required field.

aswinsuryan:
> > Would prefer this to go in after #17
> > pull/17 refactors this code, adds better error handling and logging. almost all the code in handler.go will be much simpler after pull/17 changes are factored in.
> 
> I'm fine with that. I did some refactoring in the setup/setup test in my patch so I can adjust/rebase on your changes after #17 is merged.

I am fine with that too.

tpantelis:
@aswinsuryan @vthapar I rebased and incorporated the changes into setup_test and handler_test. Please review...

PR #23, opened 2019-10-09
Dns controller build

aswinsuryan:
Fixed the issues in core-dns build


tpantelis:
This PR isn't needed - the core-dns build issues have been fixed in https://github.com/submariner-io/lighthouse/pull/19.

PR #27, opened 2019-10-18
Enhanced plugin handler tests

tpantelis:
Refactored the tests too reuse code and added more tests cases.

Signed-off-by: Tom Pantelis <tompantelis@gmail.com>

### DIFF
--- a/plugin/lighthouse/handler.go
+++ b/plugin/lighthouse/handler.go
@@ -14,38 +14,43 @@ import (
 // ServeDNS implements the plugin.Handler interface.
 func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
-
-	log.Debugf("Request received for  %q ", state.QName())
-
 	qname := state.QName()
+
+	log.Debugf("Request received for %q", qname)
+
 	// qname: mysvc.default.svc.example.org.
 	// zone:  example.org.
 	// Matches will return zone in all lower cases
 	zone := plugin.Zones(lh.Zones).Matches(qname)
 	if zone == "" {
-		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+		log.Debugf("Request does not match configured zones %v", lh.Zones)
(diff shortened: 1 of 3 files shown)


plugin/lighthouse/handler.go
```diff
@@ -14,38 +14,43 @@ import (
...
 
-	qname := state.QName()
 	// qname: mysvc.default.svc.example.org.
 	// zone:  example.org.
 	// Matches will return zone in all lower cases
 	zone := plugin.Zones(lh.Zones).Matches(qname)
 	if zone == "" {
-		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r)
+		log.Debugf("Request does not match configured zones %v", lh.Zones)
+		return lh.nextOrFailure(state.Name(), ctx, w, r, dns.RcodeNotZone, "No matching zone found")
```
tpantelis: @vthapar Note I changed this to call `lh.nextOrFailure` method to use the fall-through as in the other code paths below. Is this OK or was there a particular reason you called  `plugin.NextOrFailure` directly here?

vthapar: Just missed this in refactoring.

plugin/lighthouse/handler_test.go
```diff
@@ -160,90 +157,72 @@ func testNext() {
...
+			})
 		})
-		It("Should call next plugin", func() {
-			m := tc.Msg()
-			code, err := lh.ServeDNS(context.TODO(), rec, m)
-			Expect(err).To(HaveOccurred())
-			Expect(code).Should(Equal(tc.Rcode))
+	})
+
+	When("type A DNS query for a non-matching lighthouse zone and non-matching fallthrough zone", func() {
```
vthapar: Do we have a test case when no explicit fallthrough zone is configured? It means "all zones match" and it should fallthrough. I think we had test case for it, just wanted to confirm.

tpantelis: So I added a new test case with `lh.Fall = fall.F{}` but it doesn't fall through. Looking at the `Fall.Through` code, if `Zones` is empty then `Matches` returns `""` which fails the check `!= ""`. So there must be at least one matching zone.  Is this what we expect?

PR #31, opened 2019-11-18
fix dependency error - revision is shorter than canonical

### LINKED ISSUE #25: Dependabot can't parse your go.mod
Dependabot couldn't parse the go.mod found at `/go.mod`.

roytman:
fixes #25
"Dependabot couldn't parse the go.mod found at /go.mod.

The error Dependabot encountered was:

go: github.com/submariner-io/submariner@v0.0.2-0.20190828132721-a11a9a84c90d requires
	github.com/mangelajo/goStrongswanVici@v0.0.0-20190223031456-9a5ae4453bd: invalid pseudo-version: revision is shorter than canonical (9a5ae4453bda)"

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/lighthouse/31)


go.mod
```diff
@@ -16,11 +16,9 @@ require (
 	k8s.io/apimachinery v0.0.0-20190629003722-e20a3a656cff
 	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/klog v0.3.3
-	sigs.k8s.io/controller-runtime v0.1.12
-	sigs.k8s.io/kubefed v0.1.0-rc3
 )
 
-replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190223031456-9a5ae4453bd
+replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-9a5ae4453bda
```
mangelajo: Did you try removing that replace?, all required patches are now part of goStrongSwanVici

mangelajo: Ok, removing does not fix it, because we are depending on an older version of submariner, and also from admiral. I'll merge as is and add another patch as follow.up

mangelajo:
I tried to look at the dependency problems , I think it needs to be fixed in both admiral and here, but didn't nail it. We need to look at it later so we can update the subm reference and get rid of the replace.

PR #48, opened 2019-12-23
Support for Globalnet

vthapar:
Globalnet adds a new param, GlobalIp for services that
exist on clusters with overlapping CIDRs. If a Service
has GlobalIp annotation, globalIp should be used instead
of ServiceIp to reply to DNS queries for cross-cluster
services.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -43,6 +43,8 @@ type remoteCluster struct {
 	controller      *LightHouseController
 }
 
+const submarinerIpamGlobalIp = "submariner.io/globalIp"
+
 func New(federator federate.Federator) *LightHouseController {
 	return &LightHouseController{
 		federator:                 federator,
@@ -141,6 +143,9 @@ func (r *LightHouseController) startNewServiceWatcher(clusterID string, kubeConf
 				remoteCluster.queue.Add(key)
 			}
 		},
+		UpdateFunc: func(old, newObj interface{}) {
+			r.handleUpdateService(old, newObj, remoteCluster)
+		},
 		DeleteFunc: func(obj interface{}) {
 			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 			if err == nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -162,6 +167,19 @@ func (r *LightHouseController) removeServiceWatcher(clusterID string) {
...
 }
 
+func (r *LightHouseController) handleUpdateService(old interface{}, newObj interface{}, remoteCluster *remoteCluster) {
+	var key string
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(newObj); err != nil {
+		return
+	}
+	oldGlobalIp := old.(*corev1.Service).GetAnnotations()[submarinerIpamGlobalIp]
+	newGlobalIp := newObj.(*corev1.Service).GetAnnotations()[submarinerIpamGlobalIp]
```
mpeterson: In both cases you should retrieve the value by key with care. Doing like this (instead of the form `value, ok =` is risky, no? 

Unless I'm missing something, should the key not exist this will panic and kill the execution.

pkg/controller/controller.go
```diff
@@ -260,9 +283,13 @@ func (r *remoteCluster) newMultiClusterService(service *corev1.Service, clusterI
 }
 
 func (r *remoteCluster) newClusterServiceInfo(service *corev1.Service, clusterID string) lighthousev1.ClusterServiceInfo {
+	mcsIp := service.GetAnnotations()[submarinerIpamGlobalIp]
```
mpeterson: same here `value, ok`

pkg/controller/controller.go
```diff
@@ -43,6 +43,8 @@ type remoteCluster struct {
 	controller      *LightHouseController
 }
 
+const submarinerIpamGlobalIp = "submariner.io/globalIp"
```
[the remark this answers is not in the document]
vthapar: Would prefer that, but these are across repos. Need to make it a "public" file in submariner?

vthapar:
@mpeterson I had some fixes to this that were under testing. I've pushed them for now to check CI. will be pushing patch with your comments next.

pkg/controller/controller.go
vthapar: I thhought that was an issue only when handling deletes. Is it possible on create/update too?

pkg/controller/controller.go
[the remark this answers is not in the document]
mpeterson: Yeah. It would be ok to make it public.

pkg/controller/controller.go
mpeterson: what I mean is that `GetAnnotations()` returns a `map` and then you are getting the value to store it in `xxxGlobalIp` using `submarinerIpamGlobalIp` as a key. If `GetAnnotations` returns a `map` that does not contain that key, then the get panics. That's why if there is a chance that it won't have the key you should use the getter that also returns an `ok` value.

vthapar: It doesn't panic. All services will never have the annotation when service is first created. If it caused a panic this code won't have worked at all.

pkg/controller/controller.go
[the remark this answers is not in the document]
vthapar: Will do that in follow up, need to make change on submariner first.

PR #49, opened 2020-01-02
Lighthouse dnsserver

aswinsuryan:
An alternative to LH coreDNS plugin is provided.

LH hosts a DNS server in this mode and it coreDNS can be configured to forward the request to this DNS server

PR #52, opened 2020-01-08
Standalone DNS server

tpantelis:
This is an extension of the work introduced by https://github.com/submariner-io/lighthouse/pull/49 and addresses review comments.


scripts/e2e
```diff
@@ -5,10 +5,28 @@ source $(dirname $0)/../scripts/lib/debug_functions
 
 cd $(dirname $0)
 
+echo $1
```
aswinsuryan: These echos shall be removed. It was added for debugging.

tpantelis: Done

PR #60, opened 2020-02-05
Auto federate new namespaces with services

vthapar:
When a new service is created in any cluster, federate
the namespace if not already federated.

Also uses exclusion lists of namespace and service keywords
to avoid federation of system namespaces and services.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/submariner-io/admiral/pkg/federate"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -348,3 +360,34 @@ func (r *remoteCluster) serviceDeleted(key string) error {
...
+	if err != nil {
+		klog.Warningf("Failed to get namespace %s details from cluster  %s", namespace, r.clusterID)
+		return
+	}
+	ns.SetNamespace(namespace)
+	err = r.controller.federator.Distribute(ns)
+	if err != nil {
+		klog.Warningf("Failed to federate namespace %s", namespace)
+	}
+	skipNamespaces[namespace] = true
```
tpantelis: It may be possible for `skipNamespaces` to be accessed by multiple threads so probably should be protected to be on the safe side, either via a lock or sync.Map.

That said, do we really need `skipNamespaces`, ie is it OK to re-federate a namespace? I mean that will happen on restart of the controller anyway.

vthapar: It is more of an optimization. We're already maintaining a list of skipped namespaces, might as well use it to track ones already federated.

PR #72, opened 2020-03-05
Reuse submariner Framework in lighthouse E2E

tpantelis:
When the lighthouse E2E tests were initially added, a snapshot of the submariner
project's framework module was copied into lighthouse. We should reuse annd extend the
submariner framework code instead maintaining a separate copy.

Signed-off-by: Tom Pantelis <tompantelis@gmail.com>


tpantelis:
Eventually (soon) I think it makes sense to move the reusable Framework to armada and also include some of the additional lighthouse feature like `ExecWithOptions`, creating nginx deployment etc. But this should suffice for now.

PR #83, opened 2020-03-19
Deploy with correct release tags

vthapar:
Tagged releases are not being deployed with correct version. This
fixes it by using the correct tag off travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ language: go
 branches:
   only:
   - master
+  - /^v\d+\.\d+\.\d+.*$/
+  - /^release-\d+.\d+.*$/
 
 git:
   depth: false
@@ -31,5 +33,11 @@ deploy:
       branch: master
   - provider: script
     script: bash scripts/release
+    on:
+      branch: /^release-\d+.\d+.*$/
+
+  - provider: script
+    script: bash scripts/release $TRAVIS_TAG
+    skip_cleanup: true
     on:
       tags: true
--- a/scripts/release
+++ b/scripts/release
@@ -4,17 +4,12 @@ set -e
 source $(dirname $0)/lib/version
 source $(dirname $0)/lib/debug_functions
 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -29,7 +31,15 @@ deploy:
     script: bash scripts/release
     on:
       branch: master
+      condition: $DEPLOY = true
```
mangelajo:
```suggestion
```

This part will not work here as we don't have any "DEPLOY=true" on test matrix, just one job.

.travis.yml
```diff
@@ -29,7 +31,15 @@ deploy:
...
     script: bash scripts/release
+    on:
+      branch: /^release-\d+.\d+.*$/
+
+  - provider: script
+    script: bash scripts/release $TRAVIS_TAG
+    skip_cleanup: true
     on:
       tags: true
+      condition: $DEPLOY = true
```
mangelajo:
```suggestion
```
Same

scripts/release
```diff
@@ -1,20 +1,28 @@
...
-    DOCKER_TAG=${VERSION}
-else
-    DOCKER_TAG="latest"
+
+# This flag is passed from travis, so only the right jobs will deploy the
+# container images to quay
+
+if [ "$DEPLOY" != "true" ] ; then
+    echo "We don't need to deploy from this job, if you're trying to deploy manually set DEPLOY=true"
+    exit 0
```
mangelajo:
```suggestion

# This flag is passed from travis, so only the right jobs will deploy the
# container images to quay

if [ "$DEPLOY" != "true" ] ; then
    echo "We don't need to deploy from this job, if you're trying to deploy manually set DEPLOY=true"
    exit 0
```

Also not necessary here, 

The deploy flag is something we set from the job matrix, when we only want one job to deploy. In this case all do. And just the "condition: DEPLOY=true" will have the same effect as this (which is being removed from master, we have a PR), sorry.. :)

PR #89, opened 2020-03-31
Add Lighthouse Agent

### LINKED ISSUE #80: Remove the kubefed dependency in lighthouse

vthapar:
To remove kubefed dependency in lighthouse, we'll be using a different
lighthouse agents running in each cluster syncing multiclusterservices
of their clusters to central broker, similar to way we sync clusters
and endpoints in submariner.

This patch adds the initial lighthouse agent. With these changes,
we can work on deployment code while working on finalizing changes
to remove kubefed altogether.

Fixes #80

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -1,0 +1,205 @@
+package controller
+
+import (
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+const (
+	submarinerIpamGlobalIp = "submariner.io/globalIp"
+)
+
+var excludeSvcList = map[string]bool{"kubernetes": true, "openshift": true}
+
+func New(spec *AgentSpecification, config InformerConfigStruct) *Controller {
+	exclusionNSMap := map[string]bool{"kube-system": true, "submariner": true, "openshift": true,
+		"submariner-operator": true, "kubefed-operator": true}
+	for _, v := range spec.ExcludeNS {
+		exclusionNSMap[v] = true
+	}
+	agentController := &Controller{
+		kubeClientSet:    config.KubeClientSet,
+		serviceWorkqueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Services"),
+		servicesSynced:   config.ServiceInformer.Informer().HasSynced,
+
+		excludeNamespaces: exclusionNSMap,
+	}
+
+	klog.Info("Setting up event handlers")
+	config.ServiceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: agentController.enqueueService,
+		UpdateFunc: func(old, newObj interface{}) {
+			agentController.handleUpdateService(old, newObj)
+		},
+		DeleteFunc: agentController.handleRemovedService,
+	})
+
+	return agentController
+}
+
+func (a *Controller) Run(stopCh <-chan struct{}) error {
+	defer utilruntime.HandleCrash()
+
+	// Start the informer factories to begin populating the informer caches
+	klog.Info("Starting Agent controller")
+
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, a.servicesSynced); !ok {
+		return fmt.Errorf("failed to wait for caches to sync")
+	}
+
+	go wait.Until(a.runServiceWorker, time.Second, stopCh)
+	klog.Info("Lighthouse agent workers started")
+	<-stopCh
+	klog.Info("Lighthouse Agent stopping")
+	return nil
+}
+
+func (a *Controller) enqueueService(obj interface{}) {
+	if key := a.getEnqueueKey(obj); key != "" {
+		klog.V(4).Infof("Enqueueing service %v for agent", key)
+		a.serviceWorkqueue.AddRateLimited(key)
+	}
+}
+
+func (a *Controller) getEnqueueKey(obj interface{}) string {
+	var key string
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return ""
+	}
+	ns, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return ""
+	}
+	if !a.isValidService(ns, name) {
+		return ""
+	}
+	return key
+}
+
+func (a *Controller) isValidService(namespace string, serviceName string) bool {
+	if a.excludeNamespaces[namespace] || excludeSvcList[serviceName] {
+		return false
+	}
+	return true
+}
+
+func (a *Controller) runServiceWorker() {
+	for a.processNextService() {
+
+	}
+}
+
+func (a *Controller) processNextService() bool {
+	obj, shutdown := a.serviceWorkqueue.Get()
+	if shutdown {
+		return false
+	}
+	err := func() error {
+		defer a.serviceWorkqueue.Done(obj)
+
+		key := obj.(string)
+		ns, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
+			a.serviceWorkqueue.Forget(obj)
+			return fmt.Errorf("error while splitting meta namespace key %s: %v", key, err)
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// Retrieve the latest version of object before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			service, err := a.kubeClientSet.CoreV1().Services(ns).Get(name, metav1.GetOptions{})
+
+			if err != nil {
+				if errors.IsNotFound(err) {
+					// already deleted Forget and return
+					return nil
+				}
+				return fmt.Errorf("error retrieving service %s: %v", name, err)
+			}
+
+			return a.serviceUpdater(service, key)
+		})
+
+		if retryErr != nil {
+			logAndRequeue(key, a.serviceWorkqueue)
+		}
+
+		a.serviceWorkqueue.Forget(obj)
+		return nil
+	}()
+
+	if err != nil {
+		utilruntime.HandleError(err)
+		return true
+	}
+
+	return true
+}
+
+func (a *Controller) handleUpdateService(old interface{}, newObj interface{}) {
+	// ServiceIp can't change without deleting service, so we only check for globalIp change
+	if key := a.getEnqueueKey(newObj); key != "" {
+		oldGlobalIp := getGlobalIpFromService(old.(*corev1.Service))
+		newGlobalIp := getGlobalIpFromService(newObj.(*corev1.Service))
+		if oldGlobalIp != newGlobalIp {
+			a.serviceWorkqueue.AddRateLimited(key)
+		}
+	}
+}
+
+func (a *Controller) serviceUpdater(service *corev1.Service, key string) error {
+	klog.V(2).Infof("Updating Service %s", key)
+	return nil
+}
+
+func (a *Controller) handleRemovedService(obj interface{}) {
+	var service *corev1.Service
+	var ok bool
+
+	if service, ok = obj.(*corev1.Service); !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			klog.Errorf("Could not convert object %v to Service", obj)
+			return
+		}
+		service, ok = tombstone.Obj.(*corev1.Service)
+		if !ok {
+			klog.Errorf("Could not convert object tombstone %v to Service", tombstone.Obj)
+			return
+		}
+	}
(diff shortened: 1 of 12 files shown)


pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
+package agent
```
tpantelis: _agent_ appears twice in the import path: _pkg/agent/controllers/agent_. I suggest to shorten to _pkg/agent/controllers_.

pkg/agent/controllers/agent/types.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+type Controller struct {
+	kubeClientSet     kubernetes.Interface
+	serviceWorkqueue  workqueue.RateLimitingInterface
+	servicesSynced    cache.InformerSynced
+	excludeNamespaces map[string]bool
+}
+
+type LighthouseAgentSpecification struct {
```
tpantelis:
```suggestion
type AgentSpecification struct {
```

I don't think the `Lighthouse` prefix is necessary since it's already in the import path.

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+const (
+	handlerResync          = time.Hour * 2
+	submarinerIpamGlobalIp = "submariner.io/globalIp"
+)
+
+var excludeSvcList = map[string]bool{"kubernetes": true, "openshift": true}
+
+func New(spec *LighthouseAgentSpecification, config InformerConfigStruct) (*Controller, error) {
+	exclusionNSMap := map[string]bool{"kube-system": true, "submariner": true, "openshift": true,
+		"submariner-operator": true, "kubefed-operator": true}
```
tpantelis: I guess it doesn't hurt but do we still need to exclude "kubefed-operator"?

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	klog.Info("Setting up event handlers")
+	config.ServiceInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: agentController.enqueueService,
+		UpdateFunc: func(old, newObj interface{}) {
+			agentController.handleUpdateService(old, newObj)
+		},
+		DeleteFunc: agentController.handleRemovedService,
+	}, handlerResync)
```
tpantelis: The resync period is already set to 60 sec on the Informer - any particular reason to override that here?

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+	klog.Info("Starting IPAM Controller")
+
+	// Wait for the caches to be synced before starting workers
+	klog.Info("Waiting for informer caches to sync")
+	if ok := cache.WaitForCacheSync(stopCh, a.servicesSynced); !ok {
+		return fmt.Errorf("failed to wait for caches to sync")
+	}
+
+	go wait.Until(a.runServiceWorker, time.Second, stopCh)
+	wg.Wait()
```
tpantelis: What is the reason for the `WaitGroup`? `Done` is never called so this will wait forever.

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+	klog.Info("Lighthouse agent workers started")
+	<-stopCh
+	klog.Info("Shutting down workers")
```
tpantelis:
```suggestion
	klog.Info("Lighthouse agent stopping")
```

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+		UpdateFunc: func(old, newObj interface{}) {
+			agentController.handleUpdateService(old, newObj)
+		},
+		DeleteFunc: agentController.handleRemovedService,
+	}, handlerResync)
+
+	return agentController, nil
+}
+
+func (a *Controller) Run(stopCh <-chan struct{}) error {
```
tpantelis: nit - the receiver should probably be called `c`. Same in the other methods.

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+					// could be API error, so we want to requeue
+					logAndRequeue(key, a.serviceWorkqueue)
+				}
+				return fmt.Errorf("error retrieving service %s: %v", name, err)
+			}
+
+			return a.serviceUpdater(service, key)
+		})
+		if retryErr != nil {
+			return retryErr
```
tpantelis: If max retries was reached then we'd want to either call `Forget` or re-queue. I'd suggest to return a nil error above for the `errors.IsNotFound(err)` case, then here re-queue for any error returned.

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,205 @@
...
+	klog.V(2).Infof("Updating Service %s", key)
+	return nil
+}
+
+func (a *Controller) handleRemovedService(obj interface{}) {
+	var service *corev1.Service
+	var ok bool
+	var key string
+
+	if key = a.getEnqueueKey(obj); key == "" {
```
tpantelis: You don't need to call this since you're not queueing.

pkg/signals/signal_posix.go
```diff
@@ -0,0 +1,23 @@
+// +build !windows
```
tpantelis: You don't need to copy these files - you can reuse from controller-runtime - see https://github.com/submariner-io/lighthouse/pull/88.

pkg/agent/controllers/agent/agent.go
vthapar: I am still working on this, was refactoring stuff and trying out different values. Will use the one on informer.

pkg/agent/controllers/agent/agent.go
vthapar: More of a reminder that this is Agent controller. Have used similar convention elsewhere, lighthouse Controller uses l etc.

pkg/agent/controllers/agent/agent.go
vthapar: Reusing this pattern from what I used in globalnet. Still find this a bit confusing.

pkg/agent/controllers/agent/agent.go
vthapar: For `errors.IsNotFound` we want to Forget coz it is expected to get that error when object is already deleted. If we keep requeuing on max retries, it kinda defeats point of a max retries. I think we can revisit this if we run into issues during scale/perf testing. Thoughts?

pkg/agent/controllers/agent/agent.go
vthapar: Should probably rename or split getEnqueueKey. If you notice, am also using getEnqueueKey to find out if we should process this service or not. If not, it returns emtpy key.

pkg/agent/controllers/agent/agent.go
vthapar: Problem is, main.go and agent.go can't be in same root. So I need another folder within controllers. We had similar issue with globlanet too, if you recall. I'll refactor this later once have better clarity on how to use admiral API. Was thinking we could use something like.`controllers/datasyncer`.

pkg/agent/controllers/agent/agent.go
vthapar: Will change once we remove kubefed. This helps cut out noise from kubefed services.

pkg/agent/main.go
```diff
@@ -0,0 +1,73 @@
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err = lightHouseAgent.Run(stopCh); err != nil {
+			klog.Fatalf("Failed to start lighthouse agent: %v", err)
+		}
+
+	}()
+
+	wg.Wait()
```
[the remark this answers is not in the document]
vthapar: Following the pattern I used in globalnet. I think this one allows us to have the fatal log in case it fails.

pkg/agent/controllers/agent/agent.go
tpantelis: **main.go** goes in _pkg/agent_ as you have it. For the datasyncer part, I believe you'll just instantiate and run the component from admiral, ie you won't need to implement anything.  Either way, if there is another controller or something, it can just reside in _pkg/agent/controllers_ - you don't really need a sub-directory structure under controllers.

pkg/agent/controllers/agent/agent.go
tpantelis: yeah - the `WaitGroup` here is not needed as you wait on `stopCh` below.  But it'll never get past this line 😄

pkg/agent/controllers/agent/agent.go
tpantelis: Either way you would still `Forget` on L126. But you also need to `Forget` if max retries is reached - it's really a bug otherwise. But I'm suggesting to simplify the code and just re-queue on any error here, which means `IsNotFound` above would return nil error and also not log since it's not really an error.  If max retries is reached I think we still want to retry later (see comment in `handleUpdateService`).

pkg/agent/controllers/agent/agent.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (a *Controller) handleUpdateService(old interface{}, newObj interface{}) {
+	// ServiceIp can't change without deleting service, so we only check for globalIp change
+	if key := a.getEnqueueKey(newObj); key != "" {
+		oldGlobalIp := getGlobalIpFromService(old.(*corev1.Service))
+		newGlobalIp := getGlobalIpFromService(newObj.(*corev1.Service))
+		if oldGlobalIp != newGlobalIp {
+			a.serviceWorkqueue.AddRateLimited(key)
+		}
```
tpantelis, line 159: Update will also be called at each resync period, in which case the objects will likely be equal. With this check, auto resync is rendered essentially useless and we should just disable it, which is fine as long as we re-queue/retry on all failures.

pkg/agent/controllers/agent/agent.go
tpantelis: ahh right. For the check, you just need the namespace which you can extract directly from the service `obj`.  This will also make it easer to use with the workqueue wrapper added by https://github.com/submariner-io/admiral/pull/45. We can migrate to use that later.

pkg/agent/main.go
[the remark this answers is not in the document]
tpantelis: I'm not sure I follow - are you referring to L61?  Without the `WaitGroup` and goroutine, L55-66 simplify to:

```
if err = lightHouseAgent.Run(stopCh); err != nil {
    klog.Fatalf("Failed to start lighthouse agent: %v", err)
}
```

You'd only need a `WaitGroup` if you need to wait on multiple things.

pkg/agent/controllers/agent/agent.go
vthapar: Good idea. I was thinking we may have multiple controllers, depending on how we're going to distribute the MCS crds so was thinking about multiple dirs, but that can be easily done within a single controller directory too.

pkg/agent/controllers/agent/agent.go
vthapar: Now that I got basics working, will rework it. I still don't fully understand, so would help if you keep a sharp eye on this one :)

pkg/agent/controllers/agent/agent.go
vthapar: Aha, got it. Will do.

pkg/agent/controllers/agent/agent.go
vthapar: Good idea. This sounds preferable than auto-resync.

pkg/agent/controllers/agent/agent.go
vthapar: Yeah, will refactor to separate the getKey from check.

pkg/agent/main.go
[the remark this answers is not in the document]
vthapar: We'll likely need one more for `multiclusterservices`.

pkg/agent/controller/agent.go
```diff
@@ -0,0 +1,207 @@
...
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			service, err := a.kubeClientSet.CoreV1().Services(ns).Get(name, metav1.GetOptions{})
+
+			if err != nil {
+				if errors.IsNotFound(err) {
+					// already deleted Forget and return
+					return nil
+				} else {
+					// could be API error, so we want to requeue
+					logAndRequeue(key, a.serviceWorkqueue)
```
tpantelis, line 133: `logAndRequeue` is called on L141 so the else clause here can be removed.

pkg/agent/controller/agent.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	klog.Info("Setting up event handlers")
+	config.ServiceInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: agentController.enqueueService,
+		UpdateFunc: func(old, newObj interface{}) {
+			agentController.handleUpdateService(old, newObj)
+		},
+		DeleteFunc: agentController.handleRemovedService,
+	}, 0)
```
tpantelis, line 45: I'm not sure what affect passing 0 has but we should just call `AddEventHandler`.

vthapar: 0 disables resync. without argument it uses the default specified in informer.

pkg/agent/controller/agent.go
vthapar: Good call. missed this.

pkg/agent/controller/agent.go
tpantelis: right - sorry what I meant was I'm not sure what affect passing 0 has wrt overriding the non-zero setting on the informer. I think perhaps we should configure the resync period in one place at the informer level.  The informer is actually only used here anyway. There's no point  in setting it to 2 min in **main** only to disable it here (if that is indeed the override behavior).

PR #98, opened 2020-04-21
DO NOT MERGE YET Use shipyard for e2e framework

vthapar:
Remove kubefed framework added to lighthouse and use framework from shipyard directly. Also adds lighthouse flag to Makefile so we can reuse scripts from shipyard. This also allows us to use helm for lighthouse deployment.


vthapar:
This requires following to work:
https://github.com/submariner-io/submariner-charts/pull/27
https://github.com/submariner-io/shipyard/pull/108

Makefile
```diff
@@ -4,9 +4,12 @@ logging ?= false
 coredns ?= 1.5.2
 deploytool ?= helm
 globalnet ?= false
+lighthouse ?= true
 
 TARGETS := $(shell ls scripts | grep -v deploy)
 SCRIPTS_DIR ?= /opt/shipyard/scripts
+CLUSTER_ARGS = --lighthouse ${lighthouse}
+DEPLOY_ARGS = --lighthouse ${lighthouse}
```
tpantelis, line 12: What are these for?

vthapar: Stephen mentioned they're going to remove using flags in shipyard and use these args instead. You can refer his comment on my shipyard patch.

tpantelis: yeah just saw that. We also need to switch to use shipyard's Makefile.inc. I was actually working on that but I'll wait for this to land first.

tpantelis: So do we even need the `lighthouse` flag on L7? I mean it doesn't seem to  make sense for someone to run lighthouse E2E w/o lighthouse installed 😄

mkolesnik: Take a look at my proposal here https://github.com/submariner-io/submariner/pull/516

vthapar: I'd like to do this alongwith other changes to reuse Makefile.inc from shipyard, but I ran into issues getting coredns build working with it. I'd prefer to get this in to have an working RC. The refactoring to use shipyard scripts can be done independent of lighthouse agent changes.

tpantelis: I've done the shipyard changes in another patch - all seems to be working fine. As I mentioned, I'll wait for this to land first before submitting it.

mkolesnik:
I think we can close this in favor of #101

PR #107, opened 2020-04-28
Use DEPLOY_ARGS instead of variables

mkolesnik:
We plan to drop the variabes in Shipyard's Makefile.inc in favor of the
ARGS. Hence, using DEPLOY_ARGS to pass arguments to the deploy script.


mkolesnik:
Forgot we set them internally in the Makefile :)

PR #118, opened 2020-04-30
Use specific cluster functions

mkolesnik:
After introduction of cluster functions [1] we can use them to make our
code nicer and smarter.

[1] https://github.com/submariner-io/shipyard/pull/127

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -4,6 +4,7 @@ source ${SCRIPTS_DIR}/lib/debug_functions
 source ${SCRIPTS_DIR}/lib/deploy_funcs
 source ${SCRIPTS_DIR}/lib/version
 source ${SCRIPTS_DIR}/lib/utils
+source ${SCRIPTS_DIR}/lib/cluster_settings
 
 ### Functions ###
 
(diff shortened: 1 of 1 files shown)


mkolesnik:
This one needs https://github.com/submariner-io/shipyard/pull/127 merged first, so it'll fail until it is

scripts/kind-e2e/e2e.sh
```diff
@@ -36,9 +37,9 @@ function test_with_e2e_tests {
 declare_kubeconfig
 PRJ_ROOT=$(git rev-parse --show-toplevel)
 
-run_parallel "2 3" update_coredns_configmap
+run_subm_clusters update_coredns_configmap
 import_image lighthouse-coredns
-run_parallel "{1..3}" update_coredns_deployment
+run_subm_clusters update_coredns_deployment
```
tpantelis, line 42: Interesting that the `run_subm_clusters`  calls failed as expected in the CI jobs but the jobs still succeeded 😄

mkolesnik: Yes it's because we dont `set -o pipefail` on all the scripts so some places its not set and then even if it fails the E2E the script doesn't exit so the job thinks all is good.. :/

PR #124, opened 2020-05-07
Fix version script

mkolesnik:
Copied over the script from submariner for calculating the image version

### DIFF
--- a/scripts/lib/version
+++ b/scripts/lib/version
@@ -1,16 +1,18 @@
-#!/bin/bash
-
 # This should only be sourced
 if [ "${0##*/}" = "version" ]; then
     echo "Don't run me, source me" >&2
     exit 1
 fi
 
+if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+    DIRTY="-dirty"
+fi
+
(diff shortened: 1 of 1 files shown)


skitt:
Wouldn’t this be better as a shared script in Shipyard?

aswinsuryan:
> Wouldn’t this be better as a shared script in Shipyard?

raised an issue to keep track of it. https://github.com/submariner-io/shipyard/issues/144

mkolesnik:
> Wouldn’t this be better as a shared script in Shipyard?

It's already a shared script, theres also issues open to adopt more parts of shipyard (which use this script)

skitt:
> > Wouldn’t this be better as a shared script in Shipyard?
> 
> It's already a shared script, theres also issues open to adopt more parts of shipyard (which use this script)

Indeed; the fact that this updates the copied file made me think it wasn’t.

mkolesnik:
> 
> Indeed; the fact that this updates the copied file made me think it wasn’t.

It was a stop gap solution, I'm working on adapting the Shipyard stuff now

PR #128, opened 2020-05-11
Basic Round Robin selection.

danibachar:
First step toward optimized service selection.


pkg/multiclusterservice/map.go
```diff
@@ -10,13 +10,31 @@ type serviceInfo struct {
...
+// NEW
+func (m *Map) GetBestIP(namespace string, name string) (string, bool) {
+	m.RLock()
+	defer m.RUnlock()
+
+	if val, ok := m.svcMap[keyFunc(namespace, name)]; ok {
+		ipsCount := len(val.ipList)
+		if ipsCount < 1 { return "", false  }
+		selIP := val.ipList[val.rrCount%ipsCount]
+		val.rrCount++
```
vthapar, line 31: This will need a write lock, unless we're okay with pseudo-round-robin. Am vary of adding a write lock here coz we'll be taking it during all reads. So basically only one read can proceed at a time. I think we need to revisit the locking.

danibachar: @vthapar 
hmmm... good point.
I thought we can extract the `selection state` (which will include metrics and other info in the future) into another object to avoid the lock on the Map object, and we will update it using the Put/Delete function.
I'll try to come up with some rough ideas soon and add here to the PR.
Regarding the Lock mechanism - I'm into revisit it 👍

pkg/multiclusterservice/map.go
```diff
@@ -10,13 +10,31 @@ type serviceInfo struct {
 	key         string
 	clusterInfo map[string]string
 	ipList      []string
+	rrCount     int // Counter for Round Ronbin IP selection, will be replaced by metirces object
```
mangelajo, line 13:
```suggestion
	rrCount     int // Counter for Round Robin IP selection, will be replaced by metrics object
```

pkg/multiclusterservice/map.go
mangelajo: I think the lock would be fine, since we are not doing any blocking operation during the lock time. It's just accessing structures and calculating stuff.

I agree rrCount access needs to be thread safe, I recommend looking at:

https://gobyexample.com/atomic-counters   you can use atomic.AddUintxx or atomic.LoadUintxx

tpantelis: We could also avoid locking on the fast path by using copy-on-write semantics for the map.

danibachar: @mangelajo @vthapar @tpantelis 
Fixed - let me know what do you think

PR #133, opened 2020-05-13
Create ServiceExport CRD

vthapar:


scripts/codegen
```diff
@@ -0,0 +1,20 @@
...
+
+if [ ! -f "$CODEGEN_SCRIPT" ]; then
+    echo "$CODEGEN_SCRIPT does not exist - downloading..."
+    cd $K8S_IO_DIR
+    git clone --branch $CODEGEN_RELEASE_TAG https://github.com/kubernetes/code-generator
+    cd -
+    echo "Successsfully checked out release tag $CODEGEN_RELEASE_TAG"
+fi
+
+$CODEGEN_SCRIPT all github.com/submariner-io/lighthouse/pkg/client github.com/submariner-io/lighthouse/pkg/apis lighthouse.submariner.io:v2alpha1
```
skitt, line 20:
```suggestion
$CODEGEN_SCRIPT all github.com/submariner-io/lighthouse/pkg/client github.com/submariner-io/lighthouse/pkg/apis lighthouse.submariner.io:v1,v2alpha1
```

This will generate client code for both v1 and v2alpha1, including informers.

pkg/apis/lighthouse.submariner.io/v2alpha1/doc.go
```diff
@@ -0,0 +1,4 @@
+// +k8s:deepcopy-gen=package
```
mangelajo, line 1: I know it's weird, but let's jump back to v1alpha1, I think it's something we can afford at this stage.

mangelajo: Specially since we are even changing resource names, etc.

mangelajo: We agreed on slack, to be good and move to v2alpha

pkg/apis/lighthouse.submariner.io/v1/zz_generated.deepcopy.go
```diff
@@ -71,7 +71,7 @@ func (in *MultiClusterService) DeepCopyObject() runtime.Object {
 func (in *MultiClusterServiceList) DeepCopyInto(out *MultiClusterServiceList) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
-	in.ListMeta.DeepCopyInto(&out.ListMeta)
+	out.ListMeta = in.ListMeta
```
mangelajo, line 74: Why did it change this? :? it's weird.

mangelajo: I guess they used slightly different generators, it's probably fine.

PR #134, opened 2020-05-13
Use AwaitUntil() when running curl

vthapar:
When service is created/deleted, it can take a while to update
multiclusterservices objects for it across clusters.

Added retry to curl so we wait for curl to fail and not return
at initial success during delete use case.

### DIFF
--- a/test/e2e/dataplane/service_discovery.go
+++ b/test/e2e/dataplane/service_discovery.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/e2e/dataplane/service_discovery.go
```diff
@@ -122,19 +121,23 @@ func verifyClusterIpWithDig(f *framework.Framework, cluster framework.ClusterInd
...
+	framework.AwaitUntil("verify if service is discoverable", func() (interface{}, error) {
+		stdout, _, err := f.ExecWithOptions(framework.ExecOptions{
+			Command:       cmd,
+			Namespace:     f.Namespace,
+			PodName:       pod.Name,
+			ContainerName: pod.Spec.Containers[0].Name,
+			CaptureStdout: true,
+			CaptureStderr: true,
+		}, cluster)
+		if (err != nil && expectingSuccess) || (err == nil && !expectingSuccess) {
```
mkolesnik, line 136: Won't this condition give a false positive?
`(err == nil && !expectingSuccess)`

 i.e. if err= nil then it'll return nil as the error, and youll get a success even though you were were expecting it to fail

vthapar: > Won't this condition give a false positive?
> `(err == nil && !expectingSuccess)`
> 
> i.e. if err= nil then it'll return nil as the error, and youll get a success even though you were were expecting it to fail

Yes, will fix this.

mangelajo:
I think you should at least change the commit to say "use retries" , AwaitUntil is an specific
function of the framework you could use for that :D

vthapar:
> I think you should at least change the commit to say "use retries" , AwaitUntil is an specific
> function of the framework you could use for that :D

I did update the git commit msg and title, but github doesn't update the PR. will do.

test/e2e/dataplane/service_discovery.go
```diff
@@ -122,19 +123,24 @@ func verifyClusterIpWithDig(f *framework.Framework, cluster framework.ClusterInd
...
+			Command:       cmd,
+			Namespace:     f.Namespace,
+			PodName:       pod.Name,
+			ContainerName: pod.Spec.Containers[0].Name,
+			CaptureStdout: true,
+			CaptureStderr: true,
+		}, cluster)
+		if err == nil && !expectingSuccess {
+			By(fmt.Sprintf("Retrying curl because got no error when expected to get one"))
+			time.Sleep(time.Second)
```
mkolesnik, line 140: I don't think this sleep of a second matters much

PR #135, opened 2020-05-13
Increase lint timeout, ignore generated clientset

skitt:
The updated golangci-lint identifies an issue in the clientset, but
since that’s generated we should ignore it instead.

The timeout also needs to be increased.

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -19,4 +19,5 @@ if [[ $(goimports -l ${PACKAGES} | wc -l) -gt 0 ]]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


mangelajo:
@vthapar this fails the E2E tests now, shall we disable them now? 

You can change the "Describe"s to "PDescribe"s

mangelajo:
> @vthapar this fails the E2E tests now, shall we disable them now?
> 
> You can change the "Describe"s to "PDescribe"s

Sorry, I thought I was on your patch here.

PR #137, opened 2020-05-15
Use ServiceExport in agent

vthapar:
Use ServicExport to decide which services will be multicluster. This
changes the to an opt-in model from current all-in default.

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -6,6 +6,7 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/submariner-io/lighthouse/pkg/agent/controller"
 	lighthousev1 "github.com/submariner-io/lighthouse/pkg/apis/lighthouse.submariner.io/v1"
+	lighthousev2a1 "github.com/submariner-io/lighthouse/pkg/apis/lighthouse.submariner.io/v2alpha1"
 	"k8s.io/client-go/kubernetes/scheme"
 
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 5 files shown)


test/e2e/dataplane/service_discovery.go
```diff
@@ -4,34 +4,36 @@ import (
...
 	})
-
-	When("a pod tries to resolve a service which is present locally and in a remote cluster", func() {
-		It("should resolve the local service", func() {
-			RunServiceDiscoveryLocalTest(f)
+	/*
+		When("a pod tries to resolve a service which is present locally and in a remote cluster", func() {
+			It("should resolve the local service", func() {
+				RunServiceDiscoveryLocalTest(f)
+			})
```
skitt, line 31: Why is this commented out?

vthapar: Argh. Missed to uncomment after refactoring the dig method.

pkg/agent/main.go
```diff
@@ -35,6 +36,10 @@ func main() {
 	if err != nil {
 		klog.Exitf("Error adding lighthouse V1 to the scheme: %v", err)
 	}
+	err = lighthousev2a1.AddToScheme(scheme.Scheme)
```
mangelajo, line 39: This is something the generator does not do for some reason.

We need to register them in the register.go part, not here :)

https://github.com/submariner-io/submariner/blob/master/pkg/apis/submariner.io/v1/register.go#L28

@skitt also had a fun time about it.

vthapar: > This is something the generator does not do for some reason.
> 
> We need to register them in the register.go part, not here :)
> 
> https://github.com/submariner-io/submariner/blob/master/pkg/apis/submariner.io/v1/register.go#L28
> 
> @skitt also had a fun time about it.

That is already done:
https://github.com/submariner-io/lighthouse/blob/master/pkg/apis/lighthouse.submariner.io/v2alpha1/register.go#L31

Without this line I was getting an error at runtime. We had to add it for multiclusterservices too.

mangelajo: Hmm, may be it was somewhere else @skitt do you remember?

thanks vishal

PR #151, opened 2020-05-22
Use the same e2e target clusters in lighthouse as for submariner.

mangelajo:
In this way, the e2e scripts can be reused from subctl verify.


mangelajo:
This was already handled here: https://github.com/submariner-io/lighthouse/pull/153

PR #153, opened 2020-05-24
Use only two clusters for E2E

mkolesnik:
We don't really need 3 clusters, use only 2 for testing.


mangelajo:
Makes sense, and also includes the changes in https://github.com/submariner-io/lighthouse/pull/151 which will make lighthouse e2e consumable from subctl.

PR #156, opened 2020-06-01
Modify test to use supercluster.local

aswinsuryan:
The patch modifies test to resolve supercluster.local if
lhdnserver deployment is present.


vthapar:
I tried it in a different way: https://github.com/submariner-io/lighthouse/pull/158 We want to use supercluster.local even for non lighthouse-dns model with opt-in.

aswinsuryan:
> I tried it in a different way: #158 We want to use supercluster.local even for non lighthouse-dns model with opt-in.

ya that's all we need, I shall close this one

PR #160, opened 2020-06-02
Modified tests to use forward plugin

aswinsuryan:
Changed the e2e deployment to use forward plugin implemention


scripts/deploy
```diff
@@ -36,8 +33,9 @@ declare_kubeconfig
 import_image quay.io/submariner/lighthouse-agent
 import_image quay.io/submariner/lighthouse-coredns
 
-run_subm_clusters update_coredns_configmap
-run_subm_clusters update_coredns_deployment
 #run_subm_clusters install_service_export
 
 ${SCRIPTS_DIR}/deploy.sh "$@"
+run_subm_clusters create_ligthhouse_coredns_deployment
```
vthapar, line 39: Won't this fail for operator coz it already installs it? We can check for deploytool and only do these for helm.

scripts/resources/dns-server.yaml
```diff
@@ -0,0 +1,105 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: submariner-lighthouse
```
vthapar, line 4: This is already created by helm. We create it for lighthouse-agent and can reuse for dns server

scripts/resources/coredns-cm.yaml
```diff
@@ -1,6 +1,9 @@
 apiVersion: v1
 data:
   Corefile: |
+    supercluster.local {
```
vthapar, line 4: shouldn't this be supercluster.local:53 ???

scripts/deploy
aswinsuryan: Add the check to see if config-map is already modified.

scripts/resources/coredns-cm.yaml
aswinsuryan: I added it to the config. If not added it will take 53 by default, but default config files define it explicitly.

PR #161, opened 2020-06-02
Add make flag `using=` to pass options

mkolesnik:
Use the flag from shipyard to pass options, it replaces deploytool make
flag since this is already provided by shipyard.


mkolesnik:
Thanks, I resolved it so you can squash and merge

PR #171, opened 2020-06-18
  Add ServiceImport CRD

aswinsuryan:
Add Service import to align the multicluster-sig proposal.

- An extra filed to ClusterStatus is added to hold the cluster IP.

- This will hold the pod IP list in case of headless services.


package/lighthouse-crd.yaml
```diff
@@ -1,12 +1,12 @@
...
 spec:
   group: lighthouse.submariner.io
   version: v2alpha1
   names:
-    kind: ServiceExport
-    plural: serviceexports
-    singular: serviceexport
+    kind: ServiceImport
+    plural: serviceimports
+    singular: serviceimport
```
tpantelis, line 11: Why are you replacing ServiceExport here?

vthapar: > Why are you replacing ServiceExport here?

This was stop-gap CRD used in e2e before helm+operator can install ServiceExport. Since they do, currently this is not being used. With ServiceImport, we will use this to install ServiceImport till changes are merged in helm+operator and then remove this file altogether.

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,108 @@ type ServiceExportList struct {
...
+	Spec ServiceImportSpec `json:"spec,omitempty"`
+	// +optional
+	Status ServiceImportStatus `json:"status,omitempty"`
+}
+
+// ServiceImportType designates the type of a ServiceImport
+type ServiceImportType string
+
+const (
+	// SuperclusterIP are only accessible via the Supercluster IP.
```
tpantelis, line 92:
```suggestion
	// Services are only accessible via their supercluster IP.
```

?

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,108 @@ type ServiceExportList struct {
...
+	// mycompany.com/my-custom-protocol.
+	// Field can be enabled with ServiceAppProtocol feature gate.
+	// +optional
+	AppProtocol *string `json:"appProtocol,omitempty"`
+
+	// The port that will be exposed by this service.
+	Port int32 `json:"port"`
+}
+
+//ClusterServiceInfo provides information about service in each cluster.
```
tpantelis, line 143:
```suggestion
// ClusterServiceInfo provides information about services in other clusters.
```

?

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,108 @@ type ServiceExportList struct {
...
+type ClusterServiceInfo struct {
+	//The unique identity of the cluster where the service is available. This field is
+	//not optional.
```
tpantelis, line 146:
```suggestion
	// The unique identity of the cluster where the service is available. This field is
	// not optional.
```

For readability consistency - also in other places.

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,109 @@ type ServiceExportList struct {
...
+	// +optional
+	IP string `json:"ip,omitempty"`
+	// +optional
+	Type ServiceImportType `json:"type"`
+	// +optional
+	SessionAffinity corev1.ServiceAffinity `json:"sessionAffinity"`
+	// +optional
+	SessionAffinityConfig *corev1.SessionAffinityConfig `json:"sessionAffinityConfig"`
+	// +optional
+	Items []ClusterServiceInfo `json:"clusterServiceInfo"`
```
mangelajo, line 111: I would make this part of the status instead, because it's something like the "Clusters" list that they propose -just more extended-, which is status dependent (will vary based on the status of the other clusters, etc..).

In fact you could remove this alltogether and put it into the ClusterStatus like:

```
// ClusterStatus contains service configuration mapped to a specific source cluster
type ClusterStatus struct {
 Cluster string `json:"cluster"`
 ServiceIPs []string `json:"serviceIPs,omitempty"`
}
```

And use "Cluster" instead of ClusterID to adhere to their standard :)

note the "omitempty", if not set, it's not even listed.

aswinsuryan: Moved it to ClusterStatus, it is a good one.

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,101 @@ type ServiceExportList struct {
...
+
+// ClusterStatus contains service configuration mapped to a specific source cluster
+type ClusterStatus struct {
+	Cluster string `json:"cluster"`
+
+	// The cluster IP of the service running in the cluster that is identified by the
+	// ClusterID field. In the case of headless services it will be list of pod IPs
+	// that backs the service.
+	// +optional
+	IPs []string `json:"ips,omitempty"`
```
tpantelis, line 159: Shouldn't this be singular, ie `IP string json:"ip,omitempty"` as the first sentence in the comment above indicates?

aswinsuryan: it will be one Ip if it is a cluster IP service, but if the service is headless, it would be  a list of IPs of the pods.

mangelajo: Yes this one needs to be plural, for one "fronted supercluster IP" (which we are not using yet..) there are multiple cluster specific IPs we need to know about.  We manage that by DNS with lighthouse, so we don't need the supercluster IP yet.

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -69,3 +69,101 @@ type ServiceExportList struct {
...
+	Clusters []ClusterStatus `json:"clusters"`
+}
+
+// ClusterStatus contains service configuration mapped to a specific source cluster
+type ClusterStatus struct {
+	Cluster string `json:"cluster"`
+
+	// The cluster IP of the service running in the cluster that is identified by the
+	// ClusterID field. In the case of headless services it will be list of pod IPs
+	// that backs the service.
```
tpantelis, line 157:
```suggestion
	// The IP(s) of the service running in the cluster.  In the case of a headless service, it is the list of pod IPs
	// that back the service.
```

PR #172, opened 2020-06-18
Change return code for Ipv6 queries

vthapar:
We return NotImplemented for IPv6 queries. Return NXDOMAIN instead which
is handled better by most applications like curl etc.


mangelajo:
@vthapar how did the tests go sending an empty AAAA response?

mangelajo:
I believe this can now be closed in favor of: https://github.com/submariner-io/lighthouse/pull/175 which was merged.

vthapar:
Fixed differently by https://github.com/submariner-io/lighthouse/pull/175

PR #175, opened 2020-06-19
Return empty response for AAAA queries

vthapar:


plugin/lighthouse/handler.go
```diff
@@ -26,8 +26,10 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
...
 	}
-
-	if state.QType() != dns.TypeA {
+	if state.QType() == dns.TypeAAAA {
+		// We only support TypeA
+		log.Debugf("Only TypeA queries are supported yet")
+	} else if state.QType() != dns.TypeA {
 		// We only support TypeA
 		log.Debugf("Only TypeA queries are supported yet")
 		return lh.nextOrFailure(state.Name(), ctx, w, r, dns.RcodeNotImplemented, "Only TypeA supported")
```
tpantelis, line 35:
```suggestion
	if state.QType() != dns.TypeA && state.QType() != dns.TypeAAAA {
	        msg = fmt.Sprintf("Query of type %d is not supported", state.QType())
		log.Debugf(msg)
		return lh.nextOrFailure(state.Name(), ctx, w, r, dns.RcodeNotImplemented, msg)
```

plugin/lighthouse/handler.go
```diff
@@ -55,6 +57,10 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 	}
 	serviceIp := ipList[0]
 
+	if state.QType() == dns.TypeAAAA {
+		return lh.emptyIpv6Response(state)
```
tpantelis, line 61:
```suggestion    
                log.Debugf("Returning empty response for TypeAAAA query")
		return lh.emptyIpv6Response(state)
```

vthapar: Done. Had to do manually coz of spacing.

PR #180, opened 2020-06-24
Less scripts, reusing Shipyard components

### LINKED ISSUE #167: Use shared targets from Shipyard
Validation, tests and other targets were migrated to Shipyard, use them instead of cloned scripts.

mkolesnik:
Removed the duplicate scripts and updated everything to use Shipyard shared scripts as much as we can.

Resolves #167


Makefile
```diff
@@ -22,17 +24,24 @@ endif
 
 # Targets to make
 
-build: vendor/modules.txt
-	./scripts/build-agent $(BUILD_ARGS)
+images: package/.image.lighthouse-agent
 	./scripts/build-coredns $(coredns) $(BUILD_ARGS)
 
-deploy: build clusters
+# Implicitly depend on the binary, since if it doesn't exist Shipyard won't find it
```
skitt, line 30:
```suggestion
# Explicitly depend on the binary, since if it doesn't exist Shipyard won't find it
```

PR #196, opened 2020-07-09
Fix globalnet e2e

vthapar:
E2E tests are not checking for globalnetIP correctly. To verify IP, we
use service object returned at time of creation. This will never have
globalnetIp as it is annotated later in an update. We should get the
latest service object before verifying.

### DIFF
--- a/test/e2e/discovery/service_discovery.go
+++ b/test/e2e/discovery/service_discovery.go
@@ -44,6 +44,7 @@ func RunServiceDiscoveryTest(f *lhframework.Framework) {
 
 	By(fmt.Sprintf("Creating a Nginx Service on %q", clusterBName))
 	nginxServiceClusterB := f.NewNginxService(framework.ClusterB)
+	f.AwaitGlobalnetIP(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 	f.NewServiceExport(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 
 	se := f.AwaitServiceExportStatusCondition(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
@@ -53,9 +54,15 @@ func RunServiceDiscoveryTest(f *lhframework.Framework) {
 	By(fmt.Sprintf("Creating a Netshoot Deployment on %q", clusterAName))
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 
+	if svc, err := f.GetService(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace); err == nil {
+		nginxServiceClusterB = svc
+		f.AwaitServiceImportIP(framework.ClusterA, framework.ClusterB, nginxServiceClusterB)
+	}
+
 	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, superclusterDomain, true)
 
 	f.DeleteServiceExport(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
+	f.AwaitServiceImportDelete(framework.ClusterA, framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 
 	f.DeleteService(framework.ClusterB, nginxServiceClusterB.Name)
 
@@ -78,6 +85,7 @@ func RunServiceDiscoveryLocalTest(f *lhframework.Framework) {
 
 	By(fmt.Sprintf("Creating a Nginx Service on %q", clusterBName))
 	nginxServiceClusterB := f.NewNginxService(framework.ClusterB)
+	f.AwaitGlobalnetIP(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 	f.NewServiceExport(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 
 	se := f.AwaitServiceExportStatusCondition(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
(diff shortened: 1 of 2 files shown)


test/e2e/framework/framework.go
```diff
@@ -90,3 +91,8 @@ func (f *Framework) DeleteServiceExport(cluster framework.ClusterIndex, name str
 	By("Waiting for mcs cleanup...")
 	time.Sleep(cleanupWait)
 }
+
+func (f *Framework) GetService(cluster framework.ClusterIndex, name string, namespace string) (*v1.Service, error) {
```
aswinsuryan, line 95: should we use wait and retry in case service is not yet created?

test/e2e/discovery/service_discovery.go
```diff
@@ -53,6 +53,10 @@ func RunServiceDiscoveryTest(f *lhframework.Framework) {
 	By(fmt.Sprintf("Creating a Netshoot Deployment on %q", clusterAName))
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 
+	if svc, err := f.GetService(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace); err == nil {
```
mangelajo, line 56: Don't you need to wait until service has the global IP for global IP ?

test/e2e/framework/framework.go
mangelajo: Or in case it globalIP is not set, but we know that we are testing global ip?

test/e2e/discovery/service_discovery.go
vthapar: > Don't you need to wait until service has the global IP for global IP ?

Yes. For that I need globalnetEnabled flag to be set in context. I discussed with Sridhar and will add that to shipyard code. I think it is common enough that we shouldn't duplicate in submariner and lighthouse both.

test/e2e/framework/framework.go
vthapar: That can be written by wrapping this around an AwaitUntil if needed. In this particular case we know that we've created the service. This will get replaced with AwaitUntilGlobalIp method once I have the globalnetflag check.

test/e2e/discovery/service_discovery.go
tpantelis: https://github.com/submariner-io/shipyard/issues/72 - it's on my plate for 0.5.0.  I can expedite it.

test/e2e/framework/framework.go
```diff
@@ -49,6 +52,12 @@ func createLighthouseClient(restConfig *rest.Config) *lighthouseClientset.Client
 	return clientSet
 }
 
+func setGlobalnetEnabled() {
+	listOptions := metav1.ListOptions{LabelSelector: "app=submariner-globalnet"}
+	podList, _ := framework.KubeClients[framework.ClusterB].CoreV1().Pods(framework.TestContext.SubmarinerNamespace).List(listOptions)
+	framework.TestContext.GlobalnetEnabled = podList != nil && len(podList.Items) > 0
+}
```
tpantelis, line 59: Do we still need this now that we're passing the CLI flag?

test/e2e/framework/framework.go
```diff
@@ -85,8 +94,68 @@ func (f *Framework) DeleteServiceExport(cluster framework.ClusterIndex, name str
...
+			return true, nil
+		}
+		return false, err
+	}, func(result interface{}) (bool, string, error) {
+		deleted := result.(bool)
+		return deleted, "", nil
+	})
+}
+
+func (f *Framework) AwaitAndGetGlobalnetIP(cluster framework.ClusterIndex, name string, namespace string) string {
```
tpantelis, line 144:
```suggestion
func (f *Framework) AwaitGlobalnetIP(cluster framework.ClusterIndex, name string, namespace string) string {
```

test/e2e/framework/framework.go
```diff
@@ -85,8 +94,68 @@ func (f *Framework) DeleteServiceExport(cluster framework.ClusterIndex, name str
...
+	si := LighthouseClients[targetCluster].LighthouseV2alpha1().ServiceImports(framework.TestContext.SubmarinerNamespace)
+	framework.AwaitUntil("retrieve ServiceImport", func() (interface{}, error) {
+		_, err := si.Get(siName, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}, func(result interface{}) (bool, string, error) {
+		deleted := result.(bool)
+		return deleted, "", nil
```
tpantelis, line 140:
```suggestion
		return result.(bool), "", nil
```

test/e2e/framework/framework.go
```diff
@@ -85,8 +94,68 @@ func (f *Framework) DeleteServiceExport(cluster framework.ClusterIndex, name str
...
+	if framework.TestContext.GlobalnetEnabled {
+		//TODO: Add wait for annotation
+		serviceIP = svc.Annotations[submarinerIpamGlobalIp]
+	} else {
+		serviceIP = svc.Spec.ClusterIP
+	}
+	siName := svc.Name + "-" + svc.Namespace + "-" + framework.TestContext.ClusterIDs[sourceCluster]
+	si := LighthouseClients[targetCluster].LighthouseV2alpha1().ServiceImports(framework.TestContext.SubmarinerNamespace)
+	By(fmt.Sprintf("Retrieving ServiceImport %s on %q", siName, framework.TestContext.ClusterIDs[targetCluster]))
+	siObj := framework.AwaitUntil("retrieve ServiceImport", func() (interface{}, error) {
```
tpantelis, line 116:
```suggestion
	return framework.AwaitUntil("retrieve ServiceImport", func() (interface{}, error) {
```

test/e2e/framework/framework.go
vthapar: > Do we still need this now that we're passing the CLI flag?

No, this will be cleaned up. Testing helm locally, just wanted to push to see if any operator specific issues.

PR #193, opened 2020-07-07
Replace MultiClusterService with ServiceImport

aswinsuryan:
The ServiceImport replaces the MultiClusterService.

Signed-off-by: Aswin Surayanarayanan <asuryana@redhat.com>

PR #197, opened 2020-07-09
Replace MultiClusterService with ServiceImport

aswinsuryan:
The ServiceImport replaces the MultiClusterService.


tpantelis:
What is this PR for? It looks the same as https://github.com/submariner-io/lighthouse/pull/193 which was merged

aswinsuryan:
> What is this PR for? It looks the same as #193 which was merged

This was raised to just to see if we could squash some of the commits. Seems like it is not trivial and hence closing.

PR #198, opened 2020-07-10
Remove Service Import CRD install for helm deployment

aswinsuryan:
The updated helm chart installs the ServiceImport CRD however the operator changes won't be available until the subctl version is updated in the shipyard docker base image. In the meantime, only install the CRD for an operator deployment.

### DIFF
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -17,11 +17,13 @@ function update_coredns_configmap() {
         sed 's/dnsip/'${CLUSTER_IP}'/g' scripts/resources/coredns-cm.yaml > /tmp/coredns-cm-${CLUSTER_IP}.yaml
         kubectl -n kube-system replace -f /tmp/coredns-cm-${CLUSTER_IP}.yaml
         kubectl -n kube-system describe cm coredns
-    fi
+    else 
+        run_subm_clusters install_service_import
+    fi 
 }
 
 function install_service_import() {
-    #TODO: Delete this once operator and helm changes are merged
+    #TODO: Delete this once operator changes are available
     kubectl apply -f ${DAPPER_SOURCE}/package/lighthouse-crd.yaml
     kubectl get crds | grep serviceimport
 }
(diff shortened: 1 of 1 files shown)


scripts/deploy
```diff
@@ -32,7 +34,8 @@ declare_kubeconfig
 import_image quay.io/submariner/lighthouse-agent
 import_image quay.io/submariner/lighthouse-coredns
 
-run_subm_clusters install_service_import
+echo ${deploytool}
+echo
```
tpantelis, line 38: Were these changes intentonal?

scripts/deploy
```diff
@@ -17,13 +17,15 @@ function update_coredns_configmap() {
         sed 's/dnsip/'${CLUSTER_IP}'/g' scripts/resources/coredns-cm.yaml > /tmp/coredns-cm-${CLUSTER_IP}.yaml
         kubectl -n kube-system replace -f /tmp/coredns-cm-${CLUSTER_IP}.yaml
         kubectl -n kube-system describe cm coredns
-    fi
+    else 
+        run_subm_clusters install_service_import
```
tpantelis, line 21: Why is `install_service_import` now conditional on the coredns CM?

aswinsuryan: The changes to install CRD is done via the following PRs 

Operator
https://github.com/submariner-io/submariner-operator/pull/550
Helm
https://github.com/submariner-io/submariner-charts/pull/51

While shipyard get  the latest from helm the operator changes are picked up only when we update subctl version in [1] is updated.
https://github.com/submariner-io/shipyard/blob/master/package/Dockerfile.shipyard-dapper-base#L47 

The "if" above identifies a helm based installation where core-dns map needs to be updated explicitly. So I added an else for doing things specific to operator installation. This shall be removed when the latest subctl is available.

tpantelis:
The updated helm chart installs the `ServiceImport` CRD however the operator changes won't be available until the subctl version is updated in the shipyard docker base image. In the meantime, only install the CRD for an operator deployment.

scripts/deploy
tpantelis, line 38: Are these unintentional local debugging changes?

aswinsuryan: Yes it was unintentional and removed it.

PR #210, opened 2020-07-16
Use GatewayStatus

### LINKED ISSUE #113: Don't return IPs of services in disconnected clusters
Use the Gateway status information to determine if a cluster is connected or not. If it isn't connected, don't return IPs of services running in that cluster.

vthapar:
Fixes #113


pkg/gateway/map.go
```diff
@@ -0,0 +1,27 @@
+package gateway
+
+import (
+	"sync/atomic"
+)
+
+type Map struct {
+	clustersMap atomic.Value
+}
+
```
tpantelis, line 10: I don’t think the `Map` struct really adds any value - it just wraps `Atomic.Value` so might as well use that directly.  Another idea is to make it an interface in the plugin package, ie:

```
type ClusterStatus interface {
    IsConnected(string clusterID), bool
}
```

The GW controller would implement it. This has the advantage of de-coupling the plugin from the GW in case we implement it in a different way in the future.

pkg/gateway/controller.go
```diff
@@ -0,0 +1,255 @@
...
+		return NewClientset
+	}
+
+	return func(c *rest.Config) (dynamic.Interface, error) {
+		return dynamic.NewForConfig(c)
+	}
+}
+
+func (c *Controller) Start(kubeConfig *rest.Config) error {
+	klog.Infof("Starting Gateways Controller")
```
tpantelis:
```suggestion
	klog.Infof("Starting Gateway status Controller")
```

pkg/gateway/controller.go
```diff
@@ -0,0 +1,255 @@
...
+	go c.runWorker()
+
+	return nil
+}
+
+func (c *Controller) Stop() {
+	close(c.stopCh)
+	c.queue.ShutDown()
+
+	klog.Infof("ServiceImport Controller stopped")
```
tpantelis:
```suggestion
	klog.Infof("Gateway status Controller stopped")
```

pkg/gateway/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return haStatus, connections, true
+}
+
+func (c *Controller) getClustersMap() map[string]bool {
+	clustersMap := c.clustersMap.Load()
+	if clustersMap != nil {
+		return clustersMap.(map[string]bool)
+	}
+	return make(map[string]bool)
```
tpantelis, line 230: The map should never be nil as it's initialized in the constructor. If it was then that would be a bug. Also returning a new map would hide the bug and result in the caller updating a local map. I suggest not checking and let it panic. ie

`return c.clustersMap.Load().(map[string]bool)`

pkg/gateway/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	connections, found, err := unstructured.NestedSlice(status, "connections")
+	if !found || err != nil {
+		klog.Errorf("connections field not found in %#v, err was: %v", status, err)
+		return haStatus, nil, false
+	}
+	return haStatus, connections, true
+}
+
+func (c *Controller) getClustersMap() map[string]bool {
```
tpantelis, line 225:
```suggestion
func (c *Controller) getClusterStatusMap() map[string]bool {
```

pkg/gateway/controller.go
```diff
@@ -0,0 +1,256 @@
...
+// Indirection hook for unit tests to supply fake client sets
+var NewClientset NewClientsetFunc
+
+type Controller struct {
+	newClientset NewClientsetFunc
+	informer     cache.Controller
+	store        cache.Store
+	queue        workqueue.RateLimitingInterface
+	stopCh       chan struct{}
+	clustersMap  atomic.Value
```
tpantelis, line 32:
```suggestion
	clusterStatusMap  atomic.Value
```

pkg/gateway/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	gatewayCheck bool
```
tpantelis, line 33:
```suggestion
	gatewayAvailable bool
```

pkg/gateway/controller.go
vthapar: > The map should never be nil as it's initialized in the constructor. If it was then that would be a bug. Also returning a new map would hide the bug and result in the caller updating a local map. I suggest not checking and let it panic. ie
> 
> `return c.clustersMap.Load().(map[string]bool)`

What about in case of UT?

tpantelis: Shouldn't UT call `NewController()`?

pkg/gateway/controller.go
```diff
@@ -0,0 +1,245 @@
...
+		if !ok {
+			klog.Errorf("Could not convert object tombstone %v to Unstructured", tombstone.Obj)
+			return
+		}
+	}
+
+	haStatus, _, _ := getGatewayStatus(resource)
+	if haStatus == "active" {
+		c.clusterStatusMap.Store(make(map[string]bool))
+	}
```
tpantelis, line 151: I think there's an edge case where this may be problematic. The normal case is where the active engine shuts down gracefully and deletes its `Gateway`.  However if it is killed or the delete fails then the `Gateway` will be stale. In this case, the engine has a stale `Gateway` check and thus the new engine will delete it later. This means we could get a deleted `Gateway` that has `haStatus == "active"` but isn't really the active `Gateway`. To handle this, we can keep track of the last `Gateway` resource that was created/updated in an `atomic.Value` via its UUID.  Here, if the the UUID matches the cached UUID then clear the map.

plugin/lighthouse/handler_test.go
```diff
@@ -43,9 +52,12 @@ func testWithoutFallback() {
 	)
 
 	BeforeEach(func() {
+		mockCs := new(MockClusterStatus)
+		mockCs.clusterId = clusterID
```
tpantelis, line 56: We should add a case where `IsConnected` returns false. Also a case where there's 2 clusters and one returns false and the other true.

pkg/gateway/controller.go
vthapar: Is it possible for stale gateway and new gateway to co-exist? as in, two gateways with `haStatus == "active"` ? If not, then it will eventually get overwritten and we should be fine. This logic will likely change once we get Active-ACctive HA.

tpantelis: > Is it possible for stale gateway and new gateway to co-exist? as in, two gateways with `haStatus == "active"` ? 
 
yes - new engine instance will publish its new active `Gateway` while the stale active `Gateway` will remain until the engine purges it.

If not, then it will eventually get overwritten and we should be fine. This logic will likely change once we get Active-Active HA.
> 
> EDIT: Also, we get updates every 5 seconds, so we will end up filling it with correct information.

True but there may be a 5 sec gap where no clusters are connected and user requests will fail. Unless I'm missing something...

tpantelis: Thinking some more, I'm wondering if we should do anything here, ie keep the last known state. Doesn't the engine leave the tunnels and network state as is on shutdown so as to maintain connectivity?  @sridhargaddam What do you think? Should we clear the cluster connectivity state or leave the last state when the `Gateway` is deleted, ie when the engine fails over.

vthapar: More I think about it, I like the idea of UUID. We will need to use it when we get Active-Active, so might as well get a head start on it.

mangelajo: I think it would be more stable to cache the last active gateway update connection list and use that to calculate which clusters are up or down. 

You could have a deleted/recreated gateway on a failover, or you could have duplicated active, while that's not true (at least yet)

Also consider that during those events it could be healthy to still resolve to what we knew with the hope that reconnection will happen quickly as it was before. So you allow the resolution, and then the connection will eventually succeed.

It's different to the case where a connection to a cluster goes down. (Please note that a failover on a remote cluster can also trigger this on other clusters)

I wonder if it would be helpful to (at some sprint) implement a filter in your algorithms to ignore spurious transitions like this (<x seconds) .

sridhargaddam: > Thinking some more, I'm wondering if we should do anything here, ie keep the last known state. Doesn't the engine leave the tunnels and network state as is on shutdown so as to maintain connectivity? @sridhargaddam What do you think? 

I'm working on preserving the tunnels (so that datapath is not disrupted) even when GatewayPod is restarted. So far could not get it working, as the pluto/charon process is getting killed as soon as GatewayPod is terminated. Looking into various options on how to get these processes running even when GatewayPod is terminated.

>Should we clear the cluster connectivity state or leave the last state when the `Gateway` is deleted, ie when the engine fails over.

Currently, as soon as GatewayPod is deleted, it WILL cleanup the Gateway object. **This is now properly handled with some recent fixes like clean termination of GatewayPod.** 
If we are able to successfully preserve the tunnels (even after the GatewayPod is terminated), then I think we should NOT delete the Gateway object anymore. Stale Gateway objects (based on the active leader election) should be handled either in route-agentPod or the new GatewayPod leader that is scheduled. This is a good topic to be discussed with the team.

@vthapar @tpantelis @mangelajo 
For now, I agree we can cache the last active gateway and use that to calculate which is up or down.

JYFI, it looks like we have bug in our implementation where connection status is not changing even when remoteGateway node is down.
https://github.com/submariner-io/submariner/issues/740

plugin/lighthouse/handler_test.go
vthapar: done

pkg/gateway/controller.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+	}
+
+	if newMap != nil {
+		klog.Infof("Updating the gateway status %#v ", newMap)
+		c.clusterStatusMap.Store(newMap)
+	}
+}
+
+func getGatewayStatus(obj *unstructured.Unstructured) (haStatus string, connections []interface{}, gwStatus bool) {
```
tpantelis, line 171: I wonder why the inter is forcing the use of named params here - it's not considered best practice in most cases.  `gwStatus` is unused - I'm surprised it didn't flag that 😄  Anyway just an observation - it's fine either way.

pkg/gateway/controller.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+		},
+		UpdateFunc: func(obj interface{}, new interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(new)
+			klog.V(2).Infof("GatewayStatus %q updated", key)
+			if err == nil {
+				c.queue.Add(key)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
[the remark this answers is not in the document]
vthapar: yeah, kept for debug

PR #227, opened 2020-07-28
build(coredns): build with external golang source code

SteveMattar:
use a different yet simpler approach for building lighthouse-coredns [#ref](https://coredns.io/2017/07/25/compile-time-enabling-or-disabling-plugins/#build-with-external-golang-source-code)
instead of cloning coredns repo and modify `plugin.cfg` and `go.mod` to import lighthouse
we now use lighthouse repo and import coredns as a module

Signed-off-by: Steve Mattar <smattar@redhat.com>


Makefile
```diff
@@ -8,6 +8,7 @@ include $(SHIPYARD_DIR)/Makefile.inc
 
 TARGETS := $(shell ls -p scripts | grep -v -e / -e deploy)
 CLUSTER_SETTINGS_FLAG = --cluster_settings $(DAPPER_SOURCE)/scripts/cluster_settings
+override BUILD_ARGS += $(shell source ${SCRIPTS_DIR}/lib/version; echo --ldflags \'-X main.VERSION=$${VERSION}\')
```
mkolesnik, line 11: This isnt needed as it's unused

SteveMattar: got it. I see it now

Makefile
```diff
@@ -24,15 +24,18 @@ endif
...
 package/.image.lighthouse-agent: bin/lighthouse-agent
 
+package/.image.lighthouse-coredns: bin/lighthouse-coredns
+
 bin/lighthouse-agent: vendor/modules.txt $(shell find pkg/agent)
 	${SCRIPTS_DIR}/compile.sh $@ pkg/agent/main.go
 
+bin/lighthouse-coredns: vendor/modules.txt $(shell find pkg/coredns)
+	${SCRIPTS_DIR}/compile.sh $@ pkg/coredns/main.go
 deploy: images clusters
```
skitt, line 39: Please add a blank line here ;-).

SteveMattar: Thanks @skitt. yes, debian was the image we were using so far. I will open another PR for ubi as it needs its own tests. what do you think?

pkg/coredns/main.go
```diff
@@ -0,0 +1,96 @@
+package main
```
vthapar, line 1: I believe this file is taken from coredns repo, right? Would be good to add some comments explaining that:
1. Source repo
2. Git tag/branch

This is to make it easy to track which version of coredns/main.go we're using and to make sure it aligns with versions we using in plugin, which currently is 1.5.2

SteveMattar:
> A minor tweak. It would be nice to use the UBI minimal image instead of the slim Debian image, but since the latter is the base image for CoreDNS we’d probably need more testing.

Thanks @skitt. yes, debian was the image we were using so far. I will open another PR for ubi as it needs its own tests. what do you think?

skitt:
> Thanks @skitt. yes, debian was the image we were using so far. I will open another PR for ubi as it needs its own tests. what do you think?

That would be perfect, thanks!

skitt:
I see that in CoreDNS, the list of directives is produced by a generator; it might be worth (as a separate PR) looking into whether we could use that too...

pkg/coredns/main.go
SteveMattar: Thanks @vthapar I'll add the documentation reference as a comment with some explanation.

For (2) I'm importing coredns
https://github.com/submariner-io/lighthouse/blob/2b9037545abd2c1388f1e45e9d0d1087b2da290a/pkg/coredns/main.go#L44-L45

which is the same version specified in `go.mod`
https://github.com/submariner-io/lighthouse/blob/2b9037545abd2c1388f1e45e9d0d1087b2da290a/go.mod#L7

is this what you meant?

skitt: @vthapar the list of plugins doesn’t match the list in CoreDNS 1.5.2 exactly; is that normal? We’re missing `dns64`, `acl`, `transfer`, `azure`, `clouddns`, and `kubernetes`, and we’re adding `federation` and `sign`.

package/Dockerfile.lighthouse-coredns
```diff
@@ -0,0 +1,9 @@
+FROM debian:stable-slim
+
+RUN apt-get update && apt-get -uy upgrade
+RUN apt-get -y install ca-certificates && update-ca-certificates
```
mkolesnik, line 4:
```suggestion
RUN apt-get update && apt-get -y install ca-certificates && update-ca-certificates
```

package/Dockerfile.lighthouse-coredns
```diff
@@ -0,0 +1,9 @@
+FROM debian:stable-slim
```
mkolesnik, line 1: Would it be possible to use ubi8 here to be in line with the rest of the images we build?

pkg/coredns/main.go
SteveMattar: @skitt I used `lighthouse-coredns -plugins` to compare the list on both versions

package/Dockerfile.lighthouse-coredns
SteveMattar: Thanks @mkolesnik

package/Dockerfile.lighthouse-coredns
SteveMattar: yeah we already discussed it I will open another PR for that

pkg/coredns/main.go
skitt: @SteveMattar yes, this isn’t anything wrong with your changes, I’m just curious why the list is different (which would make it less useful to give the original reference for the `main.go` file).

vthapar: Looks like I was sitting on this comment. @skitt as discussed, this is okay. we will possible prune this list later as we're not really using most of these. Let us get this merged, pruning can be done later once aswin's changes to upgrade to newer version of coredns come in.

PR #228, opened 2020-07-29
Bump to Kubernetes 1.17.0

skitt:
... to match Shipyard.


skitt:
The version of CoreDNS we’d need for 1.17 is 1.6.7 or later, but it’s affected by the `gnostic` breakage. https://github.com/submariner-io/lighthouse/pull/227 should improve things.

PR #229, opened 2020-07-29
Rename "supercluster" to "clusterset"

skitt:
The latter is the name chosen by the multi-cluster SIG.

### DIFF
--- a/pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
+++ b/pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
@@ -36,11 +36,11 @@ type ServiceExportConditionType string
 const (
 	// ServiceExportInitialized means the service export has been noticed
 	// by the controller, has passed validation, has appropriate finalizers
-	// set, and any required supercluster resources like the IP have been
+	// set, and any required clusterset resources like the IP have been
 	// reserved
 	ServiceExportInitialized ServiceExportConditionType = "Initialized"
 	// ServiceExportExported means that the service referenced by this
-	// service export has been synced to all clusters in the supercluster
+	// service export has been synced to all clusters in the clusterset
 	ServiceExportExported ServiceExportConditionType = "Exported"
 )
 
@@ -74,7 +74,7 @@ type ServiceExportList struct {
 //+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +genclient:noStatus
 
-// ServiceImport describes a service imported from clusters in a supercluster.
+// ServiceImport describes a service imported from clusters in a clusterset.
 type ServiceImport struct {
 	metav1.TypeMeta `json:",inline"`
 	// +optional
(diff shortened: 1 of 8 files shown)


test/e2e/discovery/service_discovery.go
```diff
@@ -61,14 +61,14 @@ func RunServiceDiscoveryTest(f *lhframework.Framework) {
...
-	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, superclusterDomain, true)
+	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, clustersetDomain, true)
 
 	f.DeleteServiceExport(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 	f.AwaitServiceImportDelete(framework.ClusterA, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
 
 	f.DeleteService(framework.ClusterB, nginxServiceClusterB.Name)
 
-	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, superclusterDomain, false)
+	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, clustersetDomain, false)
```
aswinsuryan, line 71: Can we try both "clusterset" and "supercluster" in verifyServiceIpWithDig and revert it once the operator/helm patch goes in?

skitt: Yes, done!

scripts/resources/coredns-cm.yaml
```diff
@@ -1,3 +1,3 @@
-    supercluster.local:53 {
+    clusterset.local:53 {
```
aswinsuryan, line 1: This may create a mis-match as helm configures the Lighthouse coreDNS authoritative for supercluster.local but only the requests to clusterset will be forwarded to it  by default coredns with this config,  Two entries as below should solve it.

    clusterset.local:53 {
        forward . dnsip
    }
    supercluster.local:53 {
        forward . dnsip
    }

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
```diff
@@ -89,7 +89,7 @@ type ServiceImport struct {
 type ServiceImportType string
 
 const (
-	// Services are only accessible via their supercluster IP.
+	// Services are only accessible via their clusterset IP.
```
vthapar, line 92: May need to change this too. I've sent mail to sig for clarification.

scripts/resources/coredns-cm.yaml
vthapar: Agree. I think we should do it in two steps, even with helm and operator:
1. Add `clusterset`. We'll have both `supercluster` and `clusterset` for 0.6.0
2. Remove 'supercluster`.

This will allow all users and test folks to transition to clusterset without breaking anything.

pkg/apis/lighthouse.submariner.io/v2alpha1/types.go
skitt: Good point, I’ll switch the PR back to draft until we get confirmation from the SIG.

mangelajo (dismissed):

vthapar (approved):

dfarrell07 (dismissed):

mangelajo (approved):

test/e2e/discovery/service_discovery.go
mangelajo: @skitt oops, I think the helm patch only configured the new clusterset, and not the "compatibility" one, so it will fail.

skitt: @mangelajo what will fail? If you’re referring to the e2e tests, they should be OK; the tests pass if either one of the domains is available.

Issue #114, opened 2020-04-29
Add round robin selection to IPs returned by lighthouse plugin


vthapar:
Modify lighthouse plugin to use round robin selection from list of IPs for a given multi cluster service.

danibachar:
On it 👍

nyechiel:
@manosnoam @pinikomarov code is node, ready for testing

manosnoam:
@vthapar  Can you tell me how to reproduce ?

@nyechiel I think we should add this test to E2E.

danibachar:
@manosnoam see - [this issue](https://github.com/submariner-io/lighthouse/issues/235)

manosnoam:
Opened test request:
https://github.com/submariner-io/lighthouse/issues/340

PR #231, opened 2020-08-02
Service Ip Selection - Round Robin

danibachar:
https://github.com/submariner-io/lighthouse/issues/114 - Signed-off-by: I <danibachar89@gmail.com>
@mkolesnik @vthapar @mangelajo
Notes and questions:

1) Is there a chance that we will get the same cluster more than once? in func (m *Map) Put(serviceImport *lighthousev2a1.ServiceImport) in the serviceImport.Status.Clusters object?
In addition, I saw that headless services can have a list of ips pointing to their pods, should accommodate that case?
2) I was thinking of how to apply the RR mechanism in the best way, and because we don't deal with thousands of clusters I thought that to implement it the way I did - each loop run will cause max of len(queue) increments which will yield a read lock.
Two more possible solutions I came up with:
Get before loop - Update after -> could cause race with the counting
Get before the loop (will cause an inc) and just iterate without updating at all - each lookup will cause 1 increment
3) I'm using the clasterInfo by value and not pointers, hope that is ok. It allows me to simplify the code and not worry about clearing memory on deletion. on the other hand, it feels a bit out of the way we are writing code in Golang


pkg/serviceimport/map.go
```diff
@@ -2,21 +2,53 @@ package serviceimport
...
+	Ip     string
+	Name   string
+	Weight uint64
+}
+
 type serviceInfo struct {
 	key         string
 	clusterInfo map[string]string
 	ipList      []string
+	queue       []clusterInfo // clusters queue [each could have an 1 vertual ip or an ip pool of nodes (headless)]
```
tpantelis, line 20:
```suggestion
	rrQueue       []clusterInfo // clusters queue [each could have an 1 vertual ip or an ip pool of nodes (headless)]
```

pkg/serviceimport/map.go
```diff
@@ -2,21 +2,53 @@ package serviceimport
...
+	m.RLock()
+	defer m.RUnlock()
+
+	if val, ok := m.svcMap[keyFunc(namespace, name)]; ok {
+		atomic.AddUint64(&val.rrCount, 1)
+		return val.rrCount, true
+	}
+
+	return 0, false
+}
```
tpantelis, line 50: I think it would be ideal to not expose the RR implementation to callers. I suggest to define a function to select an IP  using a passed function to check clusters:

```
    func (m *Map) SelectIP(namespace, name string, checkCluster func(string) bool) (string, bool) {
	queue, counter := func() ([]clusterInfo, *uint64) {
		m.RLock()
		defer m.RUnlock()

		si, ok := m.svcMap[keyFunc(namespace, name)]
		if !ok {
			return nil, nil
		}

		return si.rrQueue, &si.rrCount
	}()

	queueLength := len(queue)
	for i := 0; i < queueLength; i++ {
		c := atomic.LoadUint64(counter)
		atomic.AddUint64(counter, 1)

		info := queue[c%uint64(queueLength)]
		if checkCluster(info.Name) {
			return info.Ip, true
		}
	}

	return "", false
    }


```

pkg/serviceimport/map.go
```diff
@@ -2,21 +2,53 @@ package serviceimport
...
+type clusterInfo struct {
+	Ip     string
+	Name   string
+	Weight uint64
+}
+
 type serviceInfo struct {
 	key         string
 	clusterInfo map[string]string
 	ipList      []string
```
tpantelis, line 19: This is no longer needed.

pkg/serviceimport/map.go
```diff
@@ -2,21 +2,53 @@ package serviceimport
...
 
+type clusterInfo struct {
+	Ip     string
+	Name   string
+	Weight uint64
+}
+
 type serviceInfo struct {
 	key         string
 	clusterInfo map[string]string
```
tpantelis, line 18:
```suggestion
	clusterIPs map[string]string
```

pkg/serviceimport/map.go
danibachar: BTW - 2 points:
1) Eventually, I think we would need to get the loadbalancing logic out of the Map object
2) This will require to change several test cases... see the new commits (:

pkg/serviceimport/controller_test.go
```diff
@@ -118,26 +130,153 @@ func testLifecycleNotifications() {
...
+
+			Eventually(func() bool {
+				first, _ := controller.serviceImports.SelectIP(namespace1, service1, mockCs.IsConnected)
+				second, _ := controller.serviceImports.SelectIP(namespace1, service1, mockCs.IsConnected)
+				third, _ := controller.serviceImports.SelectIP(namespace1, service1, mockCs.IsConnected)
+				return (first == serviceIP) && (second == serviceIP2) && (third == serviceIP3)
+			}).Should(BeTrue())
+		})
+	})
+}
```
tpantelis, line 260: I think these tests should really be in a separate file specific to `serviceimport.Map`.  But we can leave it as is for now and I can push a patch afterwards so we can get this one in.

pkg/serviceimport/controller_test.go
```diff
@@ -118,26 +130,153 @@ func testLifecycleNotifications() {
...
 		namespace := expected.Annotations["origin-namespace"]
-		ipList, ok := controller.serviceImports.GetIps(namespace, name)
-		if ok {
-			return ipList
-		}
-		return nil
-	}).Should(Equal([]string{expected.Status.Clusters[0].IPs[0]}))
+		selectedIp, _ := controller.serviceImports.SelectIP(namespace, name, m.IsConnected)
+		return selectedIp
+	}).Should(Equal(expected.Status.Clusters[0].IPs[0]))
```
tpantelis, line 268: The controller is only responsible for populating the Map via Put/Remove therefore it really only needs to verify those function were called properly, ie I don't think we should be testing `SelectIP` here.  I know that `GetIps` is now removed - we can either access the internal `svcMap` directly or mock Put/Remove and use a chan to capture and verify the `ServiceImport`.  I think I prefer the latter. But again we can leave it as is for now.

danibachar: @tpantelis thanks Tom,  I'll change the tests here, and separate the test files into 2 after we merge...
I'm wondering if I need to have a separate file for the `serviceInfo` object like the Map object, with its own constructor and functions... for now I added it in the map.go file

tpantelis:
The DCO check is failing  - something strange with the Signed-Off...

PR #242, opened 2020-08-06
Initial support for headless services

### LINKED ISSUE #117: Add support for headless services
With "headless" service, instead of a single Service IP with load-balancing, the platform should return the IPs of associated Pods. This allows to interact directly with the Pods instead of a proxy. 

An headless services can be configured by explicitly specifying "None" for the clusterIP (.spec.clusterIP), and can be utilized with or without selectors: https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

vthapar:
This adds limited initial support for headless services.

TBD:
 * Allow creation of headless service after service export
 * Support for stateful sets
 * Conflict resolution for service type between clusters
 * E2E

Fixes #117

Reference: https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -201,7 +201,10 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 	}
 
 	if a.globalnetEnabled && svcType == lighthousev2a1.Headless {
-		klog.Errorf("Headless Services not supported with globalnet yet")
+		klog.Infof("Headless Services not supported with globalnet yet")
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+			corev1.ConditionFalse, invalidServiceType, "Headless Services not supported with Globalnet IP")
+
 		return nil, false
 	}
 
@@ -219,16 +222,26 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		Type: svcType,
 	}
 
+	ips := a.getIPsForService(svc, svcType)
+
+	if ips == nil {
+		// Failed to get ips for some reason, requeue
+		return nil, true
+	}
+
 	serviceImport.Status = lighthousev2a1.ServiceImportStatus{
 		Clusters: []lighthousev2a1.ClusterStatus{
 			{
 				Cluster: a.clusterID,
-				IPs:     a.getIpsForService(svc, svcType),
+				IPs:     ips,
 			},
 		},
 	}
 
 	if svcType == lighthousev2a1.SuperclusterIP {
+		/* We also store the clusterIP in an annotation as an optimization to recover it in case the IPs are
+		cleared out when here's no backing Endpoint pods.
+		*/
 		serviceImport.Annotations[clusterIP] = serviceImport.Status.Clusters[0].IPs[0]
 	}
 
@@ -303,22 +316,33 @@ func (a *Controller) endpointToRemoteServiceImport(obj runtime.Object, op syncer
 	serviceImport, err := a.getServiceImport(svcExport)
 
 	if err != nil {
+		if !errors.IsNotFound(err) {
+			klog.Errorf("Unable to get ServiceImport for %#v: %v", ep, err)
+		}
+
 		// Requeue
 		return nil, true
 	}
 
 	if serviceImport.Spec.Type == lighthousev2a1.Headless && a.globalnetEnabled {
-		klog.Errorf("Headless Services not supported with globalnet yet")
 		return nil, false
 	}
 
-	oldStatus := serviceImport.Status.DeepCopy()
-	ipList := getIpsFromEndpoint(ep)
+	ipList := getIPsFromEndpoint(ep)
 
 	if serviceImport.Spec.Type == lighthousev2a1.SuperclusterIP && len(ipList) > 0 {
-		// Once we cache service and related objecs, this won't be needed
+		/*
+			When there's no healthy pods, the IPs in the Endpoint will become empty and
+			thus we also clear the ServiceImport IPs to avoid clients sending a request
+			to a cluster which has no active pods to handle the request.
+
+			Once at least one backing Endpoint pod becomes healthy, we need to re-establish
+			the ServiceImport IPs from the previously cached clusterIP annotation.
+		*/
 		ipList = []string{serviceImport.Annotations[clusterIP]}
 	}
+
+	oldStatus := serviceImport.Status.DeepCopy()
 
 	serviceImport.Status = lighthousev2a1.ServiceImportStatus{
 		Clusters: []lighthousev2a1.ClusterStatus{
@@ -398,7 +422,7 @@ func serviceExportConditionEqual(c1, c2 *lighthousev2a1.ServiceExportCondition) 
 func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *lighthousev2a1.ServiceImport {
 	return &lighthousev2a1.ServiceImport{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: svcExport.Name + "-" + svcExport.Namespace + "-" + a.clusterID,
+			Name: a.getObjectNameWithClusterId(svcExport.Name, svcExport.Namespace),
 			Annotations: map[string]string{
 				originName:      svcExport.Name,
 				originNamespace: svcExport.Namespace,
@@ -413,7 +437,7 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
 }
 
 func (a *Controller) getServiceImport(svcExport *lighthousev2a1.ServiceExport) (*lighthousev2a1.ServiceImport, error) {
-	siName := svcExport.Name + "-" + svcExport.Namespace + "-" + a.clusterID
+	siName := a.getObjectNameWithClusterId(svcExport.Name, svcExport.Namespace)
 	serviceImport, err := a.lighthouseClient.LighthouseV2alpha1().ServiceImports(a.namespace).Get(siName, metav1.GetOptions{})
 
 	if err != nil {
@@ -423,7 +447,7 @@ func (a *Controller) getServiceImport(svcExport *lighthousev2a1.ServiceExport) (
 	return serviceImport, nil
 }
 
-func (a *Controller) getIpsForService(service *corev1.Service, siType lighthousev2a1.ServiceImportType) []string {
+func (a *Controller) getIPsForService(service *corev1.Service, siType lighthousev2a1.ServiceImportType) []string {
 	if siType == lighthousev2a1.SuperclusterIP {
 		mcsIp := getGlobalIpFromService(service)
 		if mcsIp == "" {
@@ -435,14 +459,22 @@ func (a *Controller) getIpsForService(service *corev1.Service, siType lighthouse
 
 	endpoint, err := a.kubeClientSet.CoreV1().Endpoints(service.Namespace).Get(service.Name, metav1.GetOptions{})
 	if err != nil {
-		klog.V(log.DEBUG).Infof("Endpoints for svc  (%s/%s) not found: %v", service.Namespace, service.Name, err)
+		if !errors.IsNotFound(err) {
+			klog.Errorf("Error obtaining Endpoint for Service  (%s/%s): %v", service.Namespace, service.Name, err)
+			return nil
+		}
+
 		return make([]string, 0)
 	}
 
-	return getIpsFromEndpoint(endpoint)
+	return getIPsFromEndpoint(endpoint)
 }
 
-func getIpsFromEndpoint(endpoint *corev1.Endpoints) []string {
+func (a *Controller) getObjectNameWithClusterId(name, namespace string) string {
+	return name + "-" + namespace + "-" + a.clusterID
+}
+
+func getIPsFromEndpoint(endpoint *corev1.Endpoints) []string {
 	ipList := make([]string, 0)
 
 	for _, eps := range endpoint.Subsets {
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -11,9 +11,11 @@ import (
 	lighthousev2a1 "github.com/submariner-io/lighthouse/pkg/apis/lighthouse.submariner.io/v2alpha1"
 	lighthouseClientset "github.com/submariner-io/lighthouse/pkg/client/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
@@ -26,8 +28,13 @@ import (
 const (
 	submarinerIpamGlobalIp = "submariner.io/globalIp"
 	serviceUnavailable     = "ServiceUnavailable"
+	invalidServiceType     = "UnsupportedServiceType"
+	clusterIP              = "cluster-ip"
 	originName             = "origin-name"
 	originNamespace        = "origin-namespace"
+	labelSourceName        = "lighthouse.submariner.io/sourceName"
+	labelSourceNamespace   = "lighthouse.submariner.io/sourceNamespace"
+	labelSourceCluster     = "lighthouse.submariner.io/sourceCluster"
 )
 
 var MaxExportStatusConditions = 10
@@ -67,17 +74,19 @@ func NewWithDetail(spec *AgentSpecification, syncerConf *broker.SyncerConfig, re
 	newSyncer func(*broker.SyncerConfig) (*broker.Syncer, error)) (*Controller, error) {
 	agentController := &Controller{
 		clusterID:        spec.ClusterID,
+		namespace:        spec.Namespace,
 		globalnetEnabled: spec.GlobalnetEnabled,
 		kubeClientSet:    kubeClientSet,
 		lighthouseClient: lighthouseClient,
 	}
 
 	svcExportResourceConfig := broker.ResourceConfig{
-		LocalSourceNamespace:  metav1.NamespaceAll,
-		LocalResourceType:     &lighthousev2a1.ServiceExport{},
-		LocalTransform:        agentController.serviceExportToRemoteServiceImport,
-		LocalOnSuccessfulSync: agentController.onSuccessfulServiceImportSync,
-		BrokerResourceType:    &lighthousev2a1.ServiceImport{},
+		LocalSourceNamespace:      metav1.NamespaceAll,
+		LocalResourceType:         &lighthousev2a1.ServiceExport{},
+		LocalTransform:            agentController.serviceExportToRemoteServiceImport,
+		LocalOnSuccessfulSync:     agentController.onSuccessfulServiceImportSync,
+		BrokerResourceType:        &lighthousev2a1.ServiceImport{},
+		BrokerResourcesEquivalent: agentController.serviceImportEquivalent,
 	}
 
 	syncerConf.Scheme = scheme
@@ -108,6 +117,23 @@ func NewWithDetail(spec *AgentSpecification, syncerConf *broker.SyncerConfig, re
 		return nil, err
 	}
 
+	agentController.endpointSyncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:                "Endpoint events",
+		SourceClient:        localClient,
+		SourceNamespace:     metav1.NamespaceAll,
+		Direction:           syncer.LocalToRemote,
+		RestMapper:          restMapper,
+		Federator:           agentController.serviceExportSyncer.GetBrokerFederator(),
+		ResourceType:        &corev1.Endpoints{},
+		Transform:           agentController.endpointToRemoteServiceImport,
+		ResourcesEquivalent: agentController.endpointEquivalent,
+		Scheme:              scheme,
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
 	return agentController, nil
 }
 
@@ -128,12 +154,20 @@ func (a *Controller) Start(stopCh <-chan struct{}) error {
 		return err
 	}
 
+	if err := a.endpointSyncer.Start(stopCh); err != nil {
+		return err
+	}
+
 	klog.Info("Lighthouse agent syncer started")
 
 	return nil
 }
 
 func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op syncer.Operation) (runtime.Object, bool) {
+	if op == syncer.Update {
+		return nil, false
+	}
+
 	svcExport := obj.(*lighthousev2a1.ServiceExport)
 	serviceImport := a.newServiceImport(svcExport)
 
@@ -156,6 +190,21 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		return nil, true
 	}
 
+	svcType, ok := getServiceImportType(svc)
+
+	if !ok {
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+			corev1.ConditionFalse, invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
+		klog.Errorf("Service type %q not supported", svc.Spec.Type)
+
+		return nil, false
+	}
+
+	if a.globalnetEnabled && svcType == lighthousev2a1.Headless {
+		klog.Errorf("Headless Services not supported with globalnet yet")
+		return nil, false
+	}
+
 	if a.globalnetEnabled && getGlobalIpFromService(svc) == "" {
 		klog.V(log.DEBUG).Infof("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
 
@@ -167,18 +216,38 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 	}
 
 	serviceImport.Spec = lighthousev2a1.ServiceImportSpec{
-		Type: lighthousev2a1.SuperclusterIP,
+		Type: svcType,
 	}
+
 	serviceImport.Status = lighthousev2a1.ServiceImportStatus{
 		Clusters: []lighthousev2a1.ClusterStatus{
-			a.newClusterStatus(svc, a.clusterID),
+			{
+				Cluster: a.clusterID,
+				IPs:     a.getIpsForService(svc, svcType),
+			},
 		},
+	}
+
+	if svcType == lighthousev2a1.SuperclusterIP {
+		serviceImport.Annotations[clusterIP] = serviceImport.Status.Clusters[0].IPs[0]
 	}
 
 	a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
 		corev1.ConditionTrue, "AwaitingSync", "Awaiting sync of the ServiceImport to the broker")
 
 	return serviceImport, false
+}
+
+func getServiceImportType(service *corev1.Service) (lighthousev2a1.ServiceImportType, bool) {
+	if service.Spec.Type != "" && service.Spec.Type != corev1.ServiceTypeClusterIP {
+		return "", false
+	}
+
+	if service.Spec.ClusterIP == corev1.ClusterIPNone {
+		return lighthousev2a1.Headless, true
+	}
+
+	return lighthousev2a1.SuperclusterIP, true
 }
 
 func (a *Controller) onSuccessfulServiceImportSync(synced runtime.Object, op syncer.Operation) {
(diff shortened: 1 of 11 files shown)


plugin/lighthouse/handler_test.go
```diff
@@ -290,11 +308,12 @@ func testClusterStatus() {
 			mockCs.clusterStatusMap[clusterID] = false
 			mockCs.clusterStatusMap[clusterID2] = false
 		})
-		It("should return RcodeServerFailure", func() {
+		It("should return empty response (NODATA)", func() {
 			executeTestCase(lh, rec, test.Case{
-				Qname: service1 + "." + namespace1 + ".svc.cluster.local.",
-				Qtype: dns.TypeA,
-				Rcode: dns.RcodeServerFailure,
+				Qname:  service1 + "." + namespace1 + ".svc.cluster.local.",
```
mangelajo, line 313: Why are you testing cluster.local instead of multicluster/supercluster/etc?

mangelajo: Same question for others

vthapar: plugin code doesn't really care as `supercluster.local` is more of a configuration parameter. Since this is handler UT, configuration doesn't apply to it. Agree that should change this, was waiting for `clusterset` to be formalized before making this change.

pkg/agent/controller/agent.go
```diff
@@ -288,10 +375,67 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
...
+	serviceImports, err := a.lighthouseClient.LighthouseV2alpha1().ServiceImports(corev1.NamespaceAll).List(siListOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(serviceImports.Items) > 1 {
+		return nil, fmt.Errorf("Found multiple ServiceImports with local cluster ID.")
+	}
+
+	return &serviceImports.Items[0], nil
```
tpantelis, line 405: `serviceImports.Items` could be empty.

But instead of using a label selector for lookup, why not  lookup by Name as constructed on L373 (and put it in a `serviceNameFrom(svcExport *lighthousev2a1.ServiceExport)` function)? The same label fields you're using for lookup are also encoded in the Name. This code would be simplified to use `Get` and not have to check for multiple.

pkg/agent/controller/agent.go
```diff
@@ -219,6 +269,55 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+		// some other error. Log and requeue
+		klog.Errorf("Unable to get ServiceExport for %#v: %v", ep, err)
+		return nil, true
+	}
+
+	serviceImport, err := a.getServiceImport(svcExport)
+
+	if err != nil {
+		// Log and requeue
+		klog.Errorf("Failed to get ServiceImport for %v", svcExport)
```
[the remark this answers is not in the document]
tpantelis: I don't think we should log an error if the `ServiceImport` isn't found.

pkg/agent/controller/agent.go
```diff
@@ -219,6 +269,55 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+		return nil, true
+	}
+
+	if serviceImport.Spec.Type == lighthousev2a1.Headless && a.globalnetEnabled {
+		klog.Errorf("Headless Services not supported with globalnet yet")
+		return nil, false
+	}
+
+	oldStatus := serviceImport.Status.DeepCopy()
```
tpantelis, line 297: Move this after L302.

pkg/agent/controller/agent.go
```diff
@@ -219,6 +269,55 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+	ipList := getIpsFromEndpoint(ep)
+
+	if serviceImport.Spec.Type == lighthousev2a1.SuperclusterIP && len(ipList) > 0 {
+		return nil, false
+	}
```
tpantelis, line 302: I'm not clear on this check. Shouldn't this code only run for `Headless`? For `SuperclusterIP` we use the service's `ClusterIP` (or GN IP) for the `ServiceImport` IP(s). With this check, we would overwrite the IPs with an empty list.

pkg/agent/controller/agent.go
```diff
@@ -288,10 +375,67 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
...
+
+	if len(serviceImports.Items) > 1 {
+		return nil, fmt.Errorf("Found multiple ServiceImports with local cluster ID.")
+	}
+
+	return &serviceImports.Items[0], nil
+}
+
+func (a *Controller) getIpsForService(service *corev1.Service, siType lighthousev2a1.ServiceImportType) []string {
```
tpantelis, line 408:
```suggestion
func (a *Controller) getIPsForService(service *corev1.Service, siType lighthousev2a1.ServiceImportType) []string {
```

pkg/agent/controller/agent.go
```diff
@@ -288,10 +375,67 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
...
+	endpoint, err := a.kubeClientSet.CoreV1().Endpoints(service.Namespace).Get(service.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Endpoints for svc  (%s/%s) not found: %v", service.Namespace, service.Name, err)
+		return make([]string, 0)
+	}
+
+	return getIpsFromEndpoint(endpoint)
+}
+
+func getIpsFromEndpoint(endpoint *corev1.Endpoints) []string {
```
tpantelis, line 427:
```suggestion
func getIPsFromEndpoint(endpoint *corev1.Endpoints) []string {
```

plugin/lighthouse/handler_test.go
```diff
@@ -269,6 +271,22 @@ func testClusterStatus() {
 		})
 	})
 
+	When("service is in two connected clusters and one has no ip", func() {
```
tpantelis, line 274:
```suggestion
	When("service is in two connected clusters and one has no IP", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -269,6 +271,22 @@ func testClusterStatus() {
...
+		JustBeforeEach(func() {
+			lh.serviceImports.Put(newServiceImport(namespace1, service1, clusterID, []string{}, lighthousev2a1.SuperclusterIP))
+		})
+		It("should succeed and write an A record response from cluster with available ip", func() {
```
tpantelis, line 278:
```suggestion
		It("should succeed and write an A record response with the available IP", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -305,11 +324,98 @@ func testClusterStatus() {
...
+		lh = &Lighthouse{
+			Zones:          []string{"cluster.local."},
+			serviceImports: serviceimport.NewMap(),
+			clusterStatus:  mockCs,
+		}
+
+		rec = dnstest.NewRecorder(&test.ResponseWriter{})
+	})
+
+	When("headless service has no ips", func() {
```
tpantelis, line 357:
```suggestion
	When("headless service has no IPs", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -305,11 +324,98 @@ func testClusterStatus() {
...
+			executeTestCase(lh, rec, test.Case{
+				Qname:  service1 + "." + namespace1 + ".svc.cluster.local.",
+				Qtype:  dns.TypeA,
+				Rcode:  dns.RcodeSuccess,
+				Answer: []dns.RR{},
+			})
+		})
+	})
+
+	When("headless service has one ip", func() {
```
tpantelis, line 371:
```suggestion
	When("headless service has one IP", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -305,11 +324,98 @@ func testClusterStatus() {
...
-				Rcode: dns.RcodeServerFailure,
+				Rcode: dns.RcodeSuccess,
+				Answer: []dns.RR{
+					test.A(service1 + "." + namespace1 + ".svc.cluster.local.    0    IN    A    " + serviceIP),
+				},
+			})
+		})
+	})
+
+	When("headless service has two ips", func() {
```
tpantelis, line 387:
```suggestion
	When("headless service has two IPs", func() {
```

pkg/agent/controller/agent.go
```diff
@@ -156,6 +180,17 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		return nil, true
 	}
 
+	svcType, ok := getServiceImportType(svc)
+
+	if !ok {
+		klog.Errorf("Service type %q not supported", svc.Spec.Type)
```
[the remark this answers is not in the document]
vthapar: Yeah, and also set serviceexport status to `Initialized,False`

pkg/agent/controller/agent.go
```diff
@@ -288,10 +375,67 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
...
+		if mcsIp == "" {
+			mcsIp = service.Spec.ClusterIP
+		}
+
+		return []string{mcsIp}
+	}
+
+	endpoint, err := a.kubeClientSet.CoreV1().Endpoints(service.Namespace).Get(service.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Endpoints for svc  (%s/%s) not found: %v", service.Namespace, service.Name, err)
```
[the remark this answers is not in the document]
vthapar: > So if it's not found, I assume we rely on the `endpointToRemoteServiceImport` to handle it. In this case we shouldn't log the error. Otherwise, I think we should return the error or nil or something to indicate to the caller to re-queue/retry.

Will change to debug, this was for my troubleshooting.

pkg/serviceimport/map.go
```diff
@@ -66,6 +59,37 @@ func (m *Map) SelectIP(namespace, name string, checkCluster func(string) bool) (
...
+		}
+
+		return serviceIps, found
+	}
+
+	for cluster, ips := range svcInfo.clusterIPs {
+		if checkCluster(cluster) && len(ips) > 0 {
+			serviceIps = append(serviceIps, ips...)
+		}
+	}
```
tpantelis, line 88: The prior code strived to release the lock ASAP and capture all fields to access in the critical section, as `queue` et al are recreated on change - we should do he same here:

```
clusterIPs, queue, counter, isHeadless := func() (map[string][]string, []clusterInfo, *uint64, bool) {
    m.RLock()
    defer m.RUnlock()

    si, ok := m.svcMap[keyFunc(namespace, name)]
    if !ok {
        return nil, nil, nil, false
    }

    return si.clusterIPs , si.clustersQueue, &si.rrCount, si.isHeadless
}()

if clusterIPs == nil {
    return nil, false
}

if !isHeadless {
    ip := m.selectIP(queue, counter, checkCluster)
    ...
}

for cluster, ips := range clusterIPs {
    ...
}
```

pkg/serviceimport/map.go
```diff
@@ -33,18 +36,8 @@ type Map struct {
...
-		defer m.RUnlock()
-
-		si, ok := m.svcMap[keyFunc(namespace, name)]
-		if !ok {
-			return nil, nil
-		}
-
-		return si.clustersQueue, &si.rrCount
-	}()
+func (m *Map) selectIP(si *serviceInfo, checkCluster func(string) bool) (string, bool) {
```
tpantelis, line 39:
```suggestion
func (m *Map) selectIP(queue []clusterInfo, counter *uint64, checkCluster func(string) bool) (string, bool) {
```

Also we don't need to return a bool anymore nor check for nil queue.

pkg/serviceimport/map.go
```diff
@@ -81,19 +105,23 @@ func (m *Map) Put(serviceImport *lighthousev2a1.ServiceImport) {
...
 			}
 		}
 
 		for _, info := range serviceImport.Status.Clusters {
-			remoteService.clusterIPs[info.Cluster] = info.IPs[0]
+			remoteService.clusterIPs[info.Cluster] = info.IPs
 		}
 
-		remoteService.buildClusterInfoQueue()
+		if serviceImport.Spec.Type == lighthousev2a1.SuperclusterIP {
```
tpantelis, line 122:
```suggestion
		if !remoteService.isHeadless {
```

pkg/agent/controller/agent.go
```diff
@@ -288,10 +402,59 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
 				originName:      svcExport.Name,
 				originNamespace: svcExport.Namespace,
 			},
+			Labels: map[string]string{
+				labelSourceName:      svcExport.Name,
+				labelSourceNamespace: svcExport.Namespace,
+				labelSourceCluster:   a.clusterID,
+			},
```
tpantelis, line 409: This is no longer needed.

pkg/agent/controller/agent.go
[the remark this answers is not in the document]
tpantelis: bump

pkg/agent/controller/agent.go
```diff
@@ -26,8 +28,13 @@ import (
 const (
 	submarinerIpamGlobalIp = "submariner.io/globalIp"
 	serviceUnavailable     = "ServiceUnavailable"
+	invalidServiceType     = "UnsupportedServiceType"
+	clusterIP              = "cluster-ip"
 	originName             = "origin-name"
 	originNamespace        = "origin-namespace"
+	labelSourceName        = "lighthouse.submariner.io/sourceName"
+	labelSourceNamespace   = "lighthouse.submariner.io/sourceNamespace"
+	labelSourceCluster     = "lighthouse.submariner.io/sourceCluster"
```
tpantelis, line 37: These are no longer needed.

pkg/agent/controller/agent.go
```diff
@@ -219,6 +287,54 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+		// Requeue
+		return nil, true
+	}
+
+	if serviceImport.Spec.Type == lighthousev2a1.Headless && a.globalnetEnabled {
+		klog.Errorf("Headless Services not supported with globalnet yet")
+		return nil, false
+	}
+
+	oldStatus := serviceImport.Status.DeepCopy()
```
tpantelis, line 314: Move this after L320.

pkg/agent/controller/agent.go
```diff
@@ -219,6 +287,54 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+	ipList := getIpsFromEndpoint(ep)
+
+	if serviceImport.Spec.Type == lighthousev2a1.SuperclusterIP && len(ipList) > 0 {
+		// Once we cache service and related objecs, this won't be needed
+		ipList = []string{serviceImport.Annotations[clusterIP]}
+	}
```
tpantelis, line 320: I'm not clear on this. Shouldn't this code only run for Headless since for SuperclusterIP we've already set the ServiceImport IP(s) to ether the service's ClusterIP (or GN IP) in `getIpsForService`? I'd think we'd do:

```
if serviceImport.Spec.Type == lighthousev2a1.SuperclusterIP {
    return nil, false
}
```

after L307.

pkg/agent/controller/agent.go
```diff
@@ -167,11 +216,18 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 	}
 
 	serviceImport.Spec = lighthousev2a1.ServiceImportSpec{
-		Type: lighthousev2a1.SuperclusterIP,
+		Type: svcType,
 	}
+	if svcType == lighthousev2a1.SuperclusterIP {
+		serviceImport.Annotations[clusterIP] = svc.Spec.ClusterIP
```
tpantelis, line 222: What is this annotation for?

pkg/agent/controller/agent.go
```diff
@@ -288,10 +402,59 @@ func (a *Controller) newServiceImport(svcExport *lighthousev2a1.ServiceExport) *
...
+				labelSourceName:      svcExport.Name,
+				labelSourceNamespace: svcExport.Namespace,
+				labelSourceCluster:   a.clusterID,
+			},
 		},
 	}
 }
 
+func (a *Controller) getServiceImport(svcExport *lighthousev2a1.ServiceExport) (*lighthousev2a1.ServiceImport, error) {
+	siName := svcExport.Name + "-" + svcExport.Namespace + "-" + a.clusterID
```
tpantelis, line 415: This is duplicated with L400 - please put in a `serviceImportNameFrom` function.

pkg/serviceimport/map.go
```diff
@@ -81,19 +105,23 @@ func (m *Map) Put(serviceImport *lighthousev2a1.ServiceImport) {
...
-				clusterIPs: make(map[string]string),
+				clusterIPs: make(map[string][]string),
 				rrCount:    0,
+				isHeadless: serviceImport.Spec.Type == lighthousev2a1.Headless,
 			}
 		}
 
 		for _, info := range serviceImport.Status.Clusters {
-			remoteService.clusterIPs[info.Cluster] = info.IPs[0]
+			remoteService.clusterIPs[info.Cluster] = info.IPs
```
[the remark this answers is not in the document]
vthapar: I've tested the use case, it works fine if `IPs` is empty list. I had to change some code to make list empty instead of inserting `""` for this reason.

pkg/agent/controller/agent.go
```diff
@@ -156,6 +186,21 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
...
+	if !ok {
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+			corev1.ConditionFalse, invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
+		klog.Errorf("Service type %q not supported", svc.Spec.Type)
+
+		return nil, false
+	}
+
+	if a.globalnetEnabled && svcType == lighthousev2a1.Headless {
+		klog.Errorf("Headless Services not supported with globalnet yet")
```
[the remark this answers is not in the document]
vthapar: Good catch, will do. We need some error status for this.

pkg/agent/controller/agent.go
```diff
@@ -156,6 +186,21 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		return nil, true
 	}
 
+	svcType, ok := getServiceImportType(svc)
+
+	if !ok {
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
```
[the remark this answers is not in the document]
vthapar: Agree. Will take a look at newer types. But it will need changes to spec, will likely bring those changes in subsequent PR

pkg/gateway/controller.go
```diff
@@ -207,12 +212,34 @@ func getGatewayStatus(obj *unstructured.Unstructured) (haStatus string, connecti
...
+	return clusterId, true
+}
+
 func (c *Controller) getClusterStatusMap() map[string]bool {
 	return c.clusterStatusMap.Load().(map[string]bool)
 }
 
 func (c *Controller) IsConnected(clusterId string) bool {
-	return !c.gatewayAvailable || c.getClusterStatusMap()[clusterId]
+	return !c.gatewayAvailable || c.localCluster == clusterId || c.getClusterStatusMap()[clusterId]
```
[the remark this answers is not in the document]
vthapar: This is an optimization to avoid reading localEndpoint everytime, which will never change but can become nil if there are no gateways. This also helps with scenario where we can become the only cluster coz all others are gone. In such scenarios local services that have been exported should still be available. Ideally I'd like to remove reading `localEndpoint` altogether. `clusterid` is deployment config that is available multiple places, so should get at init somehow.

pkg/serviceimport/map.go
vthapar: Will do.

pkg/agent/controller/agent.go
vthapar: Plan is to replace the annotation with labels. Since annotation is used in many other places, this is to introduce labels without breaking annotation dependent code. EndpointSlice and Import aggregation will require labels.

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
vthapar: This is to handle following use case:
1. There are no healthy pods, remove `ClusterIP` from `ServiceImport`
2. At least one pod is now healthy, add `ClusterIP` back to `ServiceImport`. I am using annotation to avoid doing a read on `Service` to get `ClusterIP`. We can remove this later once we add some caching for objects that we're interested in.

pkg/agent/controller/agent.go
vthapar: This is to handle following use case:

1. There are no healthy pods, remove `ClusterIP` from `ServiceImport`
2. At least one pod is now healthy, add `ClusterIP` back to `ServiceImport`.

Since we remove `ClusterIP` when l`en(ipList) == 0`, we have to add it back when it becomes non-zero.

pkg/agent/controller/agent.go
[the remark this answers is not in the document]
vthapar: Done

pkg/gateway/controller.go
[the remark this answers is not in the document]
tpantelis: You're obtaining the local clusterID from the `localEndpoint`  in the `Gateway`. But you're storing it as a field which can be accessed by multiple threads so technically needs to be synchronized. I'm suggesting to avoid this as outlined above.  Either way you're still obtaining the clusterID from the `localEndpoint`  in the `Gateway`.

pkg/agent/controller/agent.go
```diff
@@ -201,7 +201,10 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 	}
 
 	if a.globalnetEnabled && svcType == lighthousev2a1.Headless {
-		klog.Errorf("Headless Services not supported with globalnet yet")
+		klog.Infof("Headless Services not supported with globalnet yet")
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+			corev1.ConditionFalse, invalidServiceType, "Headless Services not supported with Global IP")
```
tpantelis, line 206:
```suggestion
			corev1.ConditionFalse, invalidServiceType, "Headless Services not supported with GlobalNet")
```

pkg/serviceimport/map.go
```diff
@@ -36,9 +36,7 @@ type Map struct {
 	sync.RWMutex
 }
 
-func (m *Map) selectIP(si *serviceInfo, checkCluster func(string) bool) (string, bool) {
-	queue, counter := si.clustersQueue, &si.rrCount
-
+func (m *Map) selectIP(queue []clusterInfo, counter *uint64, checkCluster func(string) bool) (string, bool) {
 	if queue == nil {
 		return "", false
 	}
```
tpantelis, line 42: We don't need this anymore.  Nor do we need to return the "found" boolean as the caller has already determined that.

pkg/agent/controller/agent.go
```diff
@@ -156,6 +190,24 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		return nil, true
 	}
 
+	svcType, ok := getServiceImportType(svc)
+
+	if !ok {
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+			corev1.ConditionFalse, invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
+		klog.Errorf("Service type %q not supported", svc.Spec.Type)
```
tpantelis, line 198:
```suggestion
		klog.Errorf("Service \"%s/%s\" type %q not supported", svcExport.Namespace, svc.Name, svc.Spec.Type)
```

pkg/agent/controller/agent.go
```diff
@@ -156,6 +190,24 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
...
+
+		return nil, false
+	}
+
+	if a.globalnetEnabled && svcType == lighthousev2a1.Headless {
+		klog.Infof("Headless Services not supported with globalnet yet")
```
tpantelis, line 204:
```suggestion
		klog.Infof("GlobalNet is enabled - ignoring headless Service \"%s/%s\", svcExport.Namespace, svc.Name)
```

pkg/agent/controller/agent.go
```diff
@@ -219,6 +291,55 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+
+	serviceImport, err := a.getServiceImport(svcExport)
+
+	if err != nil {
+		// Requeue
+		return nil, true
+	}
+
+	if serviceImport.Spec.Type == lighthousev2a1.Headless && a.globalnetEnabled {
+		klog.Errorf("Headless Services not supported with globalnet yet")
```
tpantelis, line 314: We already log this on L284.

pkg/agent/controller/agent.go
tpantelis: OK - I think I get it now. Suggested comment to clarify:  "_When there's no healthy pods, the IPs in the Endpoint will become empty and thus we also clear the ServiceImport IPs to avoid clients sending a request to a cluster which has no active pods to handle the request. Once at least one backing Endpoint pod becomes healthy, we need to re-establish the ServiceImport IPs from the previously cached clusterIP annotation._"

pkg/agent/controller/agent.go
```diff
@@ -219,6 +291,55 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, op syncer.
...
+		// some other error. Log and requeue
+		klog.Errorf("Unable to get ServiceExport for %#v: %v", ep, err)
+		return nil, true
+	}
+
+	serviceImport, err := a.getServiceImport(svcExport)
+
+	if err != nil {
+		// Requeue
+		return nil, true
```
tpantelis, line 310: Perhaps we should log the error if err is `IsNotFound`.

pkg/agent/controller/agent.go
```diff
@@ -263,35 +384,86 @@ func (a *Controller) updateExportedServiceStatus(name, namespace string, condTyp
...
+		if mcsIp == "" {
+			mcsIp = service.Spec.ClusterIP
+		}
+
+		return []string{mcsIp}
+	}
+
+	endpoint, err := a.kubeClientSet.CoreV1().Endpoints(service.Namespace).Get(service.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.V(log.DEBUG).Infof("Endpoints for svc  (%s/%s) not found: %v", service.Namespace, service.Name, err)
```
tpantelis, line 442: Most likely the error is `IsNotFound` but we shouldn't assume it.

```suggestion
                if !errors.IsNotFound(err) {
		    klog.Errorf("Error obtaining Endpoint for Service  (%s/%s): %v", service.Namespace, service.Name, err)
		}
```

pkg/agent/controller/agent.go
```diff
@@ -263,35 +384,86 @@ func (a *Controller) updateExportedServiceStatus(name, namespace string, condTyp
...
+		return make([]string, 0)
```
tpantelis, line 443: If the error is `IsNotFound`, I think we should return something (nil should be fine)  to indicate to the caller to re-queue to be complete. I don't think we can assume that `endpointToRemoteServiceImport` will subsequently run.

pkg/agent/controller/agent.go
```diff
@@ -167,20 +219,40 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
...
-			a.newClusterStatus(svc, a.clusterID),
+			{
+				Cluster: a.clusterID,
+				IPs:     a.getIpsForService(svc, svcType),
+			},
 		},
 	}
 
+	if svcType == lighthousev2a1.SuperclusterIP {
+		serviceImport.Annotations[clusterIP] = serviceImport.Status.Clusters[0].IPs[0]
```
tpantelis, line 235: Suggested comment: "_We also store the clusterIP in an annotation as an optimization to recover it in case the IPs are cleared out when here's no backing Endpoint pods_."

pkg/agent/controller/agent.go
vthapar: Not needed, it can be a race condition between ServiceImport creation and getting endpoint event. Error in `ServiceImport` creation is already logged.

pkg/agent/controller/agent.go
vthapar: It is possible to bring up a service with 0 replicaset and scale it up later. `endpointToRemoteServiceImport` should run once endpoints are created. If they're not, we'd want to keep ipList to empty. In what scenarios will we get `NotFound` here and yet `endpointToRemoteServiceImport` won't run?

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Actually we should re-queue if error is something other than `NotFound`

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
```diff
@@ -263,35 +384,86 @@ func (a *Controller) updateExportedServiceStatus(name, namespace string, condTyp
...
+	}
+
+	return getIpsFromEndpoint(endpoint)
+}
+
+func (a *Controller) getObjectNameWithClusterId(name, namespace string) string {
+	return name + "-" + namespace + "-" + a.clusterID
+}
+
+func getIpsFromEndpoint(endpoint *corev1.Endpoints) []string {
```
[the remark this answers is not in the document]
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Done

pkg/serviceimport/map.go
vthapar: done

pkg/gateway/controller.go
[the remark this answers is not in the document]
vthapar: Done

pkg/agent/controller/agent.go
vthapar: Oh, I think you meant error is other than `NotFound`. Done that.

tpantelis: > It is possible to bring up a service with 0 replicaset and scale it up later. `endpointToRemoteServiceImport` should run once endpoints are created. If they're not, we'd want to keep ipList to empty. In what scenarios will we get `NotFound` here and yet `endpointToRemoteServiceImport` won't run?

We shouldn't. I'm referring to an error other than `IsNotFound`, eg transient failure accessing the API server. The EP might already exist in which case `endpointToRemoteServiceImport` won't run and neither will this run again unless we re-queue (or unless some update happens to occur).  It's an edge case - we can handle it in a follow-up if you want so we can move this ahead.

pkg/agent/controller/agent.go
tpantelis: I meant log if err **isn't** `IsNotFound` 😄

PR #243, opened 2020-08-06
Update codegen for 1.17

skitt:
This pulls in the changes from Submariner, supporting codegen using
the Kubernetes 1.17-compatible version.


skitt:
(Yes, this should be in Shipyard, but it’s currently broken in Lighthouse so I want to fix it there first.)

PR #262, opened 2020-08-20
Add agent controller tests for failure scenarios

tpantelis:
Adds coverage for API server failures etc. I mocked errors for CRUD operations by adding a Reactor to the fake client sets. This code is generic and will be moved to admiral. The tests uncovered a minor bug in `endpointToRemoteServiceImport`.

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -187,7 +187,7 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		// some other error. Log and requeue
 		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
 			corev1.ConditionUnknown, "ServiceRetrievalFailed", fmt.Sprintf("Error retrieving the Service: %v", err))
-		klog.Errorf("Unable to get service for %#v: %v", svc, err)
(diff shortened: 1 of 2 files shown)


pkg/agent/controller/agent.go
```diff
@@ -187,7 +187,7 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 		// some other error. Log and requeue
 		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
 			corev1.ConditionUnknown, "ServiceRetrievalFailed", fmt.Sprintf("Error retrieving the Service: %v", err))
-		klog.Errorf("Unable to get service for %#v: %v", svc, err)
+		klog.Errorf("Error retrieving Service %s/%s: %v", svcExport.Namespace, svcExport.Name, err)
```
vthapar, line 190:
```suggestion
		klog.Errorf("Error retrieving Service (%s/%s): %v", svcExport.Namespace, svcExport.Name, err)
```

nit: We're using `(%s/%s)` in rest of code so would be good to follow that convention. Applies to similar changes in rest of code too.

vthapar (changes requested):
Looks good except for one cosmetic nit.

pkg/agent/controller/agent.go
tpantelis: Done

vthapar (approved):

Issue #215, opened 2020-07-22
Add support for StatefulSets


vthapar:
To support StatefulSets [https://kubernetes.io/docs/concepts/workloads/controllers/statefulset] we need to provide means to access individual pods in the set using hostnames, rather than just return list of PodIPs as we do in Headless Services. This is currently not possible with what we have in ServiceImports and will likely need EndpointSlice support.

sridhargaddam:
@vthapar If this issue targets only non-globalnet use-cases, can you please open a new issue for Globalnet use-case so that we don't miss it. Thanks.

manosnoam:
I'm closing this and open a new E2E test request

Issue #200, opened 2020-07-10
Access remote services in a specific cluster explicitly


davecaplinger:
I'm looking at options for deploying submariner and lighthouse in pre-existing clusters.  If two clusters `east` and `west` have the same (default) dns suffix `cluster.local` is there a way for pods in cluster `east` to access services in cluster `west` explicitly using some form of a forward plugin config for CoreDNS in `east` so that lookups for `*.svc.west.local` (or something else) are sent to lighthouse and will only match services exported by the `west` cluster?

Note that this is a little different from but not necessarily incompatible with the current (0.4.0) design of Lighthouse which supports multi-cluster services which can be exported from (and therefore serviced by) one or more clusters.  Such multi-cluster services are visible in all clusters, and discoverable via a new `supercluster.local` domain name, and it's assumed that namespaces that export such services will be globally unique across the set of clusters.  (I assume this looks like `_service_._namespace_.svc.supercluster.local` ?)

However, if the clusters are pre-existing, it's possible that they may have existing namespaces that are not globally unique, such as `kube-system` or `monitoring` (or in my case `kafka`).  So I'm trying to connect to a specific remote service in a specific remote cluster, not a "multi-cluster" service that may exist in one or more clusters.  This makes the use case more similar to the pre-0.4.0 design of Lighthouse, but since both clusters are pre-existing, they do not have unique DNS names and both consider themselves to be `.cluster.local`.

mangelajo:
may be something like _service_._namespace_._cluster_.svc.supercluster*s*.local ? 

This is something that we may need to standardize from U/S but I see the use cases, like connecting to remote k8s API, without traversing the public network, or without an internet-exposed public API endpoint, etc..

vthapar:
MCS API spec supports `cluster.service.namespace.svc.clusterset.local` for headless services. But with way we do discovery in lighthouse, we can easily support such use case for ClusterIP services too.

@davecaplinger Does that work for you? You will need to know `ClusterID` of the cluster, not domain.

davecaplinger:
@vthapar - I don't think I understand how `cluster.service.namespace.svc.clusterset.local` would apply.  Can you provide an example fqdn to access a specific service in a specific namespace in a specific (remote) cluster even if that namespace exists in the local cluster?

manosnoam:
@davecaplinger  that's the documentation:
https://submariner.io/quickstart/openshift/service_discovery/#verify

I'll add a new issue to test it in our E2E.

[...]

PR #269, opened 2020-08-26
DNS Parser changes to support host.cluster.*

vthapar:
Current parser is taken as-is from coredns/kubernetes plugin and doesn't
handle the `<host>.<cluster>.<svcname>.<ns>.clusterset.local` type queries
for headless multicluster service use case. Since we currently don't
support SRV records, replace port and protocol parsing with
hostname and cluster.

This is a pre-requisite to support StatefulSets use case #215


vthapar:
With this change we can easily support https://github.com/submariner-io/lighthouse/issues/200

aswinsuryan (approved):

PR #282, opened 2020-09-05
Access services in specific cluster

### LINKED ISSUE #200: Access remote services in a specific cluster explicitly

vthapar:
Fixes #200

### DIFF
--- a/pkg/serviceimport/map.go
+++ b/pkg/serviceimport/map.go
@@ -53,7 +53,7 @@ func (m *Map) selectIP(queue []clusterInfo, counter *uint64, checkCluster func(s
 	return ""
 }
 
-func (m *Map) GetIPs(namespace, name string, checkCluster func(string) bool) ([]string, bool) {
+func (m *Map) GetIPs(namespace, name, cluster string, checkCluster func(string) bool) ([]string, bool) {
 	clusterIPs, queue, counter, isHeadless := func() (map[string][]string, []clusterInfo, *uint64, bool) {
 		m.RLock()
 		defer m.RUnlock()
@@ -68,6 +68,11 @@ func (m *Map) GetIPs(namespace, name string, checkCluster func(string) bool) ([]
 
 	if clusterIPs == nil {
 		return nil, false
+	}
+
+	if cluster != "" {
+		ips, found := clusterIPs[cluster]
+		return ips, found
 	}
 
 	if !isHeadless {
--- a/pkg/serviceimport/map_test.go
+++ b/pkg/serviceimport/map_test.go
@@ -37,7 +37,7 @@ var _ = Describe("ServiceImport Map", func() {
 	}
 
 	getIPs := func(ns, name string) []string {
-		ips, found := serviceImportMap.GetIPs(ns, name, checkCluster)
+		ips, found := serviceImportMap.GetIPs(ns, name, "", checkCluster)
 		Expect(found).To(BeTrue())
 		return ips
 	}
@@ -163,7 +163,7 @@ var _ = Describe("ServiceImport Map", func() {
 
 	When("a service does not exist", func() {
 		It("should return not found", func() {
-			_, found := serviceImportMap.GetIPs(namespace1, service1, checkCluster)
+			_, found := serviceImportMap.GetIPs(namespace1, service1, "", checkCluster)
 			Expect(found).To(BeFalse())
 		})
 	})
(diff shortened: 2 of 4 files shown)


pkg/serviceimport/map.go
```diff
@@ -70,6 +70,11 @@ func (m *Map) GetIPs(namespace, name string, checkCluster func(string) bool) ([]
 		return nil, false
 	}
 
+	if cluster != "" {
+		ips, found := clusterIPs[cluster]
+		return ips, found
```
tpantelis, line 75:
```suggestion
		return clusterIPs[cluster]
```

plugin/lighthouse/handler_test.go
```diff
@@ -407,15 +420,29 @@ func testHeadlessService() {
...
-		It("should succeed and write all IPs as A records in response", func() {
-			executeTestCase(lh, rec, test.Case{
-				Qname: service1 + "." + namespace1 + ".svc.cluster.local.",
-				Qtype: dns.TypeA,
-				Rcode: dns.RcodeSuccess,
-				Answer: []dns.RR{
-					test.A(service1 + "." + namespace1 + ".svc.cluster.local.    0    IN    A    " + serviceIP),
-					test.A(service1 + "." + namespace1 + ".svc.cluster.local.    0    IN    A    " + serviceIP2),
-				},
+		When("query for headless service", func() {
```
tpantelis, line 423:
```suggestion
		When("no cluster is requested", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -407,15 +420,29 @@ func testHeadlessService() {
...
+					Qtype: dns.TypeA,
+					Rcode: dns.RcodeSuccess,
+					Answer: []dns.RR{
+						test.A(service1 + "." + namespace1 + ".svc.clusterset.local.    0    IN    A    " + serviceIP),
+						test.A(service1 + "." + namespace1 + ".svc.clusterset.local.    0    IN    A    " + serviceIP2),
+					},
+				})
+			})
+		})
+		When("query for headless service in specific cluster", func() {
```
tpantelis, line 436:
```suggestion
		When("requested for a specific cluster", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -407,15 +420,29 @@ func testHeadlessService() {
...
+			It("should succeed and write one IP as A record in response", func() {
```
tpantelis, line 437:
```suggestion
			It("should succeed and write the cluster's IP as A record in response", func() {
```

pkg/serviceimport/map_test.go
```diff
@@ -163,7 +163,7 @@ var _ = Describe("ServiceImport Map", func() {
```
tpantelis, line 163: I think we should add a test case for non-empty cluster to be complete.

pkg/serviceimport/map.go
vthapar: This gives `Not enough arguments to return`

tpantelis (approved):

pkg/serviceimport/map_test.go
vthapar: Done

tpantelis (dismissed):

aswinsuryan (approved):

tpantelis (approved):

PR #283, opened 2020-09-05
Initial changes for StatefulSet

### LINKED ISSUE #215: Add support for StatefulSets

vthapar:
This adds the following to the plugin code:
1. Endpointslice controller to get CRUD events for Endpointslices. We only watch for endpointslices managed by lighthouse.
2. An endpoints map to store hostname->IP information from endpoints present in EndpointSlices
3. `handler.go` changes to use endpoints map for queries of type `hostname.clusterX.svcname.ns.clusterset.local`

This also refactors some existing code.
1. Create lhconstants file to store all lighthouse specific strings used across controllers.
2. Add a localtransform to EndpointSlice syncer to filter out ones not managed by lighthouse.
3. Disable local and remote equivalency checks for EndpointSlice syncer. This is coz we already have that check when creating endpoint slices and these were not adding any value, just unnecessary overhead.

Fixes #215 

### DIFF
--- a/pkg/agent/controller/endpoint.go
+++ b/pkg/agent/controller/endpoint.go
@@ -93,6 +93,8 @@ func (e *EndpointController) Start(stopCh <-chan struct{}, labelSelector fmt.Str
 		},
 	)
 
+	klog.V(log.DEBUG).Infof("Starting Endpoint watcher for %q started", e.serviceImportName)
+
 	go e.endpointInformer.Run(e.stopCh)
 	go e.runEndpointWorker(e.endpointInformer, e.endPointqueue)
 }
@@ -106,7 +108,7 @@ func (e *EndpointController) runEndpointWorker(informer cache.Controller, queue 
 	for {
 		keyObj, shutdown := queue.Get()
 		if shutdown {
-			klog.Infof("Lighthouse watcher for ServiceImports stopped")
+			klog.V(log.DEBUG).Infof("Endpoints watcher for %q stopped", e.serviceImportName)
 			return
 		}
 
@@ -131,10 +133,6 @@ func (e *EndpointController) runEndpointWorker(informer cache.Controller, queue 
 			}
 
 			if err != nil {
-				if !exists {
-					e.endpointDeletedMap.Store(key, obj)
-				}
-
 				queue.AddRateLimited(key)
 			} else {
 				queue.Forget(key)
@@ -181,8 +179,6 @@ func (e *EndpointController) endPointDeleted(key string) error {
 		return nil
 	}
 
-	e.endpointDeletedMap.Delete(key)
-
 	endPoints := obj.(corev1.Endpoints)
 	endpointSliceName := endPoints.Name + "-" + e.clusterID
 	err := e.kubeClientSet.DiscoveryV1beta1().EndpointSlices(endPoints.Namespace).
@@ -191,6 +187,8 @@ func (e *EndpointController) endPointDeleted(key string) error {
 		klog.Errorf("Deleting EndpointSlice for endPoints %v from NameSpace %s failed due to %v", endPoints, endPoints.Namespace, err)
 		return err
 	}
+
+	e.endpointDeletedMap.Delete(key)
 
 	return nil
 }
@@ -257,12 +255,17 @@ func endpointFromAddress(address corev1.EndpointAddress, ready bool) discovery.E
 		topology["kubernetes.io/hostname"] = *address.NodeName
 	}
 
-	return discovery.Endpoint{
+	endpoint := discovery.Endpoint{
 		Addresses:  []string{address.IP},
 		Conditions: discovery.EndpointConditions{Ready: &ready},
 		Topology:   topology,
-		Hostname:   &address.Hostname,
 	}
+
+	if address.Hostname != "" {
+		endpoint.Hostname = &address.Hostname
+	}
+
+	return endpoint
 }
 
 func allAddressesIPv6(addresses []corev1.EndpointAddress) bool {
--- a/pkg/agent/controller/endpoint.go
+++ b/pkg/agent/controller/endpoint.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"fmt"
 
+	"github.com/submariner-io/admiral/pkg/log"
+	lhconstants "github.com/submariner-io/lighthouse/pkg/constants"
 	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -52,21 +54,21 @@ func (e *EndpointController) Start(stopCh <-chan struct{}, labelSelector fmt.Str
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				key, err := cache.MetaNamespaceKeyFunc(obj)
-				klog.Infof("Endpoint %q added", key)
+				klog.V(log.DEBUG).Infof("Endpoint %q added", key)
 				if err == nil {
 					e.endPointqueue.Add(key)
 				}
 			},
 			UpdateFunc: func(obj interface{}, new interface{}) {
 				key, err := cache.MetaNamespaceKeyFunc(new)
-				klog.Infof("Endpoint %q updated", key)
+				klog.V(log.DEBUG).Infof("Endpoint %q updated", key)
 				if err == nil {
 					e.endPointqueue.Add(key)
 				}
 			},
 			DeleteFunc: func(obj interface{}) {
 				key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-				klog.Infof("Endpoint %q deleted", key)
+				klog.V(log.DEBUG).Infof("Endpoint %q deleted", key)
 				if err == nil {
 					var endPoints *corev1.Endpoints
 					var ok bool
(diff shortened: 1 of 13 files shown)


vthapar:
This is complete in terms of implementing the StatefulSets use case, but doesn't do any cleanup and refactoring. As of this PR, there is bit of duplication of information in EndpointSlices and ServiceImports, which will be cleaned up in subsequent patches. Aim is to enable the basic use case, add tests [E2E and UT] and then do refactoring.

aswinsuryan (approved):

vthapar:
Operator e2e needs https://github.com/submariner-io/shipyard/pull/302 to pass. globalnet passes coz we don't run headless services and stateful e2e for globalnet.

aswinsuryan (approved):

test/e2e/discovery/statefulsets.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+		})
+	})
+})
+
+func RunSSDiscoveryTest(f *lhframework.Framework) {
+	clusterAName := framework.TestContext.ClusterIDs[framework.ClusterA]
+	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
+
+	By(fmt.Sprintf("Creating an Nginx Stateful Set on on %q", clusterBName))
```
mangelajo, line 54: "on on" -> "on"

pkg/endpointslice/controller.go
```diff
@@ -0,0 +1,123 @@
...
+			},
+		},
+		&discovery.EndpointSlice{},
+		0,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				key, err := cache.MetaNamespaceKeyFunc(obj)
+				if err == nil {
+					c.endpointSliceCreatedOrUpdated(key)
+				}
```
tpantelis, line 68: `obj` is the `EndpointSlice`  so you don't need to obtain it from the informer store.

```suggestion
                                c.mapStore.Put( obj.(*discovery.EndpointSlice))
```

pkg/endpointslice/controller.go
```diff
@@ -0,0 +1,123 @@
...
+			},
+			UpdateFunc: func(obj interface{}, new interface{}) {
+				key, err := cache.MetaNamespaceKeyFunc(obj)
+				if err == nil {
+					c.endpointSliceCreatedOrUpdated(key)
+				}
```
tpantelis, line 74:
```suggestion
				c.mapStore.Put( new.(*discovery.EndpointSlice))
```

pkg/endpointslice/controller.go
```diff
@@ -0,0 +1,123 @@
...
+			},
+			DeleteFunc: func(obj interface{}) {
+				key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+				if err == nil {
```
tpantelis, line 78: You don't ned to extract the key here.

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,70 @@
...
+	m.Lock()
+	defer m.Unlock()
+
+	epInfo := &endpointInfo{
+		key:     key,
+		hostIPs: make(map[string][]string),
+	}
+
+	for _, endpoint := range es.Endpoints {
+		epInfo.hostIPs[*endpoint.Hostname] = endpoint.Addresses
```
tpantelis: We should check if `Hostname` is nil as it is optional. Also not sure if `Addresses` could be nil but wouldn't hurt to check as well.

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,70 @@
...
+	key := keyFunc(namespace, name, cluster)
+
+	result, ok := m.epMap[key]
+	if !ok {
+		return nil, false
+	}
+
+	ips, ok := result.hostIPs[hostname]
+
+	return ips, ok
```
tpantelis, line 30:
```suggestion
	return result.hostIPs[hostname]
```

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type Map struct {
+	epMap map[string]*endpointInfo
+	sync.RWMutex
+}
+
+func (m *Map) GetIPs(hostname, cluster, namespace, name string) ([]string, bool) {
+	key := keyFunc(namespace, name, cluster)
+
```
tpantelis, line 22: Need to lock the mutex here.

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (m *Map) Put(es *discovery.EndpointSlice) {
+	key, ok := es.Labels[constants.LabelServiceImportName]
+	if !ok {
+		return
+	}
+
+	m.Lock()
+	defer m.Unlock()
```
tpantelis, line 46: These cold be moved just before L57 to reduce the critical section area.

pkg/endpointslice/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+type Controller struct {
+	// Indirection hook for unit tests to supply fake client sets
+	NewClientset func(kubeConfig *rest.Config) (kubernetes.Interface, error)
+	epsInformer  cache.Controller
+	stopCh       chan struct{}
+	mapStore     Store
```
tpantelis, line 23:
```suggestion
	store     Store
```

test/e2e/discovery/statefulsets.go
```diff
@@ -0,0 +1,263 @@
...
+
+	for _, endpointSlice := range endpointSlices.Items {
+		sourceCluster := endpointSlice.Labels[lhconstants.LabelSourceCluster]
+
+		for _, endpoint := range endpointSlice.Endpoints {
+			verifyEndpointsWithDig(f.Framework, framework.ClusterA, netshootPodList, endpoint, sourceCluster,
+				nginxServiceClusterB.Name, checkedDomains, true)
+			verifyCount++
+		}
+	}
```
tpantelis, line 143: This looks to be duplicated below and could be put into a function.

tpantelis (changes requested):

pkg/agent/controller/endpoint.go
```diff
@@ -93,6 +93,8 @@ func (e *EndpointController) Start(stopCh <-chan struct{}, labelSelector fmt.Str
 		},
 	)
 
+	klog.V(log.DEBUG).Infof("Starting Endpoint watcher for %q started", e.serviceImportName)
```
tpantelis, line 96:
```suggestion
	klog.V(log.DEBUG).Infof("Endpoints watcher for %q started", e.serviceImportName)
```

test/e2e/discovery/statefulsets.go
```diff
@@ -0,0 +1,269 @@
...
+
+	Expect(verifyCount).To(Equal(1))
```
mangelajo, line 84: When this fails it won't be very descriptive, I suggest we add some meaningful explanation in the expect:
```
	Expect(verifyCount).To(Equal(1), "Explaining message...")
```

test/e2e/discovery/statefulsets.go
```diff
@@ -0,0 +1,269 @@
...
+		for _, endpoint := range endpointSlice.Endpoints {
+			verifyEndpointsWithDig(f.Framework, framework.ClusterA, netshootPodList, endpoint, sourceCluster,
+				nginxServiceClusterB.Name, checkedDomains, false)
+		}
+	}
+}
+
+func RunSSDiscoveryLocalTest(f *lhframework.Framework) {
+	clusterAName := framework.TestContext.ClusterIDs[framework.ClusterA]
+	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
```
mangelajo, line 100: Comment completely unrelated to your patch.

I'm thinking that it'd be cleaner to encapsulate the context within the framework "f" itself. So we could do

f.Context.ClusterIDs[....]

test/e2e/discovery/statefulsets.go
```diff
@@ -0,0 +1,269 @@
...
+		sourceCluster := endpointSlice.Labels[lhconstants.LabelSourceCluster]
+
+		for _, endpoint := range endpointSlice.Endpoints {
+			verifyEndpointsWithDig(f.Framework, framework.ClusterA, netshootPodList, endpoint, sourceCluster,
+				nginxServiceClusterB.Name, checkedDomains, true)
+			verifyCount++
+		}
+	}
+
+	Expect(verifyCount).To(Equal(2))
```
mangelajo, line 145:
```
	Expect(verifyCount).To(Equal(2), "Message for humans :-) ")
```

pkg/endpointslice/map.go
vthapar: Fixed in subsequent commit.

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,72 @@
...
+
+	epInfo := &endpointInfo{
+		key:     key,
+		hostIPs: make(map[string][]string),
+	}
+
+	for _, endpoint := range es.Endpoints {
+		if endpoint.Hostname != nil {
+			epInfo.hostIPs[*endpoint.Hostname] = endpoint.Addresses
+		}
```
tpantelis, line 56: Should we also check `len(endpoint.Addresses) > 0`?

pkg/endpointslice/map.go
vthapar: I had originally tried this way, it gives error `Not enough arguments to return`

tpantelis: hmm that's weird - the two should be equivalent. Not a big deal.

pkg/endpointslice/map.go
vthapar: No, empty list will generate an empty A record i.e. `NODATA` which is the correct response in this scenario. len check would wrongly return NXDOMAIN.

pkg/endpointslice/map.go
```diff
@@ -0,0 +1,75 @@
...
+		return
+	}
+
+	epInfo := &endpointInfo{
+		key:     key,
+		hostIPs: make(map[string][]string),
+	}
+
+	m.Lock()
+	defer m.Unlock()
```
tpantelis, line 54: This could be moved even further to right before L62. We don't need to hold the lock which building the `epInfo` struct.

pkg/endpointslice/controller.go
```diff
@@ -62,38 +61,29 @@ func (c *Controller) Start(kubeConfig *rest.Config) error {
...
-					c.endpointSliceCreatedOrUpdated(key)
-				}
+				c.store.Put(obj.(*discovery.EndpointSlice))
 			},
 			UpdateFunc: func(obj interface{}, new interface{}) {
-				key, err := cache.MetaNamespaceKeyFunc(obj)
-				if err == nil {
-					c.endpointSliceCreatedOrUpdated(key)
-				}
+				c.store.Put(obj.(*discovery.EndpointSlice))
```
tpantelis, line 67:
```suggestion
				c.store.Put(new.(*discovery.EndpointSlice))
```

tpantelis (approved):

PR #293, opened 2020-09-11
Filter ServiceImports from other clusters

aswinsuryan:
The serviceimport controller need to respond only for the
serviceimport created in the local cluster.

### DIFF
--- a/pkg/agent/controller/serviceimport.go
+++ b/pkg/agent/controller/serviceimport.go
@@ -173,12 +173,15 @@ func (c *ServiceImportController) serviceImportDeleted(key string) error {
 	obj, found := c.serviceImportDeletedMap.Load(key)
 	if !found {
 		klog.Warningf("No endpoint controller found  for %q", key)
+		c.serviceImportDeletedMap.Delete(key)
+
 		return nil
 	}
 
 	si := obj.(*lighthousev2a1.ServiceImport)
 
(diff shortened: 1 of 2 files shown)


pkg/agent/controller/serviceimport.go
```diff
@@ -177,6 +178,10 @@ func (c *ServiceImportController) serviceImportDeleted(key string) error {
 
 	si := obj.(*lighthousev2a1.ServiceImport)
 
+	if si.GetLabels()[lhconstants.LabelSourceCluster] != c.clusterID {
+		return nil
```
vthapar, line 182:
```suggestion
		c.serviceImportDeletedMap.Delete(key)
		return nil
```

vthapar (changes requested):
Also, modify the test case as discussed to confirm the fix works.

vthapar (approved):

pkg/agent/controller/serviceimport.go
```diff
@@ -172,11 +173,18 @@ func (c *ServiceImportController) serviceImportDeleted(key string) error {
 	obj, found := c.serviceImportDeletedMap.Load(key)
 	if !found {
 		klog.Warningf("No endpoint controller found  for %q", key)
+		c.serviceImportDeletedMap.Delete(key)
```
tpantelis, line 176: This isn't needed since it wasn't found.

PR #294, opened 2020-09-11
Use metadata.labels as selector for endpoints

vthapar:
When creating an endpoint watcher for a serviceimport, use
metadata.labels instead of selector.labels. K8s copies over a service's
metadata.labels to the endpoint labels, making for a more accurate
filter criteria than selector labels which can vary.

### DIFF
--- a/pkg/agent/controller/serviceimport.go
+++ b/pkg/agent/controller/serviceimport.go
@@ -159,7 +159,7 @@ func (c *ServiceImportController) serviceImportCreatedOrUpdated(obj interface{},
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


aswinsuryan (approved):

pkg/agent/controller/serviceimport.go
```diff
@@ -159,7 +159,7 @@ func (c *ServiceImportController) serviceImportCreatedOrUpdated(obj interface{},
 		return nil
 	}
```
tpantelis, line 160: Do we still need this check?

tpantelis (approved):

PR #295, opened 2020-09-11
Remove EndpointSlice DeleteCollection

tpantelis:
Since we set the `ServiceImport` as the owner of the `EndpointSlice`, we don't ned to explicitly delete the `EndpointSlice` on `ServiceImport` delete.

Added verification of `EndpointSlice` deleted in E2E.

### DIFF
--- a/pkg/agent/controller/serviceimport.go
+++ b/pkg/agent/controller/serviceimport.go
@@ -191,13 +191,6 @@ func (c *ServiceImportController) serviceImportDeleted(key string) error {
 		c.endpointControllers.Delete(key)
 	}
 
-	labelSelector := labels.Set(map[string]string{lhconstants.LabelSourceName: si.Name}).AsSelector()
-	err := c.kubeClientSet.DiscoveryV1beta1().EndpointSlices(si.Namespace).
-		DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: labelSelector.String()})
-	if err != nil && !errors.IsNotFound(err) {
-		return err
-	}
-
 	c.serviceImportDeletedMap.Delete(key)
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/agent/controller/endpoint.go
```diff
@@ -194,22 +194,24 @@ func (e *EndpointController) endPointDeleted(key string) error {
...
-	endpointSlice.OwnerReferences = []metav1.OwnerReference{{
-		APIVersion:         "lighthouse.submariner.io.v2alpha1",
-		Kind:               "ServiceImport",
-		Name:               e.serviceImportName,
-		UID:                e.serviceImportUID,
-		Controller:         &controllerFlag,
-		BlockOwnerDeletion: nil,
+	endpointSlice := &discovery.EndpointSlice{ObjectMeta: metav1.ObjectMeta{
+		Name:      endpoints.Name + "-" + e.clusterID,
+		Namespace: e.serviceImportSourceNameSpace,
```
tpantelis, line 200: Setting `Namespace` to that of the owner `ServiceImport` is the functional change here. The owned resource must be in the same namespace as the owner so this ensures that even tho it should be the same as the `Endpoints`.  

I inlined the rest of the code for improved readability.

pkg/agent/controller/serviceimport.go
```diff
@@ -191,13 +191,6 @@ func (c *ServiceImportController) serviceImportDeleted(key string) error {
 		c.endpointControllers.Delete(key)
 	}
 
-	labelSelector := labels.Set(map[string]string{lhconstants.LabelSourceName: si.Name}).AsSelector()
```
aswinsuryan, line 194: Is the endpoint getting deleted due to the delete in the EndpointController or it is the garbage collector?

aswinsuryan (approved):

pkg/agent/controller/serviceimport.go
tpantelis: I assume you mean the EndpointSlice. The EndpointController deletes it due to Endpoints deleted - which I assume would only happen if the service was deleted.  This isn't necessary b/c we set the ServiceImport as the owner which k8s will GC.

aswinsuryan: ok got it.

vthapar: > Is the endpoint getting deleted due to the delete in the EndpointController or it is the garbage collector?

It gets deleted due to the owner ref that sets it to serviceimport.

PR #296, opened 2020-09-11
Set non zero TTL in dns replies

vthapar:
1. Set default ttl of 5 in DNS response as 0 is not recommended.
2. Make TTL a configurable parameter with range of 0-3600.

Note that default of 5 and range both are same as used in kubernetes
plugin.

### DIFF
--- a/plugin/lighthouse/handler.go
+++ b/plugin/lighthouse/handler.go
@@ -73,7 +73,7 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 	records := make([]dns.RR, 0)
 
 	for _, ip := range ips {
-		record := &dns.A{Hdr: dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}, A: net.ParseIP(ip).To4()}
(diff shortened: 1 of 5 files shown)


plugin/lighthouse/handler.go
```diff
@@ -73,7 +73,7 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 	records := make([]dns.RR, 0)
 
 	for _, ip := range ips {
-		record := &dns.A{Hdr: dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass()}, A: net.ParseIP(ip).To4()}
+		record := &dns.A{Hdr: dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass(), Ttl: 5}, A: net.ParseIP(ip).To4()}
```
tpantelis, line 76:
```suggestion
		record := &dns.A{Hdr: dns.RR_Header{Name: state.QName(), Rrtype: dns.TypeA, Class: state.QClass(), Ttl: lh.ttl}, A: net.ParseIP(ip).To4()}
```

tpantelis (changes requested):

skitt (changes requested):
The change itself seems fine, but could you use your real email address in your sign-off?

vthapar:
> The change itself seems fine, but could you use your real email address in your sign-off?

I've always used this for security/privacy reasons on github. Does this mess up any scripts?

skitt (approved):

tpantelis (approved):

PR #316, opened 2020-10-01
Use endpointslices for headless services

vthapar:
Currently headless services use IPs stored in status field of
`ServiceImports` which is not spec compliant. Now that we have
`EndpointSlices`, use those to resolve IPs for headless services.

### DIFF
--- a/pkg/endpointslice/map.go
+++ b/pkg/endpointslice/map.go
@@ -3,13 +3,20 @@ package endpointslice
 import (
 	"sync"
 
+	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/lighthouse/pkg/constants"
 	discovery "k8s.io/api/discovery/v1beta1"
+	"k8s.io/klog"
 )
 
 type endpointInfo struct {
-	key     string
+	key         string
+	clusterInfo map[string]*clusterInfo
+}
+
+type clusterInfo struct {
 	hostIPs map[string][]string
+	ipList  []string
 }
 
 type Map struct {
@@ -17,20 +24,41 @@ type Map struct {
 	sync.RWMutex
 }
 
-func (m *Map) GetIPs(hostname, cluster, namespace, name string) ([]string, bool) {
-	key := keyFunc(namespace, name, cluster)
+func (m *Map) GetIPs(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]string, bool) {
+	key := keyFunc(name, namespace)
 
-	m.RLock()
-	defer m.RUnlock()
+	clusterInfos := func() map[string]*clusterInfo {
+		m.RLock()
+		defer m.RUnlock()
 
-	result, ok := m.epMap[key]
-	if !ok {
+		result, ok := m.epMap[key]
+		if !ok {
+			return nil
+		}
+
+		return result.clusterInfo
+	}()
+
+	if clusterInfos == nil {
 		return nil, false
 	}
 
-	ips, ok := result.hostIPs[hostname]
+	switch {
+	case cluster == "":
+		ips := make([]string, 0)
 
-	return ips, ok
+		for k, val := range clusterInfos {
+			if checkCluster != nil && checkCluster(k) {
+				ips = append(ips, val.ipList...)
+			}
+		}
+
+		return ips, true
+	case hostname == "":
+		return clusterInfos[cluster].ipList, true
+	default:
+		return clusterInfos[cluster].hostIPs[hostname], true
+	}
 }
 
 func NewMap() *Map {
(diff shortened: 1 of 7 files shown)


pkg/endpointslice/map.go
```diff
@@ -40,36 +68,85 @@ func NewMap() *Map {
...
 		m.Lock()
 		defer m.Unlock()
-		delete(m.epMap, key)
+
+		epInfo, ok := m.epMap[key]
+		if !ok {
+			return
+		}
+
+		klog.V(log.DEBUG).Infof("Adding endpointInfo %#v for %s in %s", epInfo.clusterInfo[cluster], es.Name, cluster)
```
tpantelis, line 129:
```suggestion
		klog.V(log.DEBUG).Infof("Removing clusterInfo %#v for EndpointSlice %q in %q", epInfo.clusterInfo[cluster], es.Name, cluster)
```

pkg/serviceimport/map_test.go
```diff
@@ -234,52 +222,4 @@ var _ = Describe("ServiceImport Map", func() {
...
-			serviceImportMap.Put(serviceImport1)
-			serviceImport2 := newHeadlessServiceImport(namespace1, service1, clusterID2, "10.253.2.1", "10.253.2.2", "10.253.2.3")
-			serviceImportMap.Put(serviceImport2)
-
-			expectIPs(namespace1, service1, append(serviceImport1.Status.Clusters[0].IPs, serviceImport2.Status.Clusters[0].IPs...))
-
-			serviceImportMap.Remove(serviceImport2)
-			expectIPs(namespace1, service1, serviceImport1.Status.Clusters[0].IPs)
-		})
-	})
```
tpantelis, line 276: I thin k it would be good to keep these tests - could you add a similar test file for the EndpointSlice Map?

tpantelis (changes requested):

pkg/endpointslice/map.go
```diff
@@ -3,34 +3,62 @@ package endpointslice
...
 		return nil, false
 	}
 
-	ips, ok := result.hostIPs[hostname]
+	switch {
+	case cluster == "":
+		ips := make([]string, 0)
+
+		for k, val := range clusterInfos {
+			if checkCluster != nil && checkCluster(k) {
```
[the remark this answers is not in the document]
vthapar: I was going for just nil check but this is better, gives an option to bypass clustercheck if needed.

pkg/endpointslice/map.go
```diff
@@ -40,36 +68,85 @@ func NewMap() *Map {
...
 		return
 	}
 
-	epInfo := &endpointInfo{
-		key:     key,
-		hostIPs: make(map[string][]string),
+	cluster, ok := es.Labels[constants.LabelSourceCluster]
+
+	if !ok {
+		klog.V(log.DEBUG).Infof("Cluster label missing on %s", es.Name)
```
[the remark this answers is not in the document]
vthapar: Changed to warning. Using `es.Labels` as `es` can get pretty huge and we don't want that.

pkg/serviceimport/map_test.go
vthapar: I was planning to add those later as aswin will be making even more changes to this code after mine.

pkg/endpointslice/map.go
```diff
@@ -40,36 +68,86 @@ func NewMap() *Map {
...
 		return
 	}
 
-	epInfo := &endpointInfo{
-		key:     key,
-		hostIPs: make(map[string][]string),
+	cluster, ok := es.Labels[constants.LabelSourceCluster]
+
+	if !ok {
+		klog.Warningf("Cluster label missing on %#v", es.Labels)
```
tpantelis, line 80: This is unexpected so I think it's fine to print the whole object.  Just printing the labels doesn't provide any identifying info. We need at least name and namespace as well. Same on L73.

vthapar: > This is unexpected so I think it's fine to print the whole object. Just printing the labels doesn't provide any identifying info. We need at least name and namespace as well. Same on L73.

Endpoints can go into thousands. How about printing ObjectMeta?

pkg/endpointslice/map.go
```diff
@@ -70,17 +70,20 @@ func NewMap() *Map {
 func (m *Map) Put(es *discovery.EndpointSlice) {
 	key, ok := getKey(es)
 	if !ok {
-		klog.V(log.DEBUG).Infof("Failed to get labels on %#v", es)
+		klog.V(log.DEBUG).Infof("Failed to get key labels from %#v", es.Labels)
```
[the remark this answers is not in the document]
vthapar: Yeah, was in two minds so wanted your opinion. Will go with warning.

tpantelis (approved):

aswinsuryan (approved):

PR #322, opened 2020-10-06
Use Spec.IP for ClusterIP instead IP from status field 

aswinsuryan:
To align with the upstream spec, the IP  from the status
field of ServiceImport is not populated. The clusterIP will be added
to Spec.IP in the ServiceImport.


vthapar:
Why not use Spec.IP field for it?

aswinsuryan:
> Why not use Spec.IP field for it?

+1 , that is a good thought I will update the PR.

aswinsuryan:
> > Why not use Spec.IP field for it?
> 
> +1 , that is a good thought I will update the PR.

Have updated that patch to use Spec.IP

pkg/serviceimport/map.go
```diff
@@ -108,9 +109,7 @@ func (m *Map) Put(serviceImport *lighthousev2a1.ServiceImport) {
 			}
 		}
 
-		for _, info := range serviceImport.Status.Clusters {
-			remoteService.clusterIPs[info.Cluster] = info.IPs
-		}
+		remoteService.clusterIPs[serviceImport.GetLabels()[lhconstants.LabelSourceCluster]] = []string{serviceImport.Spec.IP}
```
tpantelis, line 112: So it appears `clusterIPs` will only ever have 1 IP so do we need to store as an array? Also change `GetIPs` to return single IP?

pkg/agent/controller/agent.go
```diff
@@ -394,25 +395,19 @@ func (a *Controller) endpointToRemoteServiceImport(obj runtime.Object, op syncer
...
-		Clusters: []lighthousev2a1.ClusterStatus{
-			{
-				Cluster: a.clusterID,
-				IPs:     ipList,
-			},
-		},
-	}
-
-	if reflect.DeepEqual(oldStatus, &serviceImport.Status) {
+	oldClusterIp := serviceImport.Spec.IP
```
tpantelis, line 403:
```suggestion
	oldClusterIP := serviceImport.Spec.IP
```

pkg/serviceimport/map.go
vthapar: > So it appears `clusterIPs` will only ever have 1 IP so do we need to store as an array? Also change `GetIPs` to return single IP?

This was needed when we were using ServiceImport for headless. Now this can be changed to just a string.

pkg/agent/controller/agent.go
```diff
@@ -482,7 +477,7 @@ func (a *Controller) getServiceExport(name, namespace string) (*lighthousev2a1.S
 
 func (a *Controller) serviceImportEquivalent(obj1, obj2 *unstructured.Unstructured) bool {
 	return syncer.DefaultResourcesEquivalent(obj1, obj2) &&
-		equality.Semantic.DeepEqual(util.GetNestedField(obj1, "status"), util.GetNestedField(obj2, "status"))
+		equality.Semantic.DeepEqual(util.GetNestedField(obj1, "metadata"), util.GetNestedField(obj2, "metadata"))
```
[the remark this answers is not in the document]
vthapar: > Why was this changed to check equivalence of "metadata"?
Coz we're no longer storing any information in status, so not needed. I think even metadata isn't needed here. Default covers spec, labels and annotations... fields that we care about.

vthapar, line 480: Just use default

vthapar (changes requested):

pkg/agent/controller/agent.go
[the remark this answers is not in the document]
aswinsuryan: The metadata was added for the initial patch, it is not required now.

pkg/serviceimport/map.go
```diff
@@ -26,7 +26,7 @@ func (si *serviceInfo) buildClusterInfoQueue() {
 	si.clustersQueue = make([]clusterInfo, 0)
 	for k, v := range si.clusterIPs {
```
tpantelis, line 27:
```suggestion
	for cluster, ip := range si.clusterIPs {
```

pkg/serviceimport/map.go
```diff
@@ -26,7 +26,7 @@ func (si *serviceInfo) buildClusterInfoQueue() {
...
 		if len(v) > 0 {
```
tpantelis, line 28: This check is no longer needed.

vthapar (approved):

tpantelis (approved):

Issue #331, opened 2020-10-08
Lighthouse: prefer the local cluster first, before routing to other remote clusters


nyechiel:
Today, for the use case where a single ClusterIP service is deployed to multiple clusters for redundancy or scale, the Lighthouse plugin uses a round robin selection, so that traffic will be eventually served by all clusters in the ClusterSet.

Ideally, routing should be optimized for some sort of cost metric (e.g look at complete Service Topology information and prioritize traffic local to zone, region, etc), but until this concept matures in the upstream community I'd like to propose that we default to what we used to do with the original Lighthouse design and always prefer the local cluster first, before routing the traffic to other remote clusters (in a round-robin fashion).

I hope we can achieve this by enhancing the Lighthouse agent, and without modifying the MCS API.

aswinsuryan:
We have a weight which is assigned to zero and unused. The weight can be increased to a higher value for the local cluster and can use weighted RR to select the IP. The plugin currently does not have the awareness of the cluster which it is running. But I think we could get it from one of the Submariner CRD. This will be required to assign a higher weight.

danibachar:
Can I take this one?

nyechiel:
> Aren't DNS requests from within the cluster will not get to the Lighthouse plugin?
> In any way, can I be assigned to it?

Sure, Daniel. Please sync with @aswinsuryan and @vthapar if you have any questions, but a PR is welcome.

vthapar:
@danibachar We get local clusterid from gateway object here https://github.com/submariner-io/lighthouse/blob/master/pkg/gateway/controller.go#L175 and storing it in connection map. We can easily tweak the code here to handle it differently from remote endpoints.

nyechiel:
@danibachar we want to support this feature in our upcoming release (v0.8). Do you have cycles to work on this in the next few weeks?

danibachar:
@nyechiel Yep!
Plan to have a PR ready by the end of this weekend

nyechiel:
@manosnoam FYI

manosnoam:
Thanks @nyechiel 
Added E2E request: https://github.com/submariner-io/lighthouse/issues/365

PR #343, opened 2020-10-14
Prefer local cluster service over remote

danibachar:
LocalClusterId getter + prefer local cluster service + tests

Related issue: #331

Signed-off-by: Daniel Bachar <danibachar89@gmail.com>

### DIFF
--- a/pkg/gateway/controller.go
+++ b/pkg/gateway/controller.go
@@ -30,6 +30,7 @@ type Controller struct {
 	queue            workqueue.Interface
 	stopCh           chan struct{}
 	clusterStatusMap atomic.Value
+	localClusterID   atomic.Value
 	gatewayAvailable bool
 }
 
@@ -41,6 +42,7 @@ func NewController() *Controller {
 		gatewayAvailable: true,
 	}
 	controller.clusterStatusMap.Store(make(map[string]bool))
+	controller.localClusterID.Store("")
 
 	return controller
 }
@@ -117,10 +119,18 @@ func (c *Controller) processNextGateway(key, name, ns string) (bool, error) {
 }
 
 func (c *Controller) gatewayCreatedOrUpdated(obj *unstructured.Unstructured) {
-	connections, ok := getGatewayStatus(obj)
+	connections, localClusterID, ok := getGatewayStatus(obj)
 	if !ok {
 		return
 	}
+
+	// Updating
+	c.updateLocalClusterIDIfNeeded(localClusterID)
+
+	c.updateClusterStatusMap(connections)
+}
+
+func (c *Controller) updateClusterStatusMap(connections []interface{}) {
 	var newMap map[string]bool
 
 	currentMap := c.getClusterStatusMap()
@@ -133,28 +143,28 @@ func (c *Controller) gatewayCreatedOrUpdated(obj *unstructured.Unstructured) {
 			klog.Errorf("status field not found in %#v", connectionMap)
 		}
 
-		clusterId, found, err := unstructured.NestedString(connectionMap, "endpoint", "cluster_id")
+		clusterID, found, err := unstructured.NestedString(connectionMap, "endpoint", "cluster_id")
 		if !found || err != nil {
 			klog.Errorf("cluster_id field not found in %#v", connectionMap)
 			continue
 		}
 
 		if status == "connected" {
-			_, found := currentMap[clusterId]
+			_, found := currentMap[clusterID]
 			if !found {
 				if newMap == nil {
 					newMap = copyMap(currentMap)
 				}
 
-				newMap[clusterId] = true
+				newMap[clusterID] = true
 			}
 		} else {
-			_, found = currentMap[clusterId]
+			_, found = currentMap[clusterID]
 			if found {
 				if newMap == nil {
 					newMap = copyMap(currentMap)
 				}
-				delete(newMap, clusterId)
+				delete(newMap, clusterID)
 			}
 		}
 	}
@@ -165,22 +175,32 @@ func (c *Controller) gatewayCreatedOrUpdated(obj *unstructured.Unstructured) {
 	}
 }
 
-func getGatewayStatus(obj *unstructured.Unstructured) (connections []interface{}, gwStatus bool) {
+func (c *Controller) updateLocalClusterIDIfNeeded(clusterID string) {
+	updateNeeded := clusterID != "" && clusterID != c.LocalClusterID()
+	if updateNeeded {
+		klog.Infof("Updating the gateway localClusterID %q ", clusterID)
+		c.localClusterID.Store(clusterID)
+	}
+}
+
+func getGatewayStatus(obj *unstructured.Unstructured) (connections []interface{}, clusterID string, gwStatus bool) {
 	status, found, err := unstructured.NestedMap(obj.Object, "status")
 	if !found || err != nil {
 		klog.Errorf("status field not found in %#v, err was: %v", obj, err)
-		return nil, false
+		return nil, "", false
 	}
 
-	localClusterId, found, err := unstructured.NestedString(status, "localEndpoint", "cluster_id")
+	localClusterID, found, err := unstructured.NestedString(status, "localEndpoint", "cluster_id")
 
 	if !found || err != nil {
 		klog.Errorf("localEndpoint->cluster_id not found in %#v, err was: %v", status, err)
+
+		localClusterID = ""
 	} else {
 		connections = append(connections, map[string]interface{}{
 			"status": "connected",
 			"endpoint": map[string]interface{}{
-				"cluster_id": localClusterId,
+				"cluster_id": localClusterID,
 			},
 		})
 	}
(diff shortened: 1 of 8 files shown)


pkg/serviceimport/map_test.go
```diff
@@ -84,6 +147,50 @@ var _ = Describe("ServiceImport Map", func() {
 				Expect(getIP(namespace1, service1)).To(Equal(secondIP))
 			}
 		})
+		When("we check for local cluster", func() {
+			When("cluster override IS NOT specified", func() {
+				It("should consistently return the relevant service IP per local cluster", func() {
```
tpantelis, line 152:
```suggestion
		When("an existent local cluster is specified", func() {
		    It("should consistently return its IP", func() {
```

pkg/serviceimport/map_test.go
```diff
@@ -84,6 +147,50 @@ var _ = Describe("ServiceImport Map", func() {
 				Expect(getIP(namespace1, service1)).To(Equal(secondIP))
 			}
```
tpantelis, line 148: This can be replaced by `testRoundRobin(namespace1, service1, "", "", []string{serviceIP1, serviceIP2})`

pkg/serviceimport/map_test.go
```diff
@@ -84,6 +147,50 @@ var _ = Describe("ServiceImport Map", func() {
...
+				})
+			})
+			When("cluster override IS NOT specified", func() {
+				It("should consistently return the IPs round-robin only for local cluster without service", func() {
+					ips := []string{serviceIP1, serviceIP2}
+					testRoundRobin(namespace1, service1, "", clusterID3, ips)
+					testRoundRobin(namespace1, service1, "", "", ips)
+				})
+			})
+			When("cluster override IS specified", func() {
```
tpantelis, line 166: This think this is sufficiently covered above by the single cluster test cases above.

pkg/serviceimport/map_test.go
```diff
@@ -84,6 +147,50 @@ var _ = Describe("ServiceImport Map", func() {
...
+			When("cluster override IS NOT specified", func() {
+				It("should consistently return the relevant service IP per local cluster", func() {
+					for i := 0; i < 10; i++ {
+						Expect(getLocalClusterIP(namespace1, service1, "", clusterID1)).To(Equal(serviceIP1))
+						Expect(getLocalClusterIP(namespace1, service1, "", clusterID2)).To(Equal(serviceIP2))
+					}
+				})
+			})
+			When("cluster override IS NOT specified", func() {
+				It("should consistently return the IPs round-robin only for local cluster without service", func() {
```
tpantelis, line 160:
```suggestion
			When("a non-existent local cluster is specified", func() {
				It("should consistently return the IPs round-robin", func() {
```

pkg/serviceimport/map_test.go
```diff
@@ -129,6 +236,23 @@ var _ = Describe("ServiceImport Map", func() {
 
 			})
 		})
+		When("we check for local cluster", func() {
```
tpantelis, line 239: I think this think this is sufficiently covered above.

tpantelis (changes requested):
Not sure if someone has mentioned this before but, when submitting review changes, please create a new commit instead of amending as that makes it much easier to determine what was changed. Thanks!

vthapar:
@danibachar I'm working on another PR which has a bit of overlap with this one. It can make some of your patch redundant. But if your patch is ready first, let us get it in, I will rework mine.

vthapar (approved):

test/e2e/discovery/service_discovery.go
```diff
@@ -238,47 +251,79 @@ func RunServicesPodAvailabilityMutliClusterTest(f *lhframework.Framework) {
 		f.AwaitServiceImportIP(framework.ClusterA, nginxServiceClusterA)
 	}
 
-	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterA, netshootPodList, checkedDomains, true)
-	verifyServiceIpWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB, netshootPodList, checkedDomains, true)
+	clusterBDomain := getClusterDomain(f.Framework, framework.ClusterB, netshootPodListB)
```
tpantelis, line 377: What's the functional purpose of `clusterBDomain`? It's only used for output on L256.

test/e2e/discovery/service_discovery.go
```diff
@@ -238,47 +251,79 @@ func RunServicesPodAvailabilityMutliClusterTest(f *lhframework.Framework) {
...
+
+	By(fmt.Sprintf("Fetching nginxServiceClusterB from %q, by overriding with %q", clusterBName, clusterBDomain))
```
tpantelis, line 379: I'm not clear on this output - it seems it should be after L258. But do we really need this extra output in this one case?

test/e2e/discovery/service_discovery.go
```diff
@@ -34,6 +49,15 @@ var _ = Describe("[discovery] Test Service Discovery Across Clusters", func() {
 		})
 	})
 
+	When("a pod tries to resolve a service which is exported locally and in a remote cluster", func() {
+		It("should resolve the local service", func() {
+			RunServiceDiscoveryLocalMultiClusterTest(f)
+		})
+		It("forced clusterID should return the desired cluster", func() {
```
tpantelis, line 56:
```suggestion
		When("a specific cluster is specified", func() {
		    It("should resolve the desired cluster's service", func() {
		   
```

pkg/gateway/controller_test.go
```diff
@@ -27,6 +27,18 @@ var _ = Describe("Gateway controller", func() {
 	t := newTestDiver()
 
 	When("an active Gateway is created", func() {
+		When("LocalClusterID is called", func() {
+			It("should return the correct local cluster ID", func() {
+				t.localClusterIDValidationTest(localClusterID)
+			})
+		})
+
+		When("LocalClusterID is called after ID is updated", func() {
```
tpantelis, line 36:
```suggestion
		When("LocalClusterID is called after the ID is updated", func() {
```

pkg/gateway/controller_test.go
```diff
@@ -82,6 +94,18 @@ var _ = Describe("Gateway controller", func() {
 			Expect(unstructured.SetNestedField(t.gatewayObj.Object, "passive", "status", "haStatus")).To(Succeed())
 		})
 
+		When("LocalClusterID is called", func() {
+			It("should return the correct local cluster ID", func() {
+				t.localClusterIDValidationTest(localClusterID)
+			})
+		})
+
+		When("LocalClusterID is called after ID is updated", func() {
```
tpantelis, line 103:
```suggestion
		When("LocalClusterID is called after the ID is updated", func() {
```

tpantelis (commented):
Looks good - just some minor comments.

pkg/gateway/controller.go
```diff
@@ -175,31 +175,31 @@ func (c *Controller) updateClusterStatusMap(connections []interface{}) {
 	}
 }
 
-func (c *Controller) updateClusterLocalId(clusterId string) {
-	if clusterId != "" {
-		klog.Infof("Updating the gateway localClusterId %#v ", clusterId)
-		c.clusterLocalId.Store(clusterId)
+func (c *Controller) updateLocalClusterID(clusterID string) {
+	if clusterID != "" {
+		klog.Infof("Updating the gateway localClusterID %q ", clusterID)
```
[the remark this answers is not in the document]
vthapar: Or better, log it at all.

plugin/lighthouse/handler.go
```diff
@@ -46,7 +46,10 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 
 	var ips []string
 	var found bool
-	ip, found := lh.serviceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, lh.clusterStatus.IsConnected)
+
+	localClusterID := lh.clusterStatus.LocalClusterID()
+
+	ip, found := lh.serviceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, localClusterID, lh.clusterStatus.IsConnected)
```
vthapar, line 52: Won't it be simpler to just do something like this:

```
if pReq.cluster == "" {
    localClusterID := lh.clusterStatus.LocalClusterID()
}
ip, found := lh.serviceImports.GetIP(pReq.namespace, pReq.service, localClusterID,  lh.clusterStatus.IsConnected)
if !found {
    ip, found := lh.serviceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, lh.clusterStatus.IsConnected)
}```

With this we don't need to modify GetIP at all and reuse the existing cluster semantic.

vthapar:
@danibachar @tpantelis Take a look at https://github.com/submariner-io/lighthouse/pull/356 I copied the localClusterID code from this one. While testing I noticed that E2E needs to be improved, we're not testing stuff correctly.

I'd prefer this patch to go in without e2e changes, just enough to get it to work, and then bring in e2e changes in a separate patch. Alternately, we can get #356 in.

plugin/lighthouse/handler.go
danibachar: Hmmm I see what you mean, but first it will require 2 GetIP requests which require locks and are kind of expensive - won't you agree?
In addition, it will cause the logic IP selection to be kind of abstract away from where it is configured in the PR, and makes the selection kind of implicit and not explicit...
The logic that is within the map object explicitly says:
If a clusterID is provided in the DNS request, we would return the relevant IP for that cluster,  i.e if the service is not presented in that cluster we would like to return an empty IP.
While local cluster preference means - try to get the service IP from the local cluster, but if no presented fall back to remote once...

vthapar: Yeah, good to avoid multiple locks.

vthapar (approved):
Let us get this in for now and add/modify E2E in subsequent PRs

test/e2e/discovery/service_discovery.go
danibachar: Removed

danibachar:
I removed the E2E tets. Besides the fix for a current test. Will add more E2E in separate PRs like suggested (:

pkg/gateway/controller_test.go
danibachar: Fixed

pkg/gateway/controller_test.go
danibachar: Fixed

test/e2e/discovery/service_discovery.go
danibachar: Removed

test/e2e/discovery/service_discovery.go
danibachar: Removed

aswinsuryan (approved):

tpantelis (approved):

PR #356, opened 2020-10-28
Use clusterIP for local services

### LINKED ISSUE #355: Use ClusterIP for local services
When exported service is present in local cluster and globalnet is enabled, return the clusterIP of service instead of GlobalIP.

### LINKED ISSUE #365: [E2E] Test service-discovery prefers local cluster before remote cluster
Please add E2E test for feature https://github.com/submariner-io/lighthouse/issues/331

vthapar:
Fixes #355
Fixes #365 

### DIFF
--- a/pkg/service/controller.go
+++ b/pkg/service/controller.go
@@ -1,0 +1,89 @@
+package service
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+type Controller struct {
+	// Indirection hook for unit tests to supply fake client sets
+	NewClientset func(kubeConfig *rest.Config) (kubernetes.Interface, error)
+	svcInformer  cache.Controller
+	svcStore     cache.Store
+	stopCh       chan struct{}
+}
+
+func NewController() *Controller {
+	return &Controller{
+		NewClientset: func(c *rest.Config) (kubernetes.Interface, error) {
+			return kubernetes.NewForConfig(c)
+		},
+		stopCh: make(chan struct{}),
+	}
+}
+
+func (c *Controller) Start(kubeConfig *rest.Config) error {
+	klog.Infof("Starting Services Controller")
+
+	clientSet, err := c.NewClientset(kubeConfig)
+	if err != nil {
+		return fmt.Errorf("error creating client set: %v", err)
+	}
+
+	c.svcStore, c.svcInformer = cache.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				return clientSet.CoreV1().Services(metav1.NamespaceAll).List(options)
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return clientSet.CoreV1().Services(metav1.NamespaceAll).Watch(options)
+			},
+		},
+		&v1.Service{},
+		0,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    func(obj interface{}) {},
+			UpdateFunc: func(old interface{}, new interface{}) {},
+			DeleteFunc: func(obj interface{}) {},
+		},
+	)
+
+	go c.svcInformer.Run(c.stopCh)
+
+	return nil
+}
+
+func (c *Controller) Stop() {
+	close(c.stopCh)
+
+	klog.Infof("Services Controller stopped")
+}
+
+func (c *Controller) GetIp(name, namespace string) (string, bool) {
+	key := namespace + "/" + name
+	obj, exists, err := c.svcStore.GetByKey(key)
+
+	if err != nil {
+		return "", false
+	}
+
+	if !exists {
+		return "", false
+	}
+
+	svc := obj.(*v1.Service)
+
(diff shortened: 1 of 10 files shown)


vthapar:
WiP

pkg/service/controller.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+func (c *Controller) Stop() {
+	close(c.stopCh)
+
+	klog.Infof("Services Controller stopped")
+}
+
+func (c *Controller) GetIp(name, namespace string) (string, bool) {
```
tpantelis, line 70:
```suggestion
func (c *Controller) GetIP(name, namespace string) (string, bool) {
```

pkg/service/controller.go
```diff
@@ -0,0 +1,89 @@
...
+	key := namespace + "/" + name
+	obj, exists, err := c.svcStore.GetByKey(key)
+
+	if err != nil {
+		return "", false
```
tpantelis, line 75: We should log the error.

pkg/serviceimport/map.go
```diff
@@ -54,7 +54,7 @@ func (m *Map) selectIP(queue []clusterInfo, counter *uint64, name, namespace str
 }
 
 func (m *Map) GetIP(namespace, name, cluster, localCluster string, checkCluster func(string) bool,
-	checkEndpoint func(string, string, string) bool) (string, bool) {
+	checkEndpoint func(string, string, string) bool) (ip string, found, isLocal bool) {
```
tpantelis, line 57: The named return vars are not used. Are you naming them just for documentation?

pkg/serviceimport/map_test.go
```diff
@@ -40,12 +40,12 @@ var _ = Describe("ServiceImport Map", func() {
 	}
 
 	expectIPsNotFound := func(ns, service, cluster, localCluster string) {
-		_, found := serviceImportMap.GetIP(ns, service, cluster, localCluster, checkCluster, checkEndpoint)
+		_, found, _ := serviceImportMap.GetIP(ns, service, cluster, localCluster, checkCluster, checkEndpoint)
```
tpantelis, line 43: I think we should verify `isLocal` and add a test for it.

pkg/serviceimport/map.go
```diff
@@ -68,13 +68,13 @@ func (m *Map) GetIP(namespace, name, cluster, localCluster string, checkCluster
...
-		return "", false
+		return "", false, false
 	}
 
 	// If a clusterId is specified, we supply it even if the service is not there
 	if cluster != "" {
-		ip, found := clusterIPs[cluster]
-		return ip, found
+		ip, found = clusterIPs[cluster]
+		return ip, found, false
```
tpantelis, line 77:
```suggestion
		return ip, found, cluster == localCluster
```

plugin/lighthouse/handler.go
```diff
@@ -111,6 +111,20 @@ func (lh *Lighthouse) emptyResponse(state request.Request) (int, error) {
 	return dns.RcodeSuccess, nil
 }
 
+func (lh *Lighthouse) getClusterIpForSvc(pReq recordRequest) (ip string, found bool) {
+	localClusterID := lh.clusterStatus.LocalClusterID()
+
+	ip, found, isLocal := lh.serviceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, localClusterID, lh.clusterStatus.IsConnected,
+		lh.endpointsStatus.IsHealthy)
+
+	getLocal := isLocal || (pReq.cluster != "" && pReq.cluster == localClusterID)
```
tpantelis, line 120: The OR clause isn't needed here with the suggested change  `serviceImports.GetIP`.

plugin/lighthouse/handler.go
```diff
@@ -111,6 +111,20 @@ func (lh *Lighthouse) emptyResponse(state request.Request) (int, error) {
...
+	if found && getLocal {
+		ip, found = lh.localServices.GetIp(pReq.service, pReq.namespace)
+	}
```
tpantelis, line 123: Perhaps it would cleaner to incorporate `localServices` into `serviceImports.GetIP`. But we can consider that for a separate PR.

tpantelis (changes requested):

pkg/serviceimport/map.go
vthapar: Yes, for documentation.

plugin/lighthouse/handler.go
vthapar: ServiceImport and Service are using two different informers. I have an idea on how to make it work, but will revisit it later after 0.8.0 With the plan to use EndpointSlice instead of ServiceImport for clusterIP, it will become easier.

tpantelis: `LocalServices` is an interface so we could pass it into `serviceimport.Map`.

vthapar: Aha. Can do it in a separate PR later.

pkg/serviceimport/map_test.go
vthapar: Will have to add a new test for it. Just adding an Expect for it here breaks the tests. Can add it later along with GetIP rework.

Dockerfile.dapper718299099
```diff
@@ -0,0 +1,10 @@
+FROM quay.io/submariner/shipyard-dapper-base:devel
```
aswinsuryan, line 1: This file needs to be removed.

aswinsuryan (commented):
LGTM

Dockerfile.dapper718299099
vthapar: Argh, forgot to cleanup before updating. Will do so when I rebase.

aswinsuryan (approved):

nyechiel:
@vthapar are you still working on this for 0.8?

vthapar:
> @vthapar are you still working on this for 0.8?

Review comments are already addressed, waiting for next batch of review/approvals.

tpantelis (approved):

PR #334, opened 2020-10-09
Use the MCS APIs instead of our own

skitt:
This depends on #322.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

PR #357, opened 2020-10-29
Use the MCS APIs instead of our own

aswinsuryan:
Fixes: /submariner-io/lighthouse#336

Co-authored-by: Stephen Kitt <skitt@redhat.com>


nyechiel:
@aswinsuryan @skitt can we close https://github.com/submariner-io/lighthouse/pull/334 at this point?

skitt:
> @aswinsuryan @skitt can we close #334 at this point?

Done!

pkg/agent/controller/agent.go
```diff
@@ -194,7 +194,7 @@ func (a *Controller) serviceExportToRemoteServiceImport(obj runtime.Object, op s
 	obj, found, err := a.serviceSyncer.GetResource(svcExport.Name, svcExport.Namespace)
 	if err != nil {
 		// some other error. Log and requeue
-		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, lighthousev2a1.ServiceExportInitialized,
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
```
tpantelis, line 197: It seems odd to set it as `ServiceExportValid`. But the only other value is `ServiceExportConflict` which, the way they've defined it, is less appropriate. It seems the API  is a  bit lacking but it is what is.

tpantelis (approved):

pkg/agent/controller/agent.go
vthapar: I think it is combined with Condition to mean something like "ServiceExportValid: Unknown" and finally set it to ready when it is truly valid.

vthapar:
As discussed in call, we need a controller to convert old ServiceExport to new ones till we have changes in operator and subctl available to avoid chicken-n-egg scenario

pkg/agent/controller/agent.go
tpantelis: ok - still seems a bit odd tho, especially if one just looks at `ServiceExportValid`.  The code docs aren't clear on the usage (`ConditionStatus` has no docs) and, in fact, looks incorrect in some places. Anyway that's an upstream issue - maybe I'lll raise an  issue there.

skitt: It’s only a prima facie state: it indicates that the export is syntactically valid, and will therefore be processed. Anything that happens after that isn’t reflected in the status, but in the message. It does seem rather odd, but is apparently in keeping with K8s practice elsewhere. It would still be worth bringing it up upstream...

tpantelis: yeah - all the fields together make up the full status. It's OK just a bit different than what I'm used to.  I mean if I saw `ServiceExportValid` in output I'd think all is good but it may  not be - you also have to look at `ConditionStatus`. I'd expect a `ServiceExportFailed` state.  But it seems their design reasoning is that the field represents a category or type of condition to assert and the assertion of that condition is indicated by `ConditionStatus`, eg

- if we fail to lookup the service then we assert **ServiceExportValid: Unknown** meaning we can't determine if it's valid or not
- if the service is not exportable  or we fail to actually export it the we assert **ServiceExportValid: False** with the `Reason` providing more detail.

aswinsuryan:
> As discussed in call, we need a controller to convert old ServiceExport to new ones till we have changes in operator and subctl available to avoid chicken-n-egg scenario

I have added the controller to convert Lighthouse ServiceImport to mcs ServiceImport

pkg/agent/controller/agent.go
```diff
@@ -146,6 +158,18 @@ func NewWithDetail(spec *AgentSpecification, syncerConf *broker.SyncerConfig, re
...
 	}
 
+	agentController.serviceLHExportController, err = newLHServiceExportController(lighthouseClient, mcsClient)
+
+	if err != nil {
+		return nil, err
+	}
+
+	agentController.serviceMCSExportController, err = newMCSServiceExportController(lighthouseClient, mcsClient)
+
```
tpantelis, line 168: We should use ResourceSyncers - saves a ton of boilerplate.

aswinsuryan: Replaced with resourceSyncer, there is a UT failure "no kind is registered for the type v2alpha1.ServiceExport", I think I need to add it to scheme somewhere in the UT.

tpantelis: yes - L354 in _agent_test.go_.

aswinsuryan: Thanks I have fixed it .

pkg/agent/controller/lhserviceexport.go
```diff
@@ -56,3 +61,15 @@ func (c *LHServiceExportController) LHtoMCSServiceExport(obj runtime.Object, op
 
 	return mcsServiceExport, false
 }
+
+func (c *LHServiceExportController) deleteLHServiceExport(obj runtime.Object, op syncer.Operation) {
+	serviceExportCreated := obj.(*lighthousev2a1.ServiceExport)
+	resourceClient := c.localClient.Resource(schema.GroupVersionResource{Group: "lighthouse.submariner.io",
+		Version: "v2alpha1", Resource: "serviceexport"}).Namespace(serviceExportCreated.Namespace)
+	if resourceClient != nil {
```
tpantelis, line 69: This won't ever be nil.

pkg/agent/controller/agent.go
```diff
@@ -54,15 +55,20 @@ func New(spec *AgentSpecification, cfg *rest.Config) (*Controller, error) {
 		return nil, fmt.Errorf("error creating dynamic client: %v", err)
 	}
 
-	return NewWithDetail(spec, syncerConf, restMapper, localClient, kubeClientSet, scheme.Scheme,
+	lighthouseClient, err := lighthouseClientset.NewForConfig(cfg)
```
tpantelis, line 58: This is no longer needed.

pkg/agent/controller/agent.go
```diff
@@ -54,15 +55,20 @@ func New(spec *AgentSpecification, cfg *rest.Config) (*Controller, error) {
...
+	return NewWithDetail(spec, syncerConf, restMapper, localClient, kubeClientSet, lighthouseClient, scheme.Scheme,
 		func(config *broker.SyncerConfig) (*broker.Syncer, error) {
 			return broker.NewSyncer(*config)
 		})
 }
 
 // Constructor that takes additional detail. This is intended for unit tests.
 func NewWithDetail(spec *AgentSpecification, syncerConf *broker.SyncerConfig, restMapper meta.RESTMapper, localClient dynamic.Interface,
-	kubeClientSet kubernetes.Interface, runtimeScheme *runtime.Scheme,
+	kubeClientSet kubernetes.Interface, lighthouseClient lighthouseClientset.Interface, runtimeScheme *runtime.Scheme,
```
tpantelis, line 71:
```suggestion
	kubeClientSet kubernetes.Interface, runtimeScheme *runtime.Scheme,
```

pkg/agent/controller/agent.go
```diff
@@ -146,6 +152,12 @@ func NewWithDetail(spec *AgentSpecification, syncerConf *broker.SyncerConfig, re
 		return nil, err
 	}
 
+	agentController.serviceLHExportController, err = newLHServiceExportController(localClient, restMapper, runtimeScheme)
```
tpantelis, line 155:
```suggestion
	agentController.lhServiceExportController, err = newLHServiceExportController(localClient, restMapper, runtimeScheme)
```

pkg/agent/controller/lhserviceexport.go
```diff
@@ -0,0 +1,75 @@
...
+			Name:      objMeta.GetName(),
+			Namespace: objMeta.GetNamespace(),
+		},
+	}
+
+	return mcsServiceExport, false
+}
+
+func (c *LHServiceExportController) deleteLHServiceExport(obj runtime.Object, op syncer.Operation) {
+	serviceExportCreated := obj.(*lighthousev2a1.ServiceExport)
```
tpantelis, line 66:
```suggestion
	serviceExportDeleted := obj.(*lighthousev2a1.ServiceExport)
```

tpantelis (changes requested):

pkg/agent/controller/lhserviceexport.go
aswinsuryan: Removed the check.

pkg/agent/controller/agent.go
aswinsuryan: Removed.

pkg/agent/controller/agent.go
aswinsuryan: Removed.

pkg/agent/controller/lhserviceexport.go
aswinsuryan: Done.

pkg/agent/controller/agent.go
aswinsuryan: Done.

pkg/agent/controller/lhserviceexport.go
```diff
@@ -0,0 +1,74 @@
...
+func (c *LHServiceExportController) start(stopCh <-chan struct{}) error {
+	if err := c.lhServiceExportSyncer.Start(stopCh); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *LHServiceExportController) LHtoMCSServiceExport(obj runtime.Object, op syncer.Operation) (runtime.Object, bool) {
+	serviceExportCreated := obj.(*lighthousev2a1.ServiceExport)
```
tpantelis, line 53: We only want to handle creates here. We delete the LH ServiceExport which I think would re-trigger this syncer to delete the MCS ServiceExport.  I think it would be good to add a unit test to cover this.

```
if op != syncer.Create {
    return nil, false
}
```

aswinsuryan: Done. , added UT.

pkg/agent/controller/lhserviceexport_test.go
```diff
@@ -0,0 +1,135 @@
...
+			It("should correctly sync a ServiceImport and update the ServiceExport status", func() {
+				t.createServiceExport()
+				t.awaitNoLHServiceexport(t.cluster1.localLhServiceExportClient)
+				t.awaitMcsService(t.cluster1.localMcsServiceExportClient, t.serviceExport)
+			})
+		})
+	})
+})
+
+func newTestDiver() *testDriver {
```
mangelajo, line 84: nit typo: newTestDiver -> newTestDriver

aswinsuryan:
The operator build requires https://github.com/submariner-io/submariner-operator/pull/848 to pass

tpantelis (approved):

vthapar (approved):

PR #359, opened 2020-10-30
Use TargetRef for hostname

### LINKED ISSUE #358: Use TargetRef instead of hostname for podnames
Currently we're using `hostname` instead of `TargetRef`to determine hostname of pods in a stateful set. `hostname` is not always present as it depends on how labels are set, but `TargetRef` is always present. 

```
subsets:
- addresses:
  - ip: 10.132.2.12
    nodeName: ip-10-0-161-186.us-east-2.compute.internal
    targetRef:
      kind: Pod
      name: rocketchat-mongo-0
      namespace: default
      resourceVersion: "58107"
      uid: a6e70db5-e714-4ce7-93ae-55654099f095
  - ip: 10.133.2.11
    nodeName: ip-10-0-150-224.us-east-2.compute.internal
    targetRef:
      kind: Pod
      name: rocketchat-mongo-1
      namespace: default
      resourceVersion: "58357"
      uid: 316acb75-deed-426e-af85-fb525a14738d
```

So it is much more reliable and consistent to use `TargetRef`

vthapar:
If hostname is not present in endpoint/slice, reuse TargetRef.Name
for the pod hostname.

Fixes #358

### DIFF
--- a/pkg/agent/controller/endpoint.go
+++ b/pkg/agent/controller/endpoint.go
@@ -144,8 +144,18 @@ func endpointFromAddress(address corev1.EndpointAddress, ready bool) discovery.E
 		Topology:   topology,
 	}
 
-	if address.Hostname != "" {
+	/*
+		We only need TargetRef.Name as pod address and hostname are only relevant fields.
+		Avoid copying TargetRef coz it it has revision which can change for reasons other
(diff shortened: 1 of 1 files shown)


mangelajo (approved):
Thanks!!

tpantelis (approved):

PR #364, opened 2020-11-05
Add periodic E2E tests to find flaky failures

dfarrell07:
See commit messages for details.

### DIFF
--- a/.github/workflows/periodic_e2e.yml
+++ b/.github/workflows/periodic_e2e.yml
@@ -1,0 +1,38 @@
+---
+name: Periodic E2E
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  e2e:
+    name: E2E
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        globalnet: ['', 'globalnet']
+        deploytool: ['operator', 'helm']
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+
+      - name: Reclaim free space
+        run: |
+          sudo swapoff -a
+          sudo rm -f /swapfile
+          df -h
+          free -h
+
+      - name: Run E2E deployment and tests
+        run: |
+          make e2e using="${{ matrix.globalnet }} ${{ matrix.deploytool }}"
+
+      - name: Post mortem
+        if: failure()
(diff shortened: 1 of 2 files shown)


dfarrell07:
@skitt @tpantelis @mkolesnik @mangelajo Thinking about this more, does it actually make sense to have these periodic E2E tests on repos other than submariner-io/submariner? Doesn't this run a subset of the same deployments/tests? Having the periodic E2E tests on only one repo would make it easier to tweak the timing and would centralize the data stream.

tpantelis:
> @skitt @tpantelis @mkolesnik @mangelajo Thinking about this more, does it actually make sense to have these periodic E2E tests on repos other than submariner-io/submariner? Doesn't this run a subset of the same deployments/tests? Having the periodic E2E tests on only one repo would make it easier to tweak the timing and would centralize the data stream.

The repos run different tests specific to their functionality.  Perhaps we can run the periodic job in shipyard similar to job that runs e2e on all consuming projects triggered by the PR label.

dfarrell07:
> > @skitt @tpantelis @mkolesnik @mangelajo Thinking about this more, does it actually make sense to have these periodic E2E tests on repos other than submariner-io/submariner? Doesn't this run a subset of the same deployments/tests? Having the periodic E2E tests on only one repo would make it easier to tweak the timing and would centralize the data stream.
> 
> The repos run different tests specific to their functionality. Perhaps we can run the periodic job in shipyard similar to job that runs e2e on all consuming projects triggered by the PR label.

Okay, good idea. I sent a PR to Shipyard: https://github.com/submariner-io/shipyard/pull/358

dfarrell07:
Working through implementing this in Shipyard, I'm starting to think it's better to use this per-repo design.

1. To make the build matrix exactly match the matrixes of the consuming projects, we need a complex set of excludes that will be difficult to maintain.
2. With a more minimal set of excludes, we run 60+ jobs.
3. It's easier to maintain the test matrixes in each project by just making sure the same `using` flags are in the Periodic E2E and E2E workflows, vs keeping the Shipyard matrix up-to-date and having to think about all the projects.
4. To get better observability using per-project tests, we can add GHA step to report an issue on failures. That would be a middle ground between a very easy to see single workflow and this design.

dfarrell07:
Closing in favor of https://github.com/submariner-io/submariner-operator/pull/786.

PR #391, opened 2020-11-26
<Don't Merge> PR to capture more logs for e2e failure.

aswinsuryan:


vthapar:
Comment out headlessservices and statefulsets tests, so we only run service discovery. This will reduce both test run time as well as number of log entries.

pkg/agent/controller/agent.go
```diff
@@ -497,8 +499,11 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, op syncer.Ope
 	labels := endpointSlice.GetObjectMeta().GetLabels()
 
 	if labels[discovery.LabelManagedBy] != lhconstants.LabelValueManagedBy {
+		klog.Infof("Retuning nil since not managed by lighthouse %v", obj)
 		return nil, false
 	}
 
+	klog.Infof("In filterLocalEndpointSlices returning %v", obj)
```
vthapar, line 506: This will generate too many log entries in full run, esp in headless and stateful where we have multiple endpoints.

aswinsuryan:
The root cause is found out, hence closing.

PR #395, opened 2020-12-01
Don't build new image on tag

mkolesnik:
Since we're using the releases repository now, it'll be doing the
tagging, so no need to rebuild the image when a tag is created.

Also updated the GHA to be in line with Shipyard's.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,8 +5,6 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


mangelajo (commented):
Sense makes this my friend

mangelajo (approved):

skitt (approved):

aswinsuryan (approved):

PR #405, opened 2020-12-09
Add flaky test finder periodic E2E tests, temporarily run frequently

dfarrell07:
See commit messages for details.

### DIFF
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -3,7 +3,7 @@ name: Flake Finder
 
 on:
   schedule:
-    - cron: "0 0,1 * * *"
+    - cron: "0 0/2 * * *"
 
 jobs:
   e2e:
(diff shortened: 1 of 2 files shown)


.github/workflows/flake_finder.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: Flake Finder
+
+on:
+  schedule:
+    - cron: "0 0/2 * * *"
```
vthapar, line 6: How about hourly runs but during off-peak hours? We'll get similar results without taking CI slots during work hours.

vthapar (approved):

.github/workflows/flake_finder.yml
dfarrell07: I don't think we'll overwhelm the CI slots with this, it's only 4 jobs (smaller matrix than the operator). I'd definitely like to turn it down to two runs per day off-peak in a few days, after RC1.

aswinsuryan (approved):

PR #419, opened 2020-12-18
Fix typo in e2e logs

sridhargaddam:
When running LH e2e tests, we see the following logs
When ServiceDiscovery is expected to succeed:
STEP: Validating that dig result is "100.3.162.213"

When ServiceDiscovery is expected to fail:
STEP: Validating that dig result is not ""

The above log is confusing and should actually be
STEP: Validating that dig result is not "100.3.162.213"

Signed-Off-by: Sridhar Gaddam <sgaddam@redhat.com>

### DIFF
--- a/test/e2e/discovery/service_discovery.go
+++ b/test/e2e/discovery/service_discovery.go
@@ -304,7 +304,7 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
 		return stdout, nil
 	}, func(result interface{}) (bool, string, error) {
 		doesContain := strings.Contains(result.(string), serviceIP)
-		By(fmt.Sprintf("Validating that dig result %s %q", op, result))
+		By(fmt.Sprintf("Validating that dig result %q %s %q", result, op, serviceIP))
 		if doesContain && !shouldContain {
 			return false, fmt.Sprintf("expected execution result %q not to contain %q", result, serviceIP), nil
 		}


test/e2e/discovery/service_discovery.go
```diff
@@ -304,7 +304,7 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
 		return stdout, nil
 	}, func(result interface{}) (bool, string, error) {
 		doesContain := strings.Contains(result.(string), serviceIP)
-		By(fmt.Sprintf("Validating that dig result %s %q", op, result))
+		By(fmt.Sprintf("Validating that dig result %s %q", op, serviceIP))
```
vthapar, line 307:
```suggestion
		By(fmt.Sprintf("Validating that dig result %s %q %s", result, op, serviceIP))
```

vthapar: We need to capture result too. It lets us know what is the actual result returned, which is useful when we run into test failures.

vthapar (changes requested):

test/e2e/discovery/service_discovery.go
sridhargaddam: Yeah, it will be helpful.

aswinsuryan (approved):

tpantelis (approved):

dfarrell07 (approved):

skitt (approved):

vthapar (approved):

Issue #340, opened 2020-10-14
[E2E] Test service DNS on mutiple clusters returns diffrent IPs (Round Robin)


manosnoam:
Add test in E2E: 
When exporting same service name (with same DNS) on multiple clusters - It should return a different IP each request, according to the clusterIP (and Round Robin algorithm).

_Originally posted by @danibachar in https://github.com/submariner-io/lighthouse/issue_comments/703122297_

aswinsuryan:
We have UT for this scenario but e2e is missing and we need to add it.

nyechiel:
@aswinsuryan as discussed, this particular test will need to be skipped if 3 clusters are not available.

PR #422, opened 2020-12-21
Add e2e to verify round-robin for Service discovery

aswinsuryan:
Fixes : #340

### DIFF
--- a/test/e2e/discovery/service_discovery.go
+++ b/test/e2e/discovery/service_discovery.go
@@ -17,6 +17,7 @@ package discovery
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
@@ -69,10 +70,11 @@ var _ = Describe("[discovery] Test Service Discovery Across Clusters", func() {
 		})
 	})
 
-
 	When("a pod tries to resolve a service in a specific remote cluster by its cluster name", func() {
 		It("should resolve the service on the specified cluster", func() {
 			RunServiceDiscoveryClusterNameTest(f)
+		})
+	})
 
 	When("a pod tries to resolve a service multiple times", func() {
 		It("should resolve the service from both the clusters in a round robin fashion", func() {
@@ -488,10 +490,11 @@ func verifyRoundRobinWithDig(f *framework.Framework, srcCluster framework.Cluste
 		cmd = append(cmd, serviceName+"."+f.Namespace+".svc."+domains[i])
 	}
 
-	op := "is"
 	serviceIPMap := make(map[string]int)
 
-	By(fmt.Sprintf("Executing %q to verify IPs %q for service %q %q discoverable", strings.Join(cmd, " "), serviceIPList, serviceName, op))
+	By(fmt.Sprintf("Executing %q to verify IPs %q for service %q is discoverable", strings.Join(cmd, " "), serviceIPList, serviceName))
+
+	var retIPs []string
 
 	for count := 0; count < 10; count++ {
 		framework.AwaitUntil("verify if service IP is discoverable", func() (interface{}, error) {
@@ -509,14 +512,14 @@ func verifyRoundRobinWithDig(f *framework.Framework, srcCluster framework.Cluste
 
 			return stdout, nil
 		}, func(result interface{}) (bool, string, error) {
-			var doesContain bool
-			for _, serviceIp := range serviceIPList {
-				doesContain = strings.Contains(result.(string), serviceIp)
-				if doesContain {
-					serviceIPMap[serviceIp]++
+			for _, serviceIP := range serviceIPList {
+				if strings.Contains(result.(string), serviceIP) {
+					serviceIPMap[serviceIP]++
+					retIPs = append(retIPs, serviceIP)
+					break
 				}
 			}
-			By(fmt.Sprintf("Validating that dig result %s %q", op, result))
+
 			return true, "", nil
 		})
 	}
@@ -525,7 +528,9 @@ func verifyRoundRobinWithDig(f *framework.Framework, srcCluster framework.Cluste
 		" IP %q which was returned %d times is within the threshold", serviceIPList[0], serviceIPMap[serviceIPList[0]],
 		serviceIPList[1], serviceIPMap[serviceIPList[1]]))
 
-	Expect(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]]%10 < 5).Should(BeTrue())
+	Expect(int(math.Abs(float64(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]]))) < 3).To(BeTrue(),
+		"Service IPs were not returned in proper round-robin fashion: Expected IPs: %v,"+
+			" Returned IPs: %v, IP Counts: %v", serviceIPList, retIPs, serviceIPMap)
 }
 
 func getClusterDomain(f *framework.Framework, cluster framework.ClusterIndex, targetPod *corev1.PodList) string {
--- a/test/e2e/discovery/service_discovery.go
+++ b/test/e2e/discovery/service_discovery.go
@@ -416,8 +416,7 @@ func RunServiceDiscoveryRoundRobinTest(f *lhframework.Framework) {
 
 	serviceIPList = append(serviceIPList, serviceIPClusterC)
 
-	verifyRoundRobinWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB.Name, serviceIPList, netshootPodList, checkedDomains,
-		"", true)
+	verifyRoundRobinWithDig(f.Framework, framework.ClusterA, nginxServiceClusterB.Name, serviceIPList, netshootPodList, checkedDomains)
 }
 
 func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster framework.ClusterIndex, service *corev1.Service,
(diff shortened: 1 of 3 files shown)


nyechiel:
@aswinsuryan as discussed, this particular test will need to be skipped if 3 clusters are not available.

skitt (approved):

test/e2e/discovery/service_discovery.go
```diff
@@ -317,6 +387,62 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
+			By(fmt.Sprintf("Validating that dig result %s %q", op, result))
+			return true, "", nil
+		})
+	}
+
+	By(fmt.Sprintf("Validating that difference between the service IP %q which was returned %d times and  the service"+
+		" IP %q which was returned %d times is within the threshold", serviceIPList[0], serviceIPMap[serviceIPList[0]],
+		serviceIPList[1], serviceIPMap[serviceIPList[1]]))
+
+	Expect(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]]%10 < 5).Should(BeTrue())
```
tpantelis, line 443: I'm not clear on the algorithm/calculation here. We expect the IPs to be interleaved, ie IP1, IP2, IP1, IP2 etc.  If all 10 requests returned IP1, then the calculation `(10 - 0) % 10 = 0` would be < 5 which would pass, as would `countIP1 = 7` and `countIP2 = 3`.  Even if both IPs were returned 5 times, this wouldn't verify they were properly interleaved. I suggest:

```
    retIPs := make([]string, 0)
    seenIPs := stringset.New()
    ipCount := len(serviceIPs)
    totalReqCount := ipCount * 3
    for count := 0; count < totalReqCount; count++ {
        framework.AwaitUntil("verify round-robin service IP retrieval", func() (interface{}, error) {
	    ...
	    output := result.(string)
	    isValidIP := false
	    for _, serviceIP := range serviceIPs {
	        if strings.Contains(output, serviceIP) {
		    if count < ipCount && !seenIPs.Add(serviceIP) {
		        return false, "", fmt.Errorf("dig returned IP %s out of round-robin order - prior IPs %v",
			    serviceIP, seenIPs.Elements())
		    }

		    retIPs = append(retIPs, serviceIP)
		    isValidIP = true
		    break
		}
             }

	    if !isValidIP {
	        return false, "", fmt.Errorf("dig result %q does not contain any expected IP %v", output, serviceIPs)
	    }   

	    return true, "", nil
	})
    }

    rrIPs := retIPs[0:ipCount]
    for i := ipCount; i < totalReqCount; i += ipCount {
        Expect(retIPs[i:i+ipCount]).To(Equal(rrIPs), "Service IPs returned out of round-robin order: %v", retIPs)
    }

```

tpantelis (changes requested):

test/e2e/discovery/service_discovery.go
aswinsuryan: It may not return in exact round-robin order as there are two lighthouse core-dns instance running and any one of them could return the service IP for a  DNS request. Each of them maintains it own data structures. That is the reason for comparing it with a threshold.

test/e2e/discovery/service_discovery.go
```diff
@@ -267,6 +273,69 @@ func RunServicesPodAvailabilityMutliClusterTest(f *lhframework.Framework) {
...
+
+	f.AwaitGlobalnetIP(framework.ClusterC, nginxServiceClusterC.Name, nginxServiceClusterC.Namespace)
+	f.NewServiceExport(framework.ClusterC, nginxServiceClusterC.Name, nginxServiceClusterC.Namespace)
+
+	By(fmt.Sprintf("Creating a Netshoot Deployment on %q", clusterAName))
+
+	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
+
+	if svc, err := f.GetService(framework.ClusterC, nginxServiceClusterC.Name, nginxServiceClusterC.Namespace); err == nil {
+		nginxServiceClusterC = svc
```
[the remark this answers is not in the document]
aswinsuryan: This pattern is used  to check if there is an annotation added by global net and if so expect that IP . But I think you are right , we should fail on error . I can raise another patch to correct it in all the places.

test/e2e/discovery/service_discovery.go
```diff
@@ -317,6 +387,62 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
 	})
 }
 
+func verifyRoundRobinWithDig(f *framework.Framework, srcCluster framework.ClusterIndex, serviceName string, serviceIPList []string,
+	targetPod *corev1.PodList, domains []string, clusterName string, shouldContain bool) {
```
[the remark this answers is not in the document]
aswinsuryan: Removed.

test/e2e/discovery/service_discovery.go
tpantelis: OK. In talking with Vishal:

- we should take the absolute value of the difference, ie `int(math.Abs(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]])( % 10`
- to avoid `(10 - 0) % 10 = 0` from passing, mod by 11 instead
- use threshold 3 instead of 5 as the most times the same IP should be returned out of 10 requests is 6 - so 7 out of 10 should fail (`(7-3) % 10 = 4`).

test/e2e/discovery/service_discovery.go
```diff
@@ -332,6 +406,54 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
 
+func verifyRoundRobinWithDig(f *framework.Framework, srcCluster framework.ClusterIndex, serviceName string, serviceIPList []string,
+	targetPod *corev1.PodList, domains []string) {
+	cmd := []string{"dig", "+short"}
+
+	for i := range domains {
+		cmd = append(cmd, serviceName+"."+f.Namespace+".svc."+domains[i])
+	}
+
+	op := "is"
```
tpantelis, line 417: We don't need this.

test/e2e/discovery/service_discovery.go
```diff
@@ -332,6 +406,54 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
+		}, func(result interface{}) (bool, string, error) {
+			var doesContain bool
+			for _, serviceIp := range serviceIPList {
+				doesContain = strings.Contains(result.(string), serviceIp)
+				if doesContain {
+					serviceIPMap[serviceIp]++
+				}
+			}
+			By(fmt.Sprintf("Validating that dig result %s %q", op, result))
+			return true, "", nil
```
tpantelis, line 446: I think we should record the returned IPs in an array so we can report it on failure:

 `retIPs := make([]string, 0)`
 
```suggestion
			for _, serviceIP := range serviceIPList {
				if strings.Contains(result.(string), serviceIP) {
					serviceIPMap[serviceIP]++
					retIPs = append(retIPs, serviceIP)
					break
				}
			}
			
			return true, "", nil
```

test/e2e/discovery/service_discovery.go
```diff
@@ -332,6 +406,54 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
+		})
+	}
+
+	By(fmt.Sprintf("Validating that difference between the service IP %q which was returned %d times and  the service"+
+		" IP %q which was returned %d times is within the threshold", serviceIPList[0], serviceIPMap[serviceIPList[0]],
+		serviceIPList[1], serviceIPMap[serviceIPList[1]]))
+
+	Expect(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]]%10 < 5).Should(BeTrue())
```
tpantelis, line 454:
```suggestion
	Expect(serviceIPMap[serviceIPList[0]]-serviceIPMap[serviceIPList[1]]%10 < 5).To(BeTrue(), "Service IPs were not returned in proper round-robin fashion: Expected IPs: %v, Returned IPs: %v, IP Counts: %v", serviceIPList, retIPs, serviceIPMap)
```

tpantelis (changes requested):

test/e2e/discovery/service_discovery.go
aswinsuryan: we require either "%" or Abs. I just assumed % would always return a positive value. Now changed to use abs.

test/e2e/discovery/service_discovery.go
```diff
@@ -406,6 +482,57 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
+	targetPod *corev1.PodList, domains []string) {
+	cmd := []string{"dig", "+short"}
+
+	for i := range domains {
+		cmd = append(cmd, serviceName+"."+f.Namespace+".svc."+domains[i])
+	}
+
+	serviceIPMap := make(map[string]int)
+
+	By(fmt.Sprintf("Executing %q to verify IPs %q for service %q is discoverable", strings.Join(cmd, " "), serviceIPList, serviceName))
```
tpantelis, line 495:
```suggestion
	By(fmt.Sprintf("Executing %q to verify IPs %q for service %q are discoverable in a round-robin fashion", strings.Join(cmd, " "), serviceIPList, serviceName))
```

test/e2e/discovery/service_discovery.go
```diff
@@ -406,6 +482,57 @@ func verifyServiceIpWithDig(f *framework.Framework, srcCluster, targetCluster fr
...
+					break
+				}
+			}
+
+			return true, "", nil
+		})
+	}
+
+	By(fmt.Sprintf("Validating that difference between the service IP %q which was returned %d times and  the service"+
+		" IP %q which was returned %d times is within the threshold", serviceIPList[0], serviceIPMap[serviceIPList[0]],
```
tpantelis, line 528:
```suggestion
	By(fmt.Sprintf("Service IP %q was returned %d times and Service IP %q was returned %d times - verifying the difference between them is within the threshold", serviceIPList[0], serviceIPMap[serviceIPList[0]],
```

tpantelis (approved):

PR #430, opened 2021-01-08
Fix issues in e2e

aswinsuryan:
Changed the RunServicesPodAvailabilityMutliClusterTest to use the third
cluster. Since the local cluster is always given preference due to a recent
change this test was not validating the intended scenario.

Added check to fail on error were globalnet ip is fetched.


tpantelis (approved):

skitt (approved):

dfarrell07:
This was backported in https://github.com/submariner-io/lighthouse/pull/450.

PR #438, opened 2021-01-19
Enable OVN CNI

vthapar:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,6 +14,12 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         deploytool: ['operator', 'helm']
+        ovn: ['', 'ovn']
+        exclude:
+          - ovn: 'ovn'
+            deploytool: 'helm'
+          - ovn: 'ovn'
+            globalnet: 'globalnet'
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


scripts/cluster_settings
```diff
@@ -1,12 +1,11 @@
...
 
 # We need a minimal setup to verify that lighthouse works
-clusters=('cluster1' 'cluster2' 'cluster3')
+clusters=('cluster1' 'cluster2')
 cluster_nodes['cluster1']="control-plane worker worker"
 cluster_nodes['cluster2']="control-plane worker worker"
-cluster_nodes['cluster3']="control-plane worker"
 
-cluster_cni=( ['cluster1']="weave" ['cluster2']="weave" ['cluster3']="weave" )
+cluster_cni=( ['cluster1']="weave" ['cluster2']="weave" )
```
aswinsuryan, line 8: we need the cluster3 for round robin and healthcheck test cases.

vthapar: No plan to merge this as lighthouse doesn't care about CNI. This is more of a troubleshooting PR to try and see why it is not working. Wanted to check if using 2 clusters instead of 3 works.

mangelajo: Whall we avoid using OVN altogether here?, it should be independent.

tpantelis:
Is this still relevant?

PR #443, opened 2021-02-01
Include port numbers in ServiceImports

### LINKED ISSUE #385: In ServiceImports include port number(s) of the services
Currently, ServiceImports populated by submariner only includes IP addresses of the services.
However it would be good to include even the Port number(s) of services.
Port numbers are already part of the ServiceImport objecgt.

**Why is this needed**:
I am trying to see if we can somehow integrate OSM (openservice-mesh by Microsoft) with Submariner.
OSM `learns' about k8s services and programs envoy. Here, k8s is a `provider' and it provides a set of services (ServiceName, IP, Ports) to OSM within the cluster.

I was thinking of Submariner as another `provider' and can feed similar information to OSM about remote clusters.
So that services across clusters can seamlessly work.

vthapar:
Fixes #385

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -238,7 +238,7 @@ func (a *Controller) serviceExportToServiceImport(obj runtime.Object, numRequeue
 		SessionAffinityConfig: new(corev1.SessionAffinityConfig),
 	}
 
-	ips, err := a.getIPsForService(svc, svcType)
+	ips, ports, err := a.getIPsAndPortsForService(svc, svcType)
 	if err != nil {
 		// Failed to get ips for some reason, requeue
 		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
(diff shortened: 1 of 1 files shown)


aswinsuryan (commented):
LGTM

pkg/agent/controller/agent.go
```diff
@@ -422,27 +423,35 @@ func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1
 	}
 }
 
-func (a *Controller) getIPsForService(service *corev1.Service, siType mcsv1a1.ServiceImportType) ([]string, error) {
+func (a *Controller) getIPsAndPortsForService(service *corev1.Service, siType mcsv1a1.ServiceImportType) (
+	[]string, []mcsv1a1.ServicePort, error) {
+	mcsPort := mcsv1a1.ServicePort{}
+	if service.Spec.Ports != nil {
```
tpantelis, line 429:
```suggestion
	if len(service.Spec.Ports) > 0 {
```

tpantelis (changes requested):

dfarrell07 (approved):

tpantelis:
We should add a unit test case for this (w/ and w/o a port).  @vthapar I can do this in a followup if you want.

vthapar:
> We should add a unit test case for this (w/ and w/o a port). @vthapar I can do this in a followup if you want.

That would be great. I was planning to do follow up but after I'm done with the submariner serviceexport changes.

tpantelis:
> > We should add a unit test case for this (w/ and w/o a port). @vthapar I can do this in a followup if you want.
> 
> That would be great. I was planning to do follow up but after I'm done with the submariner serviceexport changes.

OK I'll followup with UTs.

tpantelis (approved):

dfarrell07:
This was backported in https://github.com/submariner-io/lighthouse/pull/450.

PR #452, opened 2021-02-04
Add Multiple ports to ServiceImport

vthapar:
If service has multiple ports, add them all to ServiceImport,
not just first one.

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -427,14 +427,12 @@ func (a *Controller) getIPsAndPortsForService(service *corev1.Service, siType mc
 	[]string, []mcsv1a1.ServicePort, error) {
 	var mcsPorts []mcsv1a1.ServicePort
 
-	if len(service.Spec.Ports) > 0 {
-		mcsPorts = []mcsv1a1.ServicePort{
-			{
-				Name:     service.Spec.Ports[0].Name,
-				Protocol: service.Spec.Ports[0].Protocol,
-				Port:     service.Spec.Ports[0].Port,
-			},
-		}
+	for _, port := range service.Spec.Ports {
+		mcsPorts = append(mcsPorts, mcsv1a1.ServicePort{
+			Name:     port.Name,
+			Protocol: port.Protocol,
+			Port:     port.Port,
+		})
 	}
 
 	if siType == mcsv1a1.ClusterSetIP {
--- a/pkg/agent/controller/agent_test.go
+++ b/pkg/agent/controller/agent_test.go
@@ -182,6 +182,11 @@ var _ = Describe("ServiceImport syncing", func() {
 					Protocol: corev1.ProtocolTCP,
 					Port:     123,
 				},
+				{
+					Name:     "eth1",
+					Protocol: corev1.ProtocolTCP,
+					Port:     1234,
+				},
 			}
 		})
 


skitt (changes requested):
Any chance this could have an associated unit test?

tpantelis:
> Any chance this could have an associated unit test?

yeah https://github.com/submariner-io/lighthouse/pull/451 adds test for service ports. We should merge that first and add another test case here.

pkg/agent/controller/agent.go
```diff
@@ -425,11 +425,14 @@ func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1
...
-		mcsPort.Protocol = service.Spec.Ports[0].Protocol
-		mcsPort.Port = service.Spec.Ports[0].Port
+	mcsPorts := []mcsv1a1.ServicePort{}
+
+	for _, port := range service.Spec.Ports {
+		mcsPort := mcsv1a1.ServicePort{}
+		mcsPort.Name = port.Name
+		mcsPort.Protocol = port.Protocol
+		mcsPort.Port = port.Port
+		mcsPorts = append(mcsPorts, mcsPort)
```
tpantelis:
```suggestion
		mcsPorts = append(mcsPorts, mcsv1a1.ServicePort {
		    Name:     port.Name,
		    Protocol: port.Protocol,
		    Port:        port.Port,
		})
```

tpantelis (changes requested):

pkg/agent/controller/agent.go
```diff
@@ -425,11 +425,14 @@ func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1
 
 func (a *Controller) getIPsAndPortsForService(service *corev1.Service, siType mcsv1a1.ServiceImportType) (
 	[]string, []mcsv1a1.ServicePort, error) {
-	mcsPort := mcsv1a1.ServicePort{}
-	if len(service.Spec.Ports) > 0 {
-		mcsPort.Name = service.Spec.Ports[0].Name
-		mcsPort.Protocol = service.Spec.Ports[0].Protocol
-		mcsPort.Port = service.Spec.Ports[0].Port
+	mcsPorts := []mcsv1a1.ServicePort{}
```
tpantelis: So we return nil if there's no ports.

```suggestion
	var mcsPorts []mcsv1a1.ServicePort
```

pkg/agent/controller/agent.go
```diff
@@ -451,7 +454,7 @@ func (a *Controller) getIPsAndPortsForService(service *corev1.Service, siType mc
 		return make([]string, 0), make([]mcsv1a1.ServicePort, 0), nil
```
tpantelis:
```suggestion
		return make([]string, 0), nil, nil
```

tpantelis (changes requested):

pkg/agent/controller/agent.go
vthapar:
```
type ServiceImportSpec struct {
	// +listType=atomic
	Ports []ServicePort `json:"ports"`
```
`ServicePort` is not `omitempty` but individual fields are. So we want it to be non nil.

tpantelis:
> > Any chance this could have an associated unit test?
> 
> yeah #451 adds test for service ports. We should merge that first and add another test case here.

@vthapar since #451 was merged, can you change the service port test case to have 2 ports to cover these changes?

aswinsuryan (approved):

pkg/agent/controller/agent_test.go
```diff
@@ -182,6 +182,11 @@ var _ = Describe("ServiceImport syncing", func() {
 					Protocol: corev1.ProtocolTCP,
 					Port:     123,
 				},
+				{
```
skitt, line 185: AFAICT, the test doesn’t actually verify that the ports have been copied, does it?

skitt (changes requested):

vthapar:
> Any chance this could have an associated unit test?

Done

pkg/agent/controller/agent_test.go
vthapar: It does. Tom already added that check as part of UT
```
	Expect(serviceImport.Spec.Ports).To(HaveLen(len(service.Spec.Ports)))

	for i := range service.Spec.Ports {
		Expect(serviceImport.Spec.Ports[i].Name).To(Equal(service.Spec.Ports[i].Name))
		Expect(serviceImport.Spec.Ports[i].Protocol).To(Equal(service.Spec.Ports[i].Protocol))
		Expect(serviceImport.Spec.Ports[i].Port).To(Equal(service.Spec.Ports[i].Port))
	}
```

skitt (approved):

tpantelis (approved):

PR #457, opened 2021-02-10
Handle verbosity settings internally

skitt:
Instead of relying on a shell script to handle verbosity settings from
environment variables, handle them in the agent itself; this allows
the container image to be reduced to the binary only.


pkg/agent/main.go
```diff
@@ -40,6 +42,17 @@ var (
 func main() {
 	agentSpec := controller.AgentSpecification{}
 
+	// Handle environment variables:
+	// LIGHTHOUSE_VERBOSITY determines the verbosity level
+	// LIGHTHOUSE_DEBUG, if set to true, sets the verbosity level to 3
+	if debug := os.Getenv("LIGHTHOUSE_DEBUG"); debug != "" {
+		os.Args = append(os.Args, "-v=3")
+	} else if verbosity := os.Getenv("LIGHTHOUSE_VERBOSITY"); verbosity != "" {
```
vthapar: Can we just go with "SUBMARINER_" like with metrics? I think we have a SUBMARINER_DEBUG flag in our CRD, so we'll be able to reuse that, right?

skitt: Done. CRD handling has to be done in the operator, but at least if the variables are the same everywhere, it will be easier to understand!

tpantelis (approved):

mkolesnik (approved):
LGTM

mangelajo (approved):
awesome!! :) I love scratch images

Issue #349, opened 2020-10-21
Lighthouse Prometheus Metrics


maayanf24:
Add prometheus metrics for lighthouse
See here for details: https://github.com/submariner-io/submariner/issues/678

- [x] Number of imported services, number of exported services (probably doable as a metric in the resyncer, see https://github.com/submariner-io/admiral/issues/136) - [#468](https://github.com/submariner-io/lighthouse/pull/468)

- [ ] Health of a service in a remote cluster ( the number of active pods) 

- [ ] DNS results (query for A record mapped to IP address) (CoreDNS?)

- [ ] Who is accessing what and how frequently?

nyechiel:
@maayanf24 just a reminder to also update the docs with the newly added Service Discovery metrics

tpantelis:
bump

nyechiel:
Closing old issue. Let's re-open if this is relevant and/or we have concrete user demand.

PR #468, opened 2021-02-23
Add Service Import and Service Export metrics

maayanf24:
Related issue #349

Signed-off-by: Maayan Friedman <maafried@redhat.com>

### DIFF
--- a/pkg/agent/controller/agent_test.go
+++ b/pkg/agent/controller/agent_test.go
@@ -16,8 +16,10 @@ limitations under the License.
 package controller_test
 
 import (
+	"crypto/rand"
 	"errors"
 	"fmt"
+	"math/big"
 	"reflect"
 	"sort"
 	"time"
@@ -512,7 +514,20 @@ func (c *cluster) init(syncerConfig broker.SyncerConfig) {
 
 func (c *cluster) start(t *testDriver, syncerConfig broker.SyncerConfig) {
 	syncerConfig.LocalClient = c.localDynClient
-	agentController, err := controller.New(&c.agentSpec, syncerConfig, c.localKubeClient)
+	bigint, err := rand.Int(rand.Reader, big.NewInt(1000000))
+	Expect(err).To(Succeed())
+
+	serviceImportCounterName := "submariner_service_import" + bigint.String()
+
+	bigint, err = rand.Int(rand.Reader, big.NewInt(1000000))
+	Expect(err).To(Succeed())
+
+	serviceExportCounterName := "submariner_service_export" + bigint.String()
+
+	agentController, err := controller.New(&c.agentSpec, syncerConfig, c.localKubeClient,
+		controller.AgentConfig{
+			ServiceImportCounterName: serviceImportCounterName,
+			ServiceExportCounterName: serviceExportCounterName})
 
 	Expect(err).To(Succeed())
 	Expect(agentController.Start(t.stopCh)).To(Succeed())
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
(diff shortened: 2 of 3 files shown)


maayanf24:
@skitt @tpantelis @mangelajo @sridhargaddam @nyechiel  - what do you think about prefix the metrics with "submariner_" for the lighthouse?
Should it be prefixed with "lighthouse_" instead?

nyechiel:
> @skitt @tpantelis @mangelajo @sridhargaddam @nyechiel - what do you think about prefix the metrics with "submariner_" for the lighthouse?
> Should it be prefixed with "lighthouse_" instead?

For context, see https://github.com/submariner-io/submariner/pull/1143

Either option should work fine. Technically, users can run Lighthouse without Submariner, but I think that the majority will run both (and even if not, Lighthouse is still a repo under Submariner) - so for simplicity I am leaning towards "submariner_".

tpantelis:
> > @skitt @tpantelis @mangelajo @sridhargaddam @nyechiel - what do you think about prefix the metrics with "submariner_" for the lighthouse?
> > Should it be prefixed with "lighthouse_" instead?
> 
> For context, see [submariner-io/submariner#1143](https://github.com/submariner-io/submariner/pull/1143)
> 
> Either option should work fine. Technically, users can run Lighthouse without Submariner, but I think that the majority will run both (and even if not, Lighthouse is still a repo under Submariner) - so for simplicity I am leaning towards "submariner_".

I agree with "_submariner__" - "_lighthouse_" is just a repo name.  If we were to separate it from submariner then I'd suggest "_service_discovery_".

aswinsuryan:
> > > @skitt @tpantelis @mangelajo @sridhargaddam @nyechiel - what do you think about prefix the metrics with "submariner_" for the lighthouse?
> > > Should it be prefixed with "lighthouse_" instead?
> > 
> > 
> > For context, see [submariner-io/submariner#1143](https://github.com/submariner-io/submariner/pull/1143)
> > Either option should work fine. Technically, users can run Lighthouse without Submariner, but I think that the majority will run both (and even if not, Lighthouse is still a repo under Submariner) - so for simplicity I am leaning towards "submariner_".
> 
> I agree with "_submariner__" - "_lighthouse_" is just a repo name. If we were to separate it from submariner then I'd suggest "_service_discovery_".

+1

skitt (approved):

vthapar:
@skitt @aswinsuryan @maayanf24 Changes look good but I have a question. Should ServiceExport/Import of same service in different clusters be considered as single or multiple?

pkg/agent/controller/agent_test.go
```diff
@@ -512,7 +514,20 @@ func (c *cluster) init(syncerConfig broker.SyncerConfig) {
...
+
+	bigint, err = rand.Int(rand.Reader, big.NewInt(1000000))
+	Expect(err).To(Succeed())
+
+	serviceExportCounterName := "submariner_service_export" + bigint.String()
+
+	agentController, err := controller.New(&c.agentSpec, syncerConfig, c.localKubeClient,
+		controller.AgentConfig{
+			ServiceImportCounterName: serviceImportCounterName,
+			ServiceExportCounterName: serviceExportCounterName})
```
tpantelis, line 530: Can we verify that the metrics are updated where appropriate?

maayanf24:
> @skitt @aswinsuryan @maayanf24 Changes look good but I have a question. Should ServiceExport/Import of same service in different clusters be considered as single or multiple?

@vthapar - each cluster only show local perspective of ServiceExport/Import

pkg/agent/controller/agent_test.go
mangelajo: I believe this patch does not bump those metrics, at least I don't see the code

mangelajo (commented):
> > @skitt @tpantelis @mangelajo @sridhargaddam @nyechiel - what do you think about prefix the metrics with "submariner_" for the lighthouse?
> > Should it be prefixed with "lighthouse_" instead?
> 
> For context, see [submariner-io/submariner#1143](https://github.com/submariner-io/submariner/pull/1143)
> 
> Either option should work fine. Technically, users can run Lighthouse without Submariner, but I think that the majority will run both (and even if not, Lighthouse is still a repo under Submariner) - so for simplicity I am leaning towards "submariner_".

I had exactly the same train of thought ':D

pkg/agent/controller/agent_test.go
skitt: They should be updated by the syncer (in Admiral).

tpantelis: true but for white box testing we're not supposed to know those details. Regardless, adding some counter verification would at least verify that the admiral components were setup correctly. We could also do this in E2E instead. But we could do it in another PR if folks think it's worth it.

mangelajo:
Oh I see it does actually use those prometheus counters from the admiral resource syncer, awesome.

mangelajo (approved):

mangelajo:
Hey @maayanf24 sorry, this needs a rebase due to conflicts at this point.

tpantelis (approved):

PR #478, opened 2021-03-01
Fix issue with Endpointslice delete

aswinsuryan:
Fixes : lighthouse/#475

### DIFF
--- a/pkg/agent/controller/endpoint.go
+++ b/pkg/agent/controller/endpoint.go
@@ -7,10 +7,13 @@ import (
 	lhconstants "github.com/submariner-io/lighthouse/pkg/constants"
 	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


pkg/agent/controller/endpoint.go
```diff
@@ -87,7 +109,7 @@ func (e *EndpointController) endpointSliceFromEndpoints(endpoints *corev1.Endpoi
 		lhconstants.LabelSourceName:        e.serviceName,
 	}
 	endpointSlice.OwnerReferences = []metav1.OwnerReference{{
-		APIVersion:         "lighthouse.submariner.io.v2alpha1",
+		APIVersion:         "multicluster.x-k8s.io.v1alpha1",
```
tpantelis, line 112: Do we need the `OwnerReference` since we have to manually delete anyway?

tpantelis (commented):
Can you update the unit tests? I think you just need to remove L885 in _agent_test.go_.

pkg/agent/controller/endpoint.go
vthapar: > Do we need the `OwnerReference` since we have to manually delete anyway?

Useful when we later get ServiceImport aggregation.

vthapar (approved):

pkg/agent/controller/endpoint.go
aswinsuryan: Also, we set the controller flag to false here, so that the endpointslice controller will not  modify/delete this object

aswinsuryan:
PR on devel https://github.com/submariner-io/lighthouse/pull/480

mangelajo (approved):

tpantelis:
Let's merge this - I'll follow-up with UT changes.

PR #482, opened 2021-03-04
Reclaim free space also during image release

mangelajo:
Signed-off-by: Miguel Angel Ajo <majopela@redhat.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,8 +18,10 @@ jobs:
 
       - name: Reclaim free space
         run: |
-          sudo swapoff -a
-          sudo rm -f /swapfile
+          # cleanup 30GB+ of tools we don't need for our CI
+          rm -rf /opt/ghc /usr/share/dotnet /usr/share/swift
+          df -h
+          free -h
 
       - name: Run E2E deployment and tests
         run: make e2e


aswinsuryan (approved):

sridhargaddam (approved):

vthapar (approved):

.github/workflows/release.yml
```diff
@@ -18,8 +18,10 @@ jobs:
...
         run: |
-          sudo swapoff -a
-          sudo rm -f /swapfile
+          # cleanup 30GB+ of tools we don't need for our CI
+          rm -rf /opt/ghc /usr/share/dotnet /usr/share/swift
+          df -h
+          free -h
 
       - name: Run E2E deployment and tests
         run: make e2e
```
mkolesnik, line 27:
```suggestion
      - name: Run E2E deployment and tests
        uses: submariner-io/shipyard/gh-actions/e2e@devel
```

Why not use shared GHA?

Also you can do the same for post mortem

```
      - name: Post mortem
        if: failure()
        uses: submariner-io/shipyard/gh-actions/post-mortem@devel
```

mkolesnik (changes requested):

.github/workflows/release.yml
mangelajo: This is for the release-0.8 branch only.  I didn't want to introduce unstabilities in there, because those actions could change or become incompatible with the older shipyard.

mkolesnik: Ah right sorry I missed the base branch.

Though I do see release shared GHA was backported to 0.8, so maybe it's a good idea to backport shared GHAs generally since its a CI infra thing.

Generally I think we should keep "stable branch only" change to a minimum, even on CI, since it has the possibility to cause conflicts when backporting changes later on.

mangelajo: This change is equal to what we do on the pull-requests which was backported from master, but this change never hit master on the release.yaml.

I think at this point it's better to follow the path of this patch and avoid wasting time in backporting gh-actions to the release branch.

Let's focus on the supportability job for the rest of the sprint.

What you are suggesting will come naturally when we branch release-0.9

Then at that point we can probably switch the GHA pointers to @release-0.9 for more safety so the gh-actions & shipyard base image are in parity.

WDYT?

mkolesnik: OK for now lets go with this change hopefully in releas-0.9 it will be better

mkolesnik (approved):

.github/workflows/release.yml
mangelajo: Thank you Mike!

PR #486, opened 2021-03-08
Add IPAM initialism

skitt:
The defaults come from https://staticcheck.io/docs/#configuration


staticcheck.conf
```diff
@@ -0,0 +1,15 @@
+initialisms = ["ACL", "API", "ASCII",
+	"CPU", "CSS",
```
Jaanki: Add CNI as well?

Jaanki (changes requested):

staticcheck.conf
skitt: Done!

dfarrell07 (approved):

tpantelis (approved):

PR #495, opened 2021-03-25
Test K8s versions 1.18, 1.19, and 1.20 in E2E

dfarrell07:
Use the new ability of the shared E2E GHA to configure the Kubernetes
version. Add jobs that cover the latest available patch versions
(limited by kind images) from the three most recent minor versions.

Only run one E2E job with all-default configuration per K8s version, for
some coverage with minimal overhead.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,8 +12,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        globalnet: ['', 'globalnet']
         deploytool: ['operator', 'helm']
+        globalnet: ['', 'globalnet']
+        k8s_version: ['1.17.17']
+        include:
+          # Recentness of K8s versions are limited by kindest/node image releases
+          - k8s_version: 1.18.15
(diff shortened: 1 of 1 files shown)


dfarrell07:
I think this failure is an instability in Lighthouse, not something related to this PR.

https://github.com/submariner-io/lighthouse/actions/workflows/flake_finder.yml

tpantelis (approved):

aswinsuryan (approved):

dfarrell07:
@tpantelis Can we make the upgrade test not-required here, as it's failing?

tpantelis:
Upgrade tests are failing at all repos.

dfarrell07:
> Upgrade tests are failing at all repos.

@tpantelis Right, but they aren't required/blocking merges in most places (and where they are should also be marked not-required).

tpantelis:
> @tpantelis Can we make the upgrade test not-required here, as it's failing?

I changed it.  Also updated the required status for the E2E jobs based on your changes.

PR #520, opened 2021-05-03
Cleanup DEPLOY_ARGS

tpantelis:
We no longer need to set _DEPLOY_ARGS_ conditionally based on helm or operator install - they both can pass `--service_discovery` to the deploy script. Also the lighthouse images flags are no longer valid for helm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,7 @@ override CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override E2E_ARGS += cluster1 cluster2 cluster3
 override UNIT_TEST_ARGS += test/e2e
-
-# Process extra flags from the `using=a,b,c` optional flag
-
-ifneq (,$(filter helm,$(_using)))
-override DEPLOY_ARGS += --service_discovery --deploytool_submariner_args '--set lighthouse.image.repository=localhost:5000/lighthouse-agent,lighthouse.image.tag=local,lighthouseCoredns.image.repository=localhost:5000/lighthouse-coredns,lighthouseCoredns.image.tag=local'
-else
-override DEPLOY_ARGS += --deploytool_broker_args '--service-discovery'
-endif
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

Makefile
```diff
@@ -16,14 +16,7 @@ override CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override E2E_ARGS += cluster1 cluster2 cluster3
 override UNIT_TEST_ARGS += test/e2e
-
-# Process extra flags from the `using=a,b,c` optional flag
-
-ifneq (,$(filter helm,$(_using)))
-override DEPLOY_ARGS += --service_discovery --deploytool_submariner_args '--set lighthouse.image.repository=localhost:5000/lighthouse-agent,lighthouse.image.tag=local,lighthouseCoredns.image.repository=localhost:5000/lighthouse-coredns,lighthouseCoredns.image.tag=local'
```
mkolesnik, line 23: Removing this will cause helm jobs to run with images from quay, thus not testing the latest change

mkolesnik (changes requested):

Makefile
tpantelis: Those specific variables don't exist in the helm charts.  They probably did at some point. We now use 

```
  --set submariner.images.repository="localhost:5000" \
  --set submariner.images.tag="local"
```

in the deploy_helm script which I believe covers LH images as well. Unless I'm  missing something...

mkolesnik: Ah OK since helm charts were changed and now deploy operator which in turn deploys everything, so hopefully it indeed deploys the locally built images and not ones from Quay :)

mkolesnik (approved):

Makefile
tpantelis: yeah I checked the operator code too - the submariner controller passes the `Submariner` object's `repository` and `version` fields to the `ServiceDiscovery` object and the service discovery controller uses those to build the image paths.

PR #522, opened 2021-05-05
Update to latest shipyard/admiral versions

Jaanki:
Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,12 +12,12 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
 	github.com/miekg/dns v1.1.41
-	github.com/onsi/ginkgo v1.16.1
-	github.com/onsi/gomega v1.11.0
+	github.com/onsi/ginkgo v1.16.2
+	github.com/onsi/gomega v1.12.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.10.0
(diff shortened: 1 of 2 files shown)


mangelajo:
Oh, lighthouse got never pinned to 0.9.0?

mangelajo:
Do you need those specific versions for some reason now?

Stable branch is correctly pinned to v0.9.0 
https://github.com/submariner-io/lighthouse/blob/release-0.9/go.mod#L19

Jaanki:
devel should be pinned to latest version right. So the new work is on top of 0.9.0.

mangelajo:
> devel should be pinned to latest version right. So the new work is on top of 0.9.0.

There is no harm in doing so, not sure if there are really changes , may be in shipyard, I'm fine with it ;)

mangelajo (approved):

mkolesnik (changes requested):
> devel should be pinned to latest version right. So the new work is on top of 0.9.0.

devel should be *updated* to devel so please update to the latest devel versions

also this isn't *pinning* it's just version bumps for go dependencies which are internal

mangelajo:
Yeah, mike has a point. 

I mean, we don't really need the change of pinning for any specific reason, but if we do it, let's do it to the current devel :)

mangelajo (changes requested):
see last comments

mkolesnik (changes requested):
You need to run `go mod tidy` there are some uncommitted changes there

tpantelis:
This is superseded by https://github.com/submariner-io/lighthouse/pull/528.

PR #553, opened 2021-06-04
SRV record headless

### LINKED ISSUE #526: Support for SRV records
Support SRV record in Lighthouse for exported services

**Why is this needed**:
SRV records help in discovering services by describing the protocol/s and address of headless services

More details in: https://github.com/submariner-io/enhancements/blob/devel/lighthouse/srv-records.md

aswinsuryan:
Add SRV record support for  headless services

Implements the enhancement proposal here: https://github.com/submariner-io/enhancements/blob/devel/lighthouse/srv-records.md
Resolves issue #526

### DIFF
--- a/pkg/endpointslice/map.go
+++ b/pkg/endpointslice/map.go
@@ -22,8 +22,10 @@ import (
 
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/lighthouse/pkg/constants"
+	"github.com/submariner-io/lighthouse/pkg/serviceimport"
 	discovery "k8s.io/api/discovery/v1beta1"
 	"k8s.io/klog"
+	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
 type endpointInfo struct {
@@ -32,8 +34,8 @@ type endpointInfo struct {
 }
 
 type clusterInfo struct {
-	hostIPs map[string][]string
-	ipList  []string
+	hostRecords map[string][]serviceimport.DNSRecord
+	recordList  []serviceimport.DNSRecord
 }
 
 type Map struct {
@@ -41,7 +43,7 @@ type Map struct {
 	sync.RWMutex
 }
 
-func (m *Map) GetIPs(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]string, bool) {
+func (m *Map) GetIPs(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]serviceimport.DNSRecord, bool) {
 	key := keyFunc(name, namespace)
 
 	clusterInfos := func() map[string]*clusterInfo {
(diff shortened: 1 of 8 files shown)


plugin/lighthouse/handler_test.go
```diff
@@ -658,13 +668,38 @@ func testHeadlessService() {
 				},
 			})
 		})
+		It("should succeed and write an SRV record response", func() {
+			executeTestCase(lh, rec, test.Case{
+				Qname: service1 + "." + namespace1 + ".svc.clusterset.local.",
```
mangelajo, line 673: Are we sure we are implementing what we wanted?

If I look at the k8s docs I see:

# SRV records
SRV Records are created for named ports that are part of normal or Headless Services.
For each named port, the SRV record would have the form `_my-port-name._my-port-protocol.my-svc.my-namespace.svc.cluster-domain.example`. 

* For a regular service, this resolves to the `port number` and the domain name: `my-svc.my-namespace.svc.cluster-domain.example`. 
* For a headless service, this resolves to multiple answers, one for each pod that is backing the service, and contains the` port number` and `the domain name of the pod of the form auto-generated-name.my-svc.my-namespace.svc.cluster-domain.example.`

mangelajo (commented):
Can I get a more elaborate description of what we are trying to implement exactly and how it should behave?

I was looking at https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#srv-records for the non multi-cluster case but it doesn't match what I see here.

plugin/lighthouse/handler_test.go
aswinsuryan: I added one more test case to cover "_my-port-name._my-port-protocol.my-svc.my-namespace.svc.cluster-domain.example".

We support queries of both port name and protocol and without. When port name and protocol is passed only matching entries are returned. Otherwise, all the ports in the service are returned as individual SRV records as described here 

https://github.com/submariner-io/enhancements/blob/devel/lighthouse/srv-records.md
```
### Cluster IP Service
<portname>.<protocol>.<svc-name>.<svc-namespace>.svc.clusterset.local -> will return the SRV record with the port number and domain name with matching port name and protocol from one of the joined clusters where the service is available. The clusters will be selected based on the Round Robin algorithm.

<portname>.<protocol>.<cluster-name>.<svc-name>.<svc-namespace>.svc.clusterset.local -> will return SRV record from the cluster specified, if available with matching portname and protocol.

If <port-name> and <protocol> is not specified, one SRV record will be returned for each port configured for the service in the selected/specified cluster.

### Headless Service
<portname>.<protocol>.<svc-name>.<svc-namespace>.svc.clusterset.local -> will return the SRV records with port number and domain name for each backend pod with matching portname and protocol in every joined cluster where the service is available.

<portname>.<protocol>.<cluster-name>.<svc-name>.<svc-namespace>.svc.clusterset.local -> will return the SRV records with port number and domain name for each backend pod with matching portname and protocol, in the specified cluster.

If <port-name> and <protocol> is not specified, one SRV record will be returned for each port configured for the service in the selected cluster or all clusters based on the query.

```

pkg/endpointslice/map.go
```diff
@@ -32,16 +34,16 @@ type endpointInfo struct {
...
+	recordList  []serviceimport.DNSRecord
 }
 
 type Map struct {
 	epMap map[string]*endpointInfo
 	sync.RWMutex
 }
 
-func (m *Map) GetIPs(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]string, bool) {
+func (m *Map) GetIPs(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]serviceimport.DNSRecord, bool) {
```
mangelajo:
```suggestion
func (m *Map) GetDNSRecords(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]serviceimport.DNSRecord, bool) {
```
sorry for the later nit: can we change the name ? Not really important we can do it later.

plugin/lighthouse/handler_test.go
mangelajo: Thank you, I'm going to link the enhancement proposal to the PR.

vthapar (changes requested):
Looks good, just a minor nitpick

pkg/endpointslice/map_test.go
```diff
@@ -55,7 +57,7 @@ var _ = Describe("EndpointSlice Map", func() {
 		return clusterStatusMap[id]
 	}
 
-	getIPs := func(hostname, cluster, ns, name string) []string {
+	getIPs := func(hostname, cluster, ns, name string) []serviceimport.DNSRecord {
```
tpantelis:
```suggestion
	getRecords := func(hostname, cluster, ns, name string) []serviceimport.DNSRecord {
```

tpantelis (changes requested):

tpantelis (approved):

vthapar (approved):

PR #556, opened 2021-06-10
Add GlobalIngressIP syncer

### LINKED ISSUE #551: Add GlobalIngressIP controller to Lighthouse Agent
Add GlobalIngressIP controller to Lighthouse Agent

**Why is this needed**:
To support (Globalnet v2)[https://github.com/submariner-io/submariner/issues/652] changes in Lighthouse

vthapar:
Initial changes to support Globalnetv2

* Adds a GlobalIngressIP syncer

Fixes #551

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -36,6 +36,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
@@ -154,6 +155,11 @@ func New(spec *AgentSpecification, syncerConf broker.SyncerConfig, kubeClientSet
 		return nil, err
 	}
 
+	if agentController.globalnetEnabled {
+		gvr, _ := schema.ParseResourceArg("globalingressips.v1.submariner.io")
+		agentController.ingressIPClient = syncerConf.LocalClient.Resource(*gvr)
+	}
+
 	return agentController, nil
 }
 
@@ -234,7 +240,7 @@ func (a *Controller) serviceExportToServiceImport(obj runtime.Object, numRequeue
 	klog.V(log.DEBUG).Infof("ServiceExport %s/%s %sd", svcExport.Namespace, svcExport.Name, op)
 
 	if op == syncer.Delete {
-		return a.newServiceImport(svcExport), false
+		return a.newServiceImport(svcExport.Name, svcExport.Namespace), false
 	}
 
 	obj, found, err := a.serviceSyncer.GetResource(svcExport.Name, svcExport.Namespace)
@@ -279,31 +285,38 @@ func (a *Controller) serviceExportToServiceImport(obj runtime.Object, numRequeue
 		return nil, false
 	}
 
-	if a.globalnetEnabled && getGlobalIPFromService(svc) == "" {
-		klog.V(log.DEBUG).Infof("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
+	var ips []string
+	var ports []mcsv1a1.ServicePort
+	if a.globalnetEnabled {
+		ip, reason, msg := a.getGlobalIP(svc)
+		if ip == "" {
+			klog.V(log.DEBUG).Infof("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
+			// Globalnet enabled but service doesn't have globalIp yet, Update the status and requeue
+			a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
+				corev1.ConditionFalse, reason, msg)
 
-		// Globalnet enabled but service doesn't have globalIp yet, Update the status and requeue
-		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
-			corev1.ConditionFalse, "ServiceGlobalIPUnavailable", "Service doesn't have a global IP yet")
+			return nil, true
+		}
 
-		return nil, true
+		ips = []string{ip}
+		ports = a.getPortsForService(svc)
+	} else {
+		ips, ports, err = a.getIPsAndPortsForService(svc, svcType)
+		if err != nil {
+			// Failed to get ips for some reason, requeue
+			a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
+				corev1.ConditionUnknown, "ServiceRetrievalFailed", err.Error())
+
+			return nil, true
+		}
 	}
 
-	serviceImport := a.newServiceImport(svcExport)
+	serviceImport := a.newServiceImport(svcExport.Name, svcExport.Namespace)
 
 	serviceImport.Spec = mcsv1a1.ServiceImportSpec{
 		Ports:                 []mcsv1a1.ServicePort{},
 		Type:                  svcType,
 		SessionAffinityConfig: new(corev1.SessionAffinityConfig),
-	}
-
-	ips, ports, err := a.getIPsAndPortsForService(svc, svcType)
-	if err != nil {
-		// Failed to get ips for some reason, requeue
-		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
-			corev1.ConditionUnknown, "ServiceRetrievalFailed", err.Error())
-
-		return nil, true
 	}
 
 	serviceImport.Status = mcsv1a1.ServiceImportStatus{
@@ -386,7 +399,7 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, numRequeue
 
 	svcExport := obj.(*mcsv1a1.ServiceExport)
 
-	serviceImport := a.newServiceImport(svcExport)
+	serviceImport := a.newServiceImport(svcExport.Name, svcExport.Namespace)
 
 	// Update the status and requeue
 	a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
@@ -467,17 +480,17 @@ func serviceExportConditionEqual(c1, c2 *mcsv1a1.ServiceExportCondition) bool {
 		reflect.DeepEqual(c1.Message, c2.Message)
 }
 
-func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1.ServiceImport {
+func (a *Controller) newServiceImport(name, namespace string) *mcsv1a1.ServiceImport {
 	return &mcsv1a1.ServiceImport{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: a.getObjectNameWithClusterID(svcExport.Name, svcExport.Namespace),
+			Name: a.getObjectNameWithClusterID(name, namespace),
 			Annotations: map[string]string{
-				lhconstants.OriginName:      svcExport.Name,
-				lhconstants.OriginNamespace: svcExport.Namespace,
+				lhconstants.OriginName:      name,
+				lhconstants.OriginNamespace: namespace,
 			},
 			Labels: map[string]string{
-				lhconstants.LabelSourceName:      svcExport.Name,
-				lhconstants.LabelSourceNamespace: svcExport.Namespace,
+				lhconstants.LabelSourceName:      name,
+				lhconstants.LabelSourceNamespace: namespace,
 				lhconstants.LabelSourceCluster:   a.clusterID,
 			},
 		},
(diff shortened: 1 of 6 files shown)


pkg/agent/controller/agent.go
```diff
@@ -154,6 +154,24 @@ func New(spec *AgentSpecification, syncerConf broker.SyncerConfig, kubeClientSet
...
+			RestMapper:       syncerConf.RestMapper,
+			Federator:        agentController.serviceImportSyncer.GetLocalFederator(),
+			ResourceType:     GetGlobalIngressIPObj(),
+			Transform:        agentController.globalIngressIPToServiceImport,
+			OnSuccessfulSync: agentController.onSuccessfulServiceImportSync,
+			Scheme:           syncerConf.Scheme,
+		})
+
+		if err != nil {
+			klog.Fatalf("Error creating globalIngressIP Syncer: %s", err.Error())
```
tpantelis: To consistent with the other error paths.

```suggestion
			return nil, err
```

pkg/agent/controller/agent.go
```diff
@@ -564,3 +599,107 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
 
 	return obj, false
 }
+
+func (a *Controller) globalIngressIPToServiceImport(obj runtime.Object, numRequeues int, op syncer.Operation) (runtime.Object, bool) {
+	unstruct := obj.(*unstructured.Unstructured)
+	klog.Infof("IngressIP: %#v", unstruct)
```
tpantelis: Did you intend to leave this in?

pkg/agent/controller/agent.go
```diff
@@ -564,3 +599,107 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
...
+	gip := parseIngressIP(unstruct)
+	if gip == nil {
+		return nil, false
+	}
+
+	klog.Infof("ParsedIngressIP: %#v", gip)
```
tpantelis: ...and this?

pkg/agent/controller/agent.go
```diff
@@ -564,3 +599,107 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
...
+	obj, found, err = a.serviceSyncer.GetResource(gip.svcName, gip.namespace)
+	if err != nil {
+		// some other error. Log and requeue
+		klog.Errorf("Error retrieving Service for GlobalIngressIP (%s/%s): %v",
+			unstruct.GetNamespace(), unstruct.GetName(), err)
+		return nil, true
+	}
+
+	if !found {
+		// ServiceExport most likely deleted, let Service syncer handle this
```
tpantelis:
```suggestion
		// Service most likely deleted, let Service syncer handle this
```

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,100 @@
...
+		gip.podName, found, err = unstructured.NestedString(obj.Object, "spec", "podRef", "name")
+		if !found || err != nil {
+			klog.Errorf("Expected PodRef in spec %#v", obj.Object)
+			return nil
+		}
+	}
+
+	conditions, found, err := unstructured.NestedSlice(obj.Object, "status", "conditions")
+	if !found || err != nil || len(conditions) < 1 {
+		klog.Errorf("Conditions missing from status %#v", obj.Object)
```
tpantelis: This really isn't an error if not found.

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,100 @@
...
+	gip.condition.message = latestCondition["message"].(string)
+	gip.condition.reason = latestCondition["reason"].(string)
+
+	if latestCondition["type"] == IPAllocated && latestCondition["status"] == "True" {
+		gip.condition.isAllocated = true
+		gip.allocatedIP, found, err = unstructured.NestedString(obj.Object, "status", "allocatedIP")
+		if !found || err != nil {
+			klog.Errorf("Expected AllocatedIP in status of %#v", obj.Object)
+		}
+	}
```
tpantelis, line 81: A condition may not have been set yet so I think we should check if `allocatedIP` is non-empty to determine if allocated. If `allocatedIP` is empty then try to parse the last condition for reason etc. If no condition, then use `reason="ServiceGlobalIPUnavailable" `and `message="Service doesn't have a global IP yet"` as we did before.

pkg/agent/controller/types.go
```diff
@@ -41,6 +41,7 @@ type Controller struct {
 	endpointSliceSyncer     *broker.Syncer
 	serviceSyncer           syncer.Interface
 	serviceImportController *ServiceImportController
+	gipSyncer               syncer.Interface
```
tpantelis:
```suggestion
	ingressIPSyncer               syncer.Interface
```

pkg/agent/controller/agent.go
vthapar: No.

pkg/agent/controller/agent.go
```diff
@@ -279,17 +285,20 @@ func (a *Controller) serviceExportToServiceImport(obj runtime.Object, numRequeue
...
-		klog.V(log.DEBUG).Infof("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
-
-		// Globalnet enabled but service doesn't have globalIp yet, Update the status and requeue
-		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid,
-			corev1.ConditionFalse, "ServiceGlobalIPUnavailable", "Service doesn't have a global IP yet")
+	if a.globalnetEnabled {
+		ip, reason, msg := a.getGlobalIP(svc)
+		if ip == "" {
+			klog.V(log.DEBUG).Infof("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
+			klog.Errorf("Service to be exported (%s/%s) doesn't have a global IP yet", svcExport.Namespace, svcExport.Name)
```
tpantelis: We log to debug but I don't think this warrants an error log

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,113 @@
...
+		gip.condition.isAllocated = false
+		gip.condition.message = latestCondition["message"].(string)
+		gip.condition.reason = latestCondition["reason"].(string)
+		if latestCondition["type"] == IPAllocated && latestCondition["status"] == "True" {
+			gip.condition.isAllocated = true
+			if gip.allocatedIP == "" {
+				klog.Errorf("Expected AllocatedIP in status of %#v", obj.Object)
+			}
+		}
+	}
```
tpantelis: As I mentioned before I think we should base the allocated status on `allocatedIP` and only consult the conditions to get the reason if not set. The conditions are really just to report info to the end user.

```suggestion
        conditions, _, _ := unstructured.NestedSlice(obj.Object, "status", "conditions")
        if gip.allocatedIP == "" {
            if len(conditions) > 0 {
                latestCondition := conditions[len(conditions)-1].(map[string]interface{})
                gip.unallocatedMessage = latestCondition["message"].(string)
                gip.unallocatedReason = latestCondition["reason"].(string)
            } else {
                gip.unallocatedMessage = defaultMsgIPUnavailable
                gip.unallocatedReason = defaultReasonIPUnavailable
            }
        }
```

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,113 @@
...
+	namespace   string
+	target      string
+	svcName     string
+	podName     string
+	allocatedIP string
+	condition   struct {
+		isAllocated bool
+		reason      string
+		message     string
+	}
```
tpantelis: I don't think we need the inner structure nor the `isAllocated` flag - non-empty `allocatedIP` indicates the allocated status. 

```suggestion
	allocatedIP string
	unallocatedReason      string
	unallocatedMessage     string
```

pkg/agent/controller/agent_test.go
```diff
@@ -370,6 +371,8 @@ var _ = Describe("Globalnet enabled", func() {
 			t.awaitServiceExportStatus(0, newServiceExportCondition(mcsv1a1.ServiceExportValid,
 				corev1.ConditionFalse, "ServiceGlobalIPUnavailable"))
 
+			// t.createGlobalIngressIP()
+			// TODO: Delete SetAnnotations for globalnetv2
```
tpantelis, line 375: We should be able to do this?

pkg/agent/controller/agent.go
```diff
@@ -564,3 +579,32 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
...
+		ip = a.getGlobalIPFromService(service)
+		if ip != "" {
+			return ip, "", ""
+		}
+
+		ingressIP := a.getIngressIP(service.Name, service.Namespace)
+		if ingressIP == nil || ingressIP.allocatedIP == "" {
+			reason, msg = getReasonAndMsgFromIngressIP(ingressIP)
+			return "", reason, msg
+		}
```
tpantelis:
```suggestion
		ingressIP, notFound, err := a.getIngressIP(service.Name, service.Namespace)
		if err != nil {
		    return "", "GlobalIngressIPRetrievalFailed", err.Error()
		}
		
		if notFound {
		    return "",  defaultReasonIPUnavailable, defaultMsgIPUnavailable
		}
```

pkg/agent/controller/agent.go
```diff
@@ -564,3 +579,32 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
...
+
+func (a *Controller) getIngressIP(name, namespace string) *IngressIP {
+	obj, err := a.ingressIPClient.Namespace(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+
+	if err != nil {
+		return nil
+	}
+
+	return parseIngressIP(obj)
+}
```
tpantelis: I don't think we should ignore the error from `Get`. I think we should return a bool indicating found or not and an error.

```suggestion
func (a *Controller) getIngressIP(name, namespace string) (*IngressIP, bool, error) {
    obj, err := a.ingressIPClient.Namespace(namespace).Get(context.TODO(), name, metav1.GetOptions{})
    if apierrors.IsNotFound(err) {
        return nil, false, nil 
    }
        
    if err != nil {
        return nil, false, err
    }
     
    return parseIngressIP(obj), true, nil
}
```

pkg/agent/controller/agent.go
```diff
@@ -486,18 +495,10 @@ func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1
...
-			Name:     port.Name,
-			Protocol: port.Protocol,
-			Port:     port.Port,
-		})
-	}
+	mcsPorts := a.getPortsForService(service)
 
 	if siType == mcsv1a1.ClusterSetIP {
-		mcsIP := getGlobalIPFromService(service)
+		mcsIP, _, _ := a.getGlobalIP(service)
```
tpantelis: Since we already retrieved and parsed the global IP previously, we can pass it into this function to avoid the overhead of repeating it. Also it's possible `getGlobalIP` could fail here where it didn't previously.

tpantelis (changes requested):

pkg/agent/controller/agent.go
```diff
@@ -564,3 +579,32 @@ func (a *Controller) filterLocalEndpointSlices(obj runtime.Object, numRequeues i
 
 	return obj, false
 }
+
+func (a *Controller) getGlobalIP(service *corev1.Service) (ip, reason, msg string) {
+	if a.globalnetEnabled {
+		ip = a.getGlobalIPFromService(service)
```
sridhargaddam, line 585: Cool. So we are maintaining backward compatibility. But we have to remember that if a cluster is upgraded from Globalnet V1 to Globalnet V2 and the existing annotations on the Services are not removed, we will accidentally advertise the old GlobalIPs (I.e., annotation) as part of DNS. We just have to document this in the upgrade doc for Globalnet V2.

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog"
+)
+
+const (
+	ClusterIPService           = "ClusterIPService"
+	headlessServicePod         = "HeadlessServicePod"
+	IPAllocated                = "Allocated"
+	defaultReasonIPUnavailable = "ServiceGlobalIPUnavailable"
+	defaultMsgIPUnavailable    = "Service doesn't have a global IP yet"
```
sridhargaddam: Some of these constants are global/exported and some not. Is this intentional?

pkg/agent/controller/globalingressip.go
sridhargaddam: Agree with Tom regarding the usage of conditions.

sridhargaddam (commented):
Thanks for the PR @vthapar.

pkg/agent/controller/globalingressip.go
vthapar: I was thinking of a case where it might be a stale ip.

pkg/agent/controller/agent.go
vthapar: Missed the cleanup

pkg/agent/controller/agent_test.go
vthapar: Will do once we switch. Since current code is still using annotations, want to make sure we're testing it at least once.

pkg/agent/controller/agent.go
vthapar: There will be delay in creation of svcex and globalingress. We'll be requeued 3-4 times and this can lead to unnecessary padding of status to ServiceExport. Besides, currently we don't have v2 enabled. This will result in SvcEx showing status `GlobalIngressIPRetrievalFailed` when we're not even using v2.

pkg/agent/controller/agent.go
vthapar: Personally, I'd prefer to remove this altogether when we switch. Only reason I left it in for now is to not break things till we switch to v2. Since we don't want to support v1 at all, it is better to avoid problems that can crop up from trying to half heartedly support v1 + v2.

pkg/agent/controller/globalingressip.go
vthapar: Global ones are what I use in UT.

pkg/agent/controller/agent.go
```diff
@@ -486,18 +499,10 @@ func (a *Controller) newServiceImport(svcExport *mcsv1a1.ServiceExport) *mcsv1a1
...
-			Name:     port.Name,
-			Protocol: port.Protocol,
-			Port:     port.Port,
-		})
-	}
+	mcsPorts := a.getPortsForService(service)
 
 	if siType == mcsv1a1.ClusterSetIP {
-		mcsIP := getGlobalIPFromService(service)
+		mcsIP, _, _ := a.getGlobalIP(service)
```
tpantelis: We no longer enter this path with GN enabled so we don't need to call `getGlobalIP`.

pkg/agent/controller/globalingressip.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog"
+)
+
+const (
+	ClusterIPService           = "ClusterIPService"
+	headlessServicePod         = "HeadlessServicePod"
+	IPAllocated                = "Allocated"
```
tpantelis: We no longer use this in the production code so we don 't need this constant.

tpantelis (changes requested):
Looks good - just a couple more comments.

tpantelis (approved):

sridhargaddam (approved):

PR #588, opened 2021-07-13
Move load balance logic out of the repo scope

danibachar:

Moved Load Balancing logic out of the repo scope.
First stage, towards optimized ServiceImport Load Balancing as mentioned in the [EP](https://github.com/submariner-io/enhancements/pull/2)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/caddyserver/caddy v1.0.5
 	github.com/coredns/caddy v1.1.1
 	github.com/coredns/coredns v1.8.3
+	github.com/danibachar/gobalancing v0.0.0-20210713235731-3db9fc142709
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/miekg/dns v1.1.43
 	github.com/onsi/ginkgo v1.16.4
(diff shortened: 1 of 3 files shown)


danibachar:
I'm note sure how do we feel about adding external libraries - I can move it from my GitHub to this project's if needed.
But it will be best to have the LB logic in a separate well tested library

mangelajo:
Hey @danibachar,

Thank you to keep up with this, I know we didn't have much time to follow up with you, let me share my thoughts on this.

I strongly believe that if this is a library only to be consumed by lighthouse, should live in a lighthouse package. Nothing on our side precludes you from extracting this library on it's own publishing long-term if you believe it will be useful to other project.

It's much easier for the project to maintain it as an internal detail than as an external library. If we need to make changes/updates on the future we would need to go through the external repository, just to update a library that we only consume. 

Normally, when adding a new external library we look at multiple details:
[ ] Long term sustainability: does it have enough forks/likes/community/activity/contributors.
[x] Compatibility of the license. (+/- MIT vs Apache)
[x] Is it well tested.
[ ] Does it follow good semver releasing we can target from go.mod? (not yet I guess)

The 1st point is important (For example, if you had other responsibilities -new job-, you can't handled PRs, etc), with the existing license (MIT) we couldn't even copy the code into your codebase if necessary.


Don't get me wrong @danibachar your contributions are highly appreciated, and I believe this is increasingly important for the long term of the project (in the short term we had too many other items). 

Do you think it's reasonable to contribute this to submariner/lighthouse directly under it's own package?

@skitt @dfarrell07 @tpantelis @mkolesnik @sridhargaddam @vthapar  thoughts?

tpantelis:
> Do you think it's reasonable to contribute this to submariner/lighthouse directly under it's own package?
> 
> @skitt @dfarrell07 @tpantelis @mkolesnik @sridhargaddam @vthapar thoughts?

I agree. I had a discussion with @danibachar privately on Slack and the consensus was to keep it in LH under a new `loadbalance` module.

PR #607, opened 2021-08-20
Use SHAs for GHA versions

dfarrell07:
Per GitHub's security guidelines, GHAs should be pinned using full
length commit SHAs instead of tags.

The SHAs are of the commits currently resolved by the versions.

Even "trusted" GHAs from GitHub developers are pinned because it's
possible their repo right could be compromised and a malicious GHA
published. These core repos are not frequently substantially updated.

Submariner-internal GHAs are left pinned at devel because we want
automatic updates from Shipyard's shared tooling.

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
       - name: Delete current CODEOWNERS file
         run: rm CODEOWNERS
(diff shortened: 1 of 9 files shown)


dfarrell07:
YAML linting is fixed by: https://github.com/submariner-io/lighthouse/pull/606

skitt:
Huh, the troublesome YAML file here is `upgrade-e2e.yaml` but #606 didn’t fix that ;-).

dfarrell07:
> Huh, the troublesome YAML file here is `upgrade-e2e.yaml` but #606 didn’t fix that ;-).

Ah right, I actually introduce this error here, the linter is catching a copy+paste error caused by this GHA not having a `name` field. Fixing it now...

skitt (approved):

aswinsuryan (approved):

PR #615, opened 2021-09-21
DNS Query Counter

danibachar:

Adding custom metrics for DNS queries served with the lighthouse plugin.
As the [basic CoreDNS Prometheus metrics](https://coredns.io/plugins/metrics/) don’t suffice  - note that even though CoreDNS have the `coredns_dns_requests_total ` and `coredns_dns_responses_total ` they don't hold all the information we need.

In addition, I added the `ClusterName` property to the DNSRecords returned from the ServiceImport and Service controllers (the EndpointSlice one already had that info added to the record) - I thought of having it in a separate PR but it seems an overkill - if separation is needed please let me know.

### DIFF
--- a/plugin/lighthouse/metrics.go
+++ b/plugin/lighthouse/metrics.go
@@ -1,0 +1,67 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package lighthouse
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/klog"
+)
+
+const (
+	srcClusterKey      = "source_cluster"
+	dstClusterKey      = "destination_cluster"
+	dstSvcNameKey      = "destination_service_name"
+	dstSvcIPKey        = "destination_service_ip"
+	dstSvcNamespaceKey = "destination_service_namespace"
+
+	ServiceDiscoveryQueryCounterName = "submariner_service_discovery_query_counter"
+)
+
+var dnsQueryCounter *prometheus.GaugeVec
+
+func initLighthouseMetrics() {
+	klog.Infof("Initializing dns query counter")
+
+	dnsQueryCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: ServiceDiscoveryQueryCounterName,
+			Help: "Count the numebr of dns queries",
+		},
+		[]string{srcClusterKey, dstClusterKey, dstSvcNameKey, dstSvcNamespaceKey, dstSvcIPKey},
+	)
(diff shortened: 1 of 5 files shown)


nyechiel:
@vthapar @aswinsuryan @maayanf24 can you review this please?

plugin/lighthouse/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err := prometheus.Register(dnsQueryCounter); err != nil {
+		klog.Errorf("Failed to register: %s with prometheus due to: %v", ServiceDiscoveryQueryCounterName, err)
+	}
+}
+
+func incDNSQueryCounter(srcCluster, dstCluster, dstSvcName, dstSvcNamespace, dstSvcIP string) {
+	if dnsQueryCounter == nil {
+		initLighthouseMetrics()
+	}
```
tpantelis: We shouldn't assume this won't be called concurrently and/or from multiple threads so either protect access to `dnsQueryCounter` or, better yet, initialize it statically (ie in an `init` function).

tpantelis (changes requested):

plugin/lighthouse/metrics.go
danibachar: Fixed! Thanks for that, I'm still getting familiar with Go - after reading a bit on the init function it makes total sense to do so and not using lazy init...

plugin/lighthouse/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const (
+	srcClusterKey      = "source_cluster"
+	dstClusterKey      = "destination_cluster"
+	dstSvcNameKey      = "destination_service_name"
+	dstSvcIPKey        = "destination_service_ip"
+	dstSvcNamespaceKey = "destination_service_namespace"
+
+	ServiceDiscoveryQueryCounterName = "submariner_service_discovery_query_counter"
```
maayanf24:
```suggestion
	ServiceDiscoveryQueryCounterName = "submariner_service_discovery_query"
```

plugin/lighthouse/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+
+var dnsQueryCounter *prometheus.GaugeVec
+
+func init() {
+	klog.Infof("Initializing dns query counter")
+
+	dnsQueryCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: ServiceDiscoveryQueryCounterName,
+			Help: "Count the numebr of dns queries",
```
maayanf24:
```suggestion
			Help: "Count of DNS queries",
```

plugin/lighthouse/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		[]string{srcClusterKey, dstClusterKey, dstSvcNameKey, dstSvcNamespaceKey, dstSvcIPKey},
+	)
+
+	if err := prometheus.Register(dnsQueryCounter); err != nil {
```
maayanf24: Please use `prometheus.MustRegister` instead of `prometheus.Register`

maayanf24 (changes requested):

plugin/lighthouse/metrics.go
danibachar: fixed

plugin/lighthouse/metrics.go
danibachar: fixed

plugin/lighthouse/metrics.go
danibachar: fixed

maayanf24 (approved):

tpantelis (approved):

PR #648, opened 2021-12-01
ServiceImportsWeightMap CRD and Controller

danibachar:

After a great discussion with @vthapar, we thought about the best way to update the weights for our Weighted-Round-Robin load-balancing will be to introduce a new CRD.

Most of the code here is auto-generated. Focus on:
1) The new types introduce at `pkg/apis/lighthouse.submariner.io/v1alpha1/types.go`
I basically allowed a map for source cluster -> (namespace, service, target cluster)
2) Updates to the serviceimport map where it is injected with a function to fetch the weights from - `pkg/serviceimport/map.go`
3) The new controller introduced - `pkg/serviceimport/weights/controller.go`


pkg/serviceimport/map.go
```diff
@@ -59,8 +58,9 @@ func (si *serviceInfo) resetLoadBalancing() {
 }
 
 type Map struct {
-	svcMap map[string]*serviceInfo
-	mutex  sync.RWMutex
+	svcMap             map[string]*serviceInfo
+	weightFetchingFunc func(service, namesapce, inCluster string) int64
```
tpantelis: I suggest to define a type:

`type WeightFunc(service, namespace, cluster string) int64`

```suggestion
	getWeightFor WeightFunc
```

pkg/serviceimport/weights/suite_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"k8s.io/klog"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+func TestServiceImportWeights(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "ServiceImportWeights	 Suite")
```
tpantelis:
```suggestion
	RunSpecs(t, "ServiceImportWeights Suite")
```

pkg/serviceimport/map.go
```diff
@@ -120,9 +120,10 @@ func (m *Map) GetIP(namespace, name, cluster, localCluster string, checkCluster
 	return nil, true, false
 }
 
-func NewMap() *Map {
+func NewMap(weightFetchingFunc func(service, namesapce, inCluster string) int64) *Map {
```
tpantelis:
```suggestion
func NewMap(getWeightFor WeightFunc) *Map {
```

pkg/serviceimport/map_test.go
```diff
@@ -44,7 +48,7 @@ var _ = Describe("ServiceImport Map", func() {
 
 	BeforeEach(func() {
 		clusterStatusMap = map[string]bool{clusterID1: true, clusterID2: true, clusterID3: true}
-		serviceImportMap = serviceimport.NewMap()
+		serviceImportMap = serviceimport.NewMap(GetWeightFor)
```
tpantelis: You can just inline for simplicity.

pkg/serviceimport/weights/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+func (c *Controller) Stop() {
+	close(c.stopCh)
+
+	klog.Infof("ServiceImportsWeights Controller stopped")
+}
+
+func (c *Controller) GetWeightFor(service, namesapce, inCluster string) int64 {
```
tpantelis:
```suggestion
func (c *Controller) GetWeightFor(service, namespace, inCluster string) int64 {
```

pkg/serviceimport/weights/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	obj, exists, err := c.weightsStore.GetByKey(lhconstants.SubmarinerWeightMapKey)
+	if err != nil {
+		klog.V(log.DEBUG).Infof("Error trying to get weight map for local cluster %q", c.localClusterID)
```
tpantelis:
```suggestion
		klog.Errorf("Error trying to get weight map for local cluster %q", c.localClusterID)
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func (n *weightMapTestDriver) updateWeightsMap(name, service, namesapce, inCluster string, weight int64) {
```
tpantelis:
```suggestion
func (n *weightMapTestDriver) updateWeightsMap(name, service, namespace, inCluster string, weight int64) {
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	BeforeEach(func() {
+		t.kubeClient = fakeClientSet.NewSimpleClientset()
+		t.controller = weights.NewController(cluster1)
+		t.controller.NewClientset = func(c *rest.Config) (lighthouseClientset.Interface, error) {
+			return t.kubeClient, nil
+		}
+
+		Expect(t.controller.Start(&rest.Config{})).To(Succeed())
+		Expect(v1.AddToScheme(scheme.Scheme)).To(Succeed())
```
tpantelis: Is this needed?

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+	})
+
+	AfterEach(func() {
+		t.controller.Stop()
+	})
+
+	return t
+}
+
+var _ = Describe("ServiceImportWeights controller", func() {
```
tpantelis: Since the test cases are the prominent feature in the file, I prefer to put them at the top ahead of the supporting test driver struct.

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Eventually(func() int64 {
+			return t.controller.GetWeightFor(service, namesapce, inCluster)
+		}, 5).Should(Equal(weight))
+	}
+
+	validateWeightNotAvailable := func(service, namesapce, inCluster string) {
+		Eventually(func() int64 {
+			return t.controller.GetWeightFor(service, namesapce, inCluster)
+		}, 5).Should(Equal(int64(1)))
+	}
```
tpantelis: Make these methods of `weightMapTestDriver`.

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	When("weigth does not exists for query", func() {
```
tpantelis: _weigth_ (sic ) - same in several places below as well.

```suggestion
	When("no weight is defined", func() {
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+const (
+	weightMapName = lhconstants.SubmarinerWeightMapKey
+
+	ns1      = "ns1"
+	cluster1 = "cluster1" // local
+	cluster2 = "cluster2" // remote
+	cluster3 = "cluster3" // remote
+	scv1     = "scv1"
```
tpantelis:
```suggestion
	namespace      = "test"
	localCluster = "local" // local
	remoteCluster1 = "remote1" // remote
	remoteCluster2 = "remote2" // remote
	serviceName     = "test-service"
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	When("weigth does not exists for query", func() {
+		It("should return default weight of 1", func() {
+			validateWeightNotAvailable(scv1, "non-existing-ns", cluster2)
+			validateWeightNotAvailable("non-existing-svc", ns1, cluster2)
+			validateWeightNotAvailable(scv1, ns1, "non-existing-cluster")
+		})
+	})
+
+	When("weight map is added", func() {
```
tpantelis:
```suggestion
	When("a ServiceImportWeightMap is added", func() {
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+		BeforeEach(func() {
+			t.createWeightsMap(weightMapName)
+		})
+		It("should return the relevant weight for the query", func() {
+			validateWeightAvailable(scv1, ns1, cluster2, 8)
+		})
+		When("weigth map is removed", func() {
```
tpantelis:
```suggestion
		Context("and then removed", func() {
```

pkg/serviceimport/weights/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+			It("should return default weight of 1", func() {
+				t.deleteWeightsMap(weightMapName)
+				validateWeightNotAvailable(scv1, ns1, cluster2)
+			})
+		})
+		When("weigth map is updated", func() {
```
tpantelis:
```suggestion
		When("and then updated", func() {
```

plugin/lighthouse/handler_test.go
```diff
@@ -1016,7 +1020,7 @@ func (t *handlerTestDriver) executeTestCase(rec *dnstest.Recorder, tc test.Case)
 }
 
 func setupServiceImportMap() *serviceimport.Map {
-	siMap := serviceimport.NewMap()
+	siMap := serviceimport.NewMap(GetWeightFor)
```
tpantelis: You can just inline it.

pkg/apis/lighthouse.submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ServiceImportWeights is the Schema for the serviceimportweights API.
+type ServiceImportWeightMap struct {
```
tpantelis, line 25:
```suggestion
type ServiceImportWeights struct {
```

.licenserc.json
```diff
@@ -1,30 +1,32 @@
...
+		"ignore": [
+			"pkg/client/",
+			"pkg/apis/lighthouse.submariner.io/v1alpha1/zz_generated.deepcopy.go",
+			"vendor/"
+		]
+	},
+	{
+		"LICENSE": "Apache License"
+	}
+]
```
tpantelis, line 32: Was this change intentional?

package/submariner.io.lighthouse_serviceimportweightmap-crd.yaml
```diff
@@ -0,0 +1,79 @@
...
+  creationTimestamp: null
+  name: serviceimportweightmaps.lighthouse.submariner.io
+spec:
+  group: lighthouse.submariner.io
+  names:
+    kind: ServiceImportWeightMap
+    listKind: ServiceImportWeightMapList
+    plural: serviceimportweightmaps
+    singular: serviceimportweightmap
+  scope: Namespaced
```
tpantelis, line 16: It doesn't seem like this needs to be namespaced. The code watches all namespaces anyway.

pkg/apis/lighthouse.submariner.io/v1alpha1/types.go
```diff
@@ -0,0 +1,66 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ServiceImportWeightsSpec defines the desired state of ServiceImportWeights.
+type ServiceImportWeightMapSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Weights map of [src_cluster -> [namespace -> [svc_name -> [target_cluster -> weight]]]]
+	SourceClusterWeightMap map[string]*ClusterWeightMap `json:"source_cluster_weight_map,omitempty"`
+}
```
tpantelis, line 42: The nesting levels seem overly complex. I think we can simplify by defining a single struct:

```
type WeightSpec struct {
    SourceCluster `json:"sourceCluster"`
    Service `json:"service"`
    Namespace `json:"namespace"`
    TargetCluster `json:"targetCluster"`
    Weight int64 `json:"weight"`
}

Weights []WeightSpec `json:"weights,omitempty"`
```

Is this intended to be distributed or defined per cluster? If the latter then we can drop the `SourceCluster`.

pkg/serviceimport/weights/controller.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (c *Controller) Stop() {
+	close(c.stopCh)
+
+	klog.Infof("ServiceImportsWeights Controller stopped")
+}
+
+func (c *Controller) GetWeightFor(service, namesapce, inCluster string) int64 {
+	obj, exists, err := c.weightsStore.GetByKey(lhconstants.SubmarinerWeightMapKey)
```
tpantelis, line 104: So I assume the CRD is intended to be a singleton named "submariner-weights".  This lookup assumes no namespace. However if we don't make the CRD namespaced then this will work fine.

tpantelis (changes requested):

.licenserc.json
danibachar: nope, auto formating with ctrl-s, thought I should leave it as vs code formated it but I'll revert

pkg/apis/lighthouse.submariner.io/v1alpha1/types.go
danibachar: You are right about the sourceCluster, left overs of my optimization code - where I build the full map - but no need to pass all the information to all the clusters just their local one.
About the new struct - mapping prevents searching - don't you think it worth using a map?
I thought of maybe having the key to be like a path/dns  `cluster(/.)scv(/.).namespace` = weight
Which will make the nesting simpler - wdyt? @tpantelis

pkg/serviceimport/weights/controller.go
danibachar: Got it, will remove the namespace scope for the CRD - thanks!

pkg/apis/lighthouse.submariner.io/v1alpha1/types.go
tpantelis: > About the new struct - mapping prevents searching - don't you think it worth using a map?

"_mapping prevents searching_" - I'm not sure what you mean here. 

> I thought of maybe having the key to be like a path/dns cluster(/.)scv(/.).namespace = weight
> Which will make the nesting simpler - wdyt?

I'm not clear on why it needs to be specifically keyed and what's the use case for searching in that manner. To me, this CRD is just a vehicle for a user to specify weights per service and cluster.

Another approach is to have the CRD define a single weight policy instead of a singleton list, ie

```
type ServiceWeightPolicy struct {
    metav1.TypeMeta   `json:",inline"`
    metav1.ObjectMeta `json:"metadata,omitempty"`

    Spec ServiceWeightPolicySpec `json:"spec,omitempty"`
}

type ServiceWeightPolicySpec struct {
    Service `json:"service"`
    Namespace `json:"namespace"`
    TargetCluster `json:"targetCluster"`
    Weight int64 `json:"weight"`
}
```

In fact we could even drop `Namespace` if we require the `ServiceWeightPolicy` be created in the same namespace as the service.

I think this approach is more standard in K8s.

danibachar: Hmm, I understand what you are saying.
When I referred to search I meant to find the relevant weight for the service when the import happens. If we hold an array we will have to search the relevant record.
I like what you have suggested here about simplifying the structure and scope it in the same namespace as the service!
The question is if this will allow as simple querying for the object at the time we need - similarly to here - https://github.com/submariner-io/lighthouse/blob/cf009d9861e6078570300a0dc26477015931b870/pkg/service/controller.go#L97

Would I be able to build a similar key - namespace/service/targetCluster (or similar combination)  for a fast extraction of the object and then the weight?

tpantelis: OK so you're referring to the controller implementation wrt searching. Remember this is the user-facing API and should be designed as such and not based on the backend controller implementation. You can build your own `map[string]int64` from each `ServiceWeightPolicy` keyed by `Service + "/" + Namespace + "/" + TargetCluster` similarly as is done in _lighthouse/pkg/serviceimport/controller.go_.

danibachar: Yea you are right, the thing is that calculating the weights is done all at once, and I thought of distributing them once per cluster in one structure to avoid sending a lot of smaller pieces of information for each update (in systems with a lot of services it could result in a lot of API calls).
Do you think it is over-optimizing for the moment? And once we face performance issues we aggregate the calls? Maybe there are APIs of Kubernetes I'm not aware of that could solve this foreseen issue?

pkg/serviceimport/weights/controller_test.go
danibachar: Great catch - leftovers of my scheme registration problem - thanks

pkg/apis/lighthouse.submariner.io/v1alpha1/types.go
tpantelis: > Yea you are right, the thing is that calculating the weights is done all at once

I'm not sure what you mean by this...

> , and I thought of distributing them

I'm not sure distributing them would make sense...  but, if so, I wouldn't imagine there would be enough of them to cause a performance hit.

PR #653, opened 2021-12-08
Fix fetching weight from service import annotations by local cluster ID

danibachar:
Signed-off-by: Daniel Bachar <danibachar89@gmail.com>

After debating a bit in the group I have decided to remove the need for a new CRD and support weights distribution externally to Submariner using just the ServiceImport annotations

The explanation is described in the EP - submariner-io/enhancements#2
Link to the explanation - https://github.com/danibachar/enhancements/blob/feature/lighthouse/lighthouse/service-selection/multi-cluster-service-selection.md#serviceweightpolicy-distribution

TL;DR - I based the idea on the Istio LocalityLoadBalancerSetting https://istio.io/latest/docs/reference/config/networking/destination-rule/#LocalityLoadBalancerSetting

### DIFF
--- a/pkg/serviceimport/map.go
+++ b/pkg/serviceimport/map.go
@@ -59,8 +59,9 @@ func (si *serviceInfo) resetLoadBalancing() {
 }
 
 type Map struct {
-	svcMap map[string]*serviceInfo
-	mutex  sync.RWMutex
+	svcMap         map[string]*serviceInfo
+	localClusterID string
+	mutex          sync.RWMutex
 }
 
 func (m *Map) selectIP(si *serviceInfo, name, namespace string, checkCluster func(string) bool,
@@ -120,9 +121,10 @@ func (m *Map) GetIP(namespace, name, cluster, localCluster string, checkCluster 
 	return nil, true, false
 }
 
-func NewMap() *Map {
+func NewMap(localClusterID string) *Map {
 	return &Map{
-		svcMap: make(map[string]*serviceInfo),
+		svcMap:         make(map[string]*serviceInfo),
+		localClusterID: localClusterID,
 	}
 }
 
@@ -157,7 +159,7 @@ func (m *Map) Put(serviceImport *mcsv1a1.ServiceImport) {
 			remoteService.records[clusterName] = &clusterInfo{
 				name:   clusterName,
 				record: record,
-				weight: getServiceWeightFrom(serviceImport, clusterName),
+				weight: getServiceWeightFrom(serviceImport, m.localClusterID),
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/serviceimport/map.go
```diff
@@ -195,14 +197,14 @@ func (m *Map) Remove(serviceImport *mcsv1a1.ServiceImport) {
 }
 
 func getServiceWeightFrom(si *mcsv1a1.ServiceImport, forClusterName string) int64 {
-	weightKey := "load-balancer.submariner.io/weight/" + forClusterName
+	weightKey := lhconstants.LoadBalancerWeightAnnotationPrefix + "/" + forClusterName
```
tpantelis, line 200: The name part of the annotation key cannot have a [slash](load-balancer). So we can use '-' or '.'. Also the name is limited to 63 chars which would be an issue for a cluster name longer than `62 - len(lhconstants.LoadBalancerWeightAnnotationPrefix)`.

```suggestion
	weightKey := lhconstants.LoadBalancerWeightAnnotationPrefix + "." + forClusterName
```

tpantelis (changes requested):

pkg/serviceimport/map.go
danibachar: I'm not sure I follow, maybe I missed something.
But according to https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
I can have an optional DNS like prefix for the key (up to 255 characters) hence the `lhconstants.LoadBalancerWeightAnnotationPrefix = "lighthouse-load-balancer.submariner.io/weight"` Maybe I should change it to `lhconstants.LoadBalancerWeightAnnotationPrefix = "lighthouse-load-balancer-weight.submariner.io"` 
And then construct `prefix + "/" + clusterName`?

Then the annotations will be of
```
lighthouse-load-balancer-weight.submariner.io/cluster1: 10,
lighthouse-load-balancer-weight.submariner.io/cluster2: 80,
lighthouse-load-balancer-weight.submariner.io/cluster3: 10,
```

When you refer to the 63 characters limit - it is not clear in the documentations if it includes the prefix...

tpantelis: I just realized you are correctly using the "/" separator so ignore that part of my comment.

> When you refer to the 63 characters limit - it is not clear in the documentations if it includes the prefix...

It does not include the prefix. It states that the "_name segment is required and must be 63 characters or less_" and "_prefix must...not longer than 253 characters in total_". 

You could use the form `lighthouse-load-balancer-weight.submariner.io/cluster1`.  That would give us more chars for the name part but it would still be an issue if a cluster name exceeded 63. But that's probably not likely in a production environment.

vthapar: I'd recommened even shortening it further to `lighthouse-lb-weight.submariner.io/clusterX`. Shorter the prefix we use, larger the clustername we can support before someone runs into error. Don't underestimate possibility of long clusternames with something like `country-state-city-datacenter-department-production-name`

vthapar (changes requested):

pkg/serviceimport/map.go
tpantelis: FWIU from the k8s docs Iinked, the max prefix length (253) is separate from the max name length (63).

aswinsuryan (approved):

tpantelis (approved):

vthapar (approved):

PR #658, opened 2021-12-17
Unpin Kubernetes dependencies

skitt:
This doesn't seem to break anything, and helps with go install (see
<https://github.com/golang/go/issues/44840>).


skitt:
The issues with the various unpin K8s patches are that the bump to 0.20 ends up breaking Admiral, because of changes in the fake client which result in errors handling List objects.

dfarrell07 (approved):

aswinsuryan (approved):

dfarrell07:
When we get this working, we should also consider enabling the `gomoddirectives` linter to keep replace statements out.

https://github.com/submariner-io/lighthouse/blob/devel/.golangci.yml#L61

nyechiel:
@skitt I assume we want this fixed for 0.12?

skitt:
> @skitt I assume we want this fixed for 0.12?

I would have liked it fixed for 0.12, but it’s too complex! We’ll have to punt it to 0.13.

skitt:
The build is now broken because our Kubernetes dependencies require Go 1.17.

tpantelis:
Done with https://github.com/submariner-io/lighthouse/pull/744.

PR #681, opened 2022-02-08
Use MCS-standard labels for EndpointSlices

### LINKED ISSUE #680: imported EndpointSlices missing `multicluster.kubernetes.io/service-name` label
**What happened**:

Created ServiceExport. Endpointslice is created like:
```yaml
addressType: IPv4
apiVersion: discovery.k8s.io/v1beta1
endpoints:
- addresses:
  - 10.1.0.6
  conditions:
    ready: true
  hostname: echo-cb96f8d94-mbdg9
  topology:
    kubernetes.io/hostname: cluster1-worker2
kind: EndpointSlice
metadata:
  creationTimestamp: "2022-02-08T16:20:43Z"
  generation: 1
  labels:
    endpointslice.kubernetes.io/managed-by: lighthouse-agent.submariner.io
    lighthouse.submariner.io/sourceCluster: cluster1
    lighthouse.submariner.io/sourceName: echo
    lighthouse.submariner.io/sourceNamespace: default
    submariner-io/clusterID: cluster1
    submariner-io/originatingNamespace: default
  name: echo-cluster1
  namespace: default
  resourceVersion: "3547"
  uid: 01e7eba2-fec3-45a4-9809-621965a7faf1
ports:
- name: grpc
  port: 7070
  protocol: TCP
- name: https
  port: 443
  protocol: TCP
- name: http
  port: 80
  protocol: TCP
- name: tcp-sf
  port: 9091
  protocol: TCP
- name: tcp
  port: 9090
  protocol: TCP
```

 `multicluster.kubernetes.io/service-name` label is added, per https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services

skitt:
The MCS specification indicates that EndpointSlices are labeled with
"multicluster.kubernetes.io/service-name" for the service name, and
"multicluster.kubernetes.io/source-cluster" for the source cluster.

This handles existing EndpointSlices with the "legacy" labels (without
replacing them).

Fixes: #680


pkg/endpointslice/map.go
```diff
@@ -98,7 +98,11 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 		return
 	}
 
-	cluster, ok := es.Labels[constants.LabelSourceCluster]
+	cluster, ok := es.Labels[constants.MCSLabelSourceCluster]
+
+	if !ok {
+		cluster, ok = es.Labels[constants.LighthouseLabelSourceCluster]
```
tpantelis, line 105: Perhaps we don't need to check the legacy label here since the endpoints controller should automatically update each object with the new labels on startup.

pkg/serviceimport/map.go
```diff
@@ -148,7 +148,7 @@ func (m *Map) Put(serviceImport *mcsv1a1.ServiceImport) {
 		}
 
 		if serviceImport.Spec.Type == mcsv1a1.ClusterSetIP {
-			clusterName := serviceImport.GetLabels()[lhconstants.LabelSourceCluster]
+			clusterName := serviceImport.GetLabels()[lhconstants.LighthouseLabelSourceCluster]
```
tpantelis, line 151: Shouldn't this use `MCSLabelSourceCluster`?

plugin/lighthouse/handler_test.go
```diff
@@ -1042,7 +1042,7 @@ func newServiceImport(namespace, name, clusterID, serviceIP, portName string,
 				"origin-namespace": namespace,
 			},
 			Labels: map[string]string{
-				lhconstants.LabelSourceCluster: clusterID,
+				lhconstants.LighthouseLabelSourceCluster: clusterID,
```
tpantelis, line 1045:
```suggestion
				lhconstants.MCSLabelSourceCluster: clusterID,
```

pkg/serviceimport/controller_test.go
```diff
@@ -153,7 +153,7 @@ func newServiceImport(namespace, name, serviceIP, clusterID string) *mcsv1a1.Ser
 				"origin-namespace": namespace,
 			},
 			Labels: map[string]string{
-				lhconstants.LabelSourceCluster: clusterID,
+				lhconstants.LighthouseLabelSourceCluster: clusterID,
```
tpantelis, line 156:
```suggestion
				lhconstants.MCSLabelSourceCluster: clusterID,
```

tpantelis (changes requested):

pkg/serviceimport/map.go
skitt: Not on a ServiceImport; the new labels are for EndpointSlices only.

aswinsuryan (approved):

pkg/endpointslice/map.go
vthapar: It should, but this map is populated in coredns at startup too. So for upgrade scenarios [old object without mcs label] it may not have the label yet. It will still work, but there will be a small window where we may refuse requests even when available. Am okay either ways.

vthapar (approved):

pkg/endpointslice/map.go
skitt: I’d rather keep the lenient checks; I’ve added comments to indicate when they can be removed.

tpantelis (approved):

PR #682, opened 2022-02-09
Clean up EndpointSlices and ServiceImports on uninstall

tpantelis:
Delete all from the local datastore and delete local objects from the broker.

Depends on https://github.com/submariner-io/admiral/pull/318
Depends on https://github.com/submariner-io/admiral/pull/319

### DIFF
--- a/pkg/agent/controller/cleanup.go
+++ b/pkg/agent/controller/cleanup.go
@@ -1,0 +1,90 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	lhconstants "github.com/submariner-io/lighthouse/pkg/constants"
+	discovery "k8s.io/api/discovery/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+)
+
+var (
+	serviceImportGVR = schema.GroupVersionResource{
+		Group:    mcsv1a1.GroupName,
+		Version:  mcsv1a1.SchemeGroupVersion.Version,
+		Resource: "serviceimports",
+	}
+
+	endpointSliceGVR = schema.GroupVersionResource{
+		Group:    discovery.GroupName,
+		Version:  discovery.SchemeGroupVersion.Version,
+		Resource: "endpointslices",
+	}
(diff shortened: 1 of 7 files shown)


aswinsuryan (approved):

pkg/agent/controller/cleanup.go
```diff
@@ -0,0 +1,76 @@
...
+	"k8s.io/klog"
+)
+
+func (a *Controller) Cleanup() error {
+	err := a.cleanupResources(a.serviceImportClient.Namespace(metav1.NamespaceAll), a.serviceImportSyncer, &metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	return a.cleanupResources(a.endpointSliceClient.Namespace(metav1.NamespaceAll), a.endpointSliceSyncer, &metav1.ListOptions{
```
vthapar: If broker is on data cluster, this will end up deleting everything on broker. Does uninstall run when a cluster is unjoined? We should filter out broker namespace and in broker namespace filter out anything that is not local.

pkg/agent/main.go
```diff
@@ -116,10 +114,23 @@ func main() {
 		klog.Fatalf("Failed to create lighthouse agent: %v", err)
 	}
 
+	if agentSpec.Uninstall {
+		klog.Info("Uninstalling lighthouse")
+
+		err := lightHouseAgent.Cleanup()
```
vthapar, line 120: Why not do this after all controllers are done? We run risk of syncing newer objects in between List() and Delete() in cleanup. On second thoughts, won't DeleteColletion() work out better than List() and Delete()?

pkg/agent/controller/cleanup.go
```diff
@@ -0,0 +1,76 @@
...
+		LabelSelector: labels.Set(map[string]string{discovery.LabelManagedBy: lhconstants.LabelValueManagedBy}).String(),
+	})
+}
+
+func (a *Controller) cleanupResources(client dynamic.ResourceInterface, syncer *broker.Syncer, listOptions *metav1.ListOptions) error {
+	list, err := client.List(context.TODO(), *listOptions)
```
vthapar: Why List() and Delete() instead of DeleteCollection()?

vthapar (changes requested):

pkg/agent/controller/cleanup.go
tpantelis: > If broker is on data cluster, this will end up deleting everything on broker. 

Good point. 

> Does uninstall run when a cluster is unjoined? 

It will be invoked by the operator when the ServiceDiscovery object is being deleted.

> We should filter out broker namespace and in broker namespace filter out anything that is not local.

It does the latter by checking `LabelSourceCluster`.

pkg/agent/controller/cleanup.go
tpantelis: yeah I didn't think of using `DeleteCollection` - it should work.

pkg/agent/main.go
tpantelis: I never start the controllers

pkg/agent/controller/cleanup.go
tpantelis: Actually I don't have a broker client on hand - just the `Federator` interface which doesn't have `DeleteCollection`.

tpantelis: Added `GetBrokerClient` with https://github.com/submariner-io/admiral/pull/318 - now using `DeleteCollection`.

pkg/agent/main.go
vthapar: Just cutious, how uninstall fits in agent lifecycle? User sets it in config CR, which will come in update. Will it result in restart of agent? IIRC, this code is only run once when controller starts. How will this work for already running agent unless it is restarted once Uninstall is set in CR?

EDIT: nvm, https://github.com/submariner-io/submariner-operator/pull/1843 clarifies it.

vthapar (commented):
Looks good to go.

vthapar (approved):

PR #683, opened 2022-02-11
Build images without Dapper

skitt:
We can achieve the same effect as running with Dapper, without Dapper,
by building images in multiple stages. The first stage runs in a
container built from our Shipyard base image, and subsequent stages
build our deliverable images.

This gives us more flexibility when changing build systems (building
containers within containers doesn't work well everywhere). It also
allows us to build the right binaries for multi-arch builds (since the
binary is built inside the container on the target platform).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,8 @@ bin/lighthouse-coredns: coredns/vendor/modules.txt $(shell find coredns)
 	cd coredns && ${SCRIPTS_DIR}/compile.sh $@ . $(BUILD_ARGS)
 	mv coredns/$@ $@
 
+e2e: vendor/modules.txt
+
 licensecheck: BUILD_ARGS=--noupx
 licensecheck: $(BINARIES) bin/lichen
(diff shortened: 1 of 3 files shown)


dfarrell07 (dismissed):

dfarrell07:
@sridhargaddam Can you take a look at this one?

dfarrell07 (approved):

mkolesnik (commented):
Let's discuss here first  https://github.com/submariner-io/submariner/pull/1676#issuecomment-1086866786

dfarrell07:
My review is very old, but this still looks good to me. :heavy_check_mark:

dfarrell07:
@tpantelis Can you take a look at this one?

tpantelis:
> @tpantelis Can you take a look at this one?

It looks OK to me but I'd prefer @mkolesnik give the final blessing.

tpantelis (approved):

PR #691, opened 2022-02-22
Fix agent cleanup failures

tpantelis:
`DeleteCollection` requires a specific RBAC permission which would need to be set in various places so convert the cleanup to use `List`/`Delete` instead to avoid this.

Depends on https://github.com/submariner-io/admiral/pull/321

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/onsi/gomega v1.18.1
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
-	github.com/submariner-io/admiral v0.12.0-m3.0.20220211050139-69a40598bdd6
+	github.com/submariner-io/admiral v0.12.0-m3.0.20220223161649-65232cddf5c0
 	github.com/submariner-io/shipyard v0.12.0-m3.0.20220217165059-b87e9080b8d1
 	github.com/uw-labs/lichen v0.1.5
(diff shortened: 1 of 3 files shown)


sridhargaddam:
> `DeleteCollection` requires a specific RBAC permission which would need to be set in various places so convert the cleanup to use `List`/`Delete` instead to avoid this.

@tpantelis If its an issue only about RBAC permission, probably it will be easier to add it, as the solution will be more optimal/robust compared to list/delete right? Apart from RBAC is there is any other reason for choosing list/delete?

tpantelis:
> > `DeleteCollection` requires a specific RBAC permission which would need to be set in various places so convert the cleanup to use `List`/`Delete` instead to avoid this.
> 
> @tpantelis If its an issue only about RBAC permission, probably it will be easier to add it, as the solution will be more optimal/robust compared to list/delete right? Apart from RBAC is there is any other reason for choosing list/delete?

The RBAC permissions would require changes in several places in `submariner-operator ` b/c we also need the permissions for the broker role. This would also need a change in the add-on (and maybe the helm charts?).  I don't think it's worth the risk of missing something as this code will run infrequently. 

However I got a couple other errors as well. One was "_the server does not allow this method on the requested resource_" which occurred for `EndpointSlice` which indicates to me that the resource doesn't allow "deletecollection".

sridhargaddam (approved):

aswinsuryan (approved):

PR #694, opened 2022-02-24
Build Lighthouse images for x86-64 and ARM64

skitt:
This relies on cross-building support in the Shipyard base image and
project Makefiles to avoid having to provide a native Shipyard base
image (and slower compilation using QEMU). The final image build stage
relies on support for the target architecture however.

The TARGETPLATFORM default is dropped from the container definitions;
it is expected to be provided by the build environment (buildx
currently), and setting it prevents us from actually building foreign
images.

Depends on https://github.com/submariner-io/lighthouse/pull/683
Depends on https://github.com/submariner-io/shipyard/pull/721
Depends on https://github.com/submariner-io/shipyard/pull/728

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -1,0 +1,19 @@
+---
+name: Multi-arch Builds
+
+on:
+  pull_request:
+
+jobs:
+  check-multiarch:
+    name: Check the multi-arch builds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - name: Set up QEMU (to support building on non-native architectures)
+        uses: docker/setup-qemu-action@8b122486cedac8393e77aa9734c3528886e4a1a8
+      - name: Build the multi-arch images
+        run: make multiarch-images
+      - name: Check that we actually build multi-arch images
+        run: bash -c '[ "$(echo package/*.tar)" != "package/*.tar" ]'
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,19 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-BINARIES := bin/lighthouse-agent bin/lighthouse-coredns
+gotodockerarch = $(patsubst arm,arm/v7,$(1))
+dockertogoarch = $(patsubst arm/v7,arm,$(1))
+
+nullstring :=
+space := $(nullstring) # end of the line
+comma := ,
+
+PLATFORMS ?= linux/amd64,linux/arm64
+BINARIES := lighthouse-agent lighthouse-coredns
+ARCH_BINARIES := $(foreach platform,$(subst $(comma),$(space),$(PLATFORMS)),$(foreach binary,$(BINARIES),bin/$(call gotodockerarch,$(platform))/$(binary)))
 IMAGES := lighthouse-agent lighthouse-coredns
 PRELOAD_IMAGES := submariner-gateway submariner-operator submariner-route-agent $(IMAGES)
+MULTIARCH_IMAGES := $(IMAGES)
 SETTINGS = $(DAPPER_SOURCE)/.shipyard.e2e.yml
 
 include $(SHIPYARD_DIR)/Makefile.inc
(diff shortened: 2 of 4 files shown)


dfarrell07 (dismissed):

sridhargaddam (dismissed):

skitt:
I’m dismissing the reviews since reviews on PRs on the backlog don’t make much sense.

Makefile
```diff
@@ -5,9 +5,19 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-BINARIES := bin/lighthouse-agent bin/lighthouse-coredns
+gotodockerarch = $(patsubst arm,arm/v7,$(1))
+dockertogoarch = $(patsubst arm/v7,arm,$(1))
```
mkolesnik, line 9: This is already defined is Submariner's `Makefile` and I suspect would be needed in other projects (operator?)

Perhaps it's best to defined these functions in `Makefile.inc`?

Makefile
```diff
@@ -5,9 +5,19 @@ ifneq (,$(DAPPER_HOST_ARCH))
...
+
+nullstring :=
+space := $(nullstring) # end of the line
+comma := ,
```
mkolesnik, line 13: This too :) ?

Makefile
```diff
@@ -19,24 +29,19 @@ override DEPLOY_ARGS += --service_discovery
...
+bin/%/lighthouse-coredns: vendor/modules.txt $(shell find pkg/coredns)
+	GOARCH=$(call dockertogoarch,$(patsubst bin/linux/%/,%,$(dir $@))) ${SCRIPTS_DIR}/compile.sh $@ pkg/coredns/main.go $(BUILD_ARGS)
 
 e2e: vendor/modules.txt
 
 licensecheck: BUILD_ARGS=--noupx
-licensecheck: $(BINARIES) bin/lichen
-	bin/lichen -c .lichen.yaml $(BINARIES)
+licensecheck: $(ARCH_BINARIES) bin/lichen
+	bin/lichen -c .lichen.yaml $(ARCH_BINARIES)
```
mkolesnik, line 46: I wonder if it's really necessary to do it for all the binaries, and not just one arch (shouldnt they have similar license regardless of platform?)

package/Dockerfile.lighthouse-agent
```diff
@@ -1,20 +1,20 @@
...
 ARG SOURCE
 ARG TARGETPLATFORM
 
 COPY . ${SOURCE}
 
-RUN make -C ${SOURCE} LOCAL_BUILD=1 bin/lighthouse-agent
+RUN make -C ${SOURCE} LOCAL_BUILD=1 bin/${TARGETPLATFORM}/lighthouse-agent
 
-FROM scratch
+FROM --platform=${TARGETPLATFORM} scratch
```
mkolesnik, line 12: This seems redundant per [this blog post]?

> For this stage, we will not use --platform in the FROM definition. We could write FROM --platform=$TARGETPLATFORM but that is the default value for all builds stages anyway so using a flag is redundant.

[this blog post]: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

Makefile
skitt: Yes, but I needed them here to have a working PR ;-). I’ll move everything appropriate to Shipyard once this gets merged.

Makefile
skitt: Some platforms can result in different dependencies (_e.g._ Windows).

package/Dockerfile.lighthouse-agent
skitt: It’s redundant per the official docs too, but I included it for balance. I thought it preferable to be explicit, so that readers wouldn’t have to wonder what the platform is when it’s not specified.

skitt:
The UPX logs in the Multi-arch job show that this is building amd64 and arm64 binaries and images.

.github/workflows/multiarch.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Multi-arch builds
```
dfarrell07:
```suggestion
name: Multi-arch Builds
```

tpantelis (approved):

.github/workflows/multiarch.yml
dfarrell07: We can go ahead and merge this, fix the name in a follow-up commit as was done in other repos.

dfarrell07 (approved):

PR #700, opened 2022-03-03
Add packagedoc-lint linting, fix errors

dfarrell07:
The idea here is to make sure the license headers are detached from the
package declarations, as otherwise the headers are considered a part of
the package documentation and included with it.

Add per-PR job to maintain consistent standards across all repos.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -94,6 +94,15 @@ jobs:
       - name: Run markdownlint
         run: make markdownlint
 
+  packagedoc-lint:
+    name: Package Documentation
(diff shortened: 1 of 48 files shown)


skitt (approved):
This won’t affect backports to 0.12, so it seems safe to merge without waiting for the start of 0.13.

dfarrell07:
> This won’t affect backports to 0.12, so it seems safe to merge without waiting for the start of 0.13.

@skitt As we discussed on some call last week, I think we should backport these packagedoc fixes, as they are bugfixes for our auto-gen'd Go docs.

mkolesnik (approved):

PR #715, opened 2022-03-24
Only retain last state for condition types

### LINKED ISSUE #714: Submariner ServiceExport does not have unique condition types to wait for readiness
**What happened**:
After creating service export with subctl, ther's no way to validate it with kubectl/oc wait command,
since Submariner ServiceExport does not have unique condition types to wait for readiness (all types are the same "Valid").

For example, on a working service export status it failed on:

$ oc wait serviceexport nginx-cl-bc --for=condition=Valid
error: timed out waiting for the condition on serviceexports/nginx-cl-bc

If the service export is working, it should return something like:

multicluster.x-k8s.io/ServiceExportList condition met

Create a service export:
$ oc get serviceexport
NAME AGE
nginx-cl-bc 3h5m

$ oc get serviceexport -o yaml
apiVersion: v1
items:

apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceExport
metadata:
creationTimestamp: "2022-03-21T08:25:40Z"
generation: 1
name: nginx-cl-bc
namespace: test-submariner
resourceVersion: "115495575"
selfLink: /apis/multicluster.x-k8s.io/v1alpha1/namespaces/test-submariner/serviceexports/nginx-cl-bc
uid: 7af4a9ae-24f6-4d1d-9ac5-100664e03dd9
status:
conditions:
lastTransitionTime: "2022-03-21T08:25:39Z"
message: Service doesn't have a global IP yet
reason: ServiceGlobalIPUnavailable
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T08:25:39Z"
message: Awaiting sync of the ServiceImport to the broker
reason: AwaitingSync
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T08:25:39Z"
message: Service was successfully synced to the broker
reason: ""
status: "True"
type: Valid
lastTransitionTime: "2022-03-21T09:59:32Z"
message: Service to be exported doesn't exist
reason: ServiceUnavailable
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T09:59:32Z"
message: Service doesn't have a global IP yet
reason: ServiceGlobalIPUnavailable
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T09:59:33Z"
message: Awaiting sync of the ServiceImport to the broker
reason: AwaitingSync
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T09:59:35Z"
message: Service was successfully synced to the broker
reason: ""
status: "True"
type: Valid
lastTransitionTime: "2022-03-21T10:20:16Z"
message: Service to be exported doesn't exist
reason: ServiceUnavailable
status: "False"
type: Valid
lastTransitionTime: "2022-03-21T10:20:16Z"
message: Service doesn't have a global IP yet
reason: ServiceGlobalIPUnavailable
status: "False"
type: Valid
kind: List
metadata:
resourceVersion: ""
selfLink: ""
$ oc wait serviceexport nginx-cl-bc --for=condition=Valid
error: timed out waiting for the condition on serviceexports/nginx-cl-bc

**Environment**:
Submariner version 0.12.0

- Install tools:
- Others:

vthapar:
Condition types are expected to be singelton and only last
state of a given condition type shoud be maintained. We currently
maintain a list of all transitions to a state as a list with max of
10. This causes problems with condition check APIs. Currently
support just one type, Valid.

Fixes #714

### DIFF
--- a/pkg/agent/controller/agent.go
+++ b/pkg/agent/controller/agent.go
@@ -56,8 +56,6 @@ type AgentConfig struct {
 	ServiceExportCounterName string
 }
 
-var MaxExportStatusConditions = 10
-
 // nolint:gocritic // (hugeParam) This function modifies syncerConf so we don't want to pass by pointer.
 func New(spec *AgentSpecification, syncerConf broker.SyncerConfig, kubeClientSet kubernetes.Interface,
 	syncerMetricNames AgentConfig) (*Controller, error) {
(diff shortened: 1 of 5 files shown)


pkg/agent/controller/agent.go
```diff
@@ -410,20 +408,14 @@ func (a *Controller) updateExportedServiceStatus(name, namespace string, status
 			Message:            &msg,
 		}
 
-		numCond := len(toUpdate.Status.Conditions)
-		if numCond > 0 && serviceExportConditionEqual(&toUpdate.Status.Conditions[numCond-1], &exportCondition) {
+		// TODO: Currently we only check for conditionType Valid. Revisit this when Conflict is supported.
```
tpantelis, line 411: To preserve the equivalent of the historical trace, we could add other condition types like "SyncedToBroker". But we can look into that later.

tpantelis (approved):

pkg/agent/controller/agent.go
vthapar: We should, but we will have to propose these to MCS sig first.

tpantelis: I wouldn't think so. The MCS API defines some common, general ones but I'd think an implementation should be able to define their own specific types - the MCS spec doesn't preclude that. Eg, the type "SyncedToBroker" would be specific to our implementation.

sridhargaddam (approved):

dfarrell07:
@vthapar Do you agree this should be backported? Do you want to send it?

PR #720, opened 2022-04-05
Switch EndpointSlices to discovery.k8s.io/v1

### LINKED ISSUE #576: Switch EndpointSlices to discovery.k8s.io/v1
This will need to be done by the summer of 2022.

In K8s 1.20, discovery.k8s.io/v1 was added. v1beta1 will no longer be served starting with 1.25, so before that happens we need to migrate to /v1 (assuming we no longer need to support versions before 1.20).

This is also a blocker for the migration to CoreDNS 1.8.4 which relies on /v1.

vthapar:
Fixes #576


skitt (approved):

skitt:
This change has a nasty effect on the Lighthouse unit tests.

Jaanki (approved):

tpantelis (approved):

PR #736, opened 2022-04-27
Bump all Lighthouse E2E timeouts from 30m to 45m

dfarrell07:
It seems like successful jobs can finish in about 20m, but if there are
redeploy/retries then the jobs typically time out in the post mortem.

This is already the timeout used in the main repo.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

PR #735, opened 2022-04-27
Use latest `Shipyard` for simpleserver

mkolesnik:
Shipyard's E2E framework was updated to use a simple server, the
consuming projects need to update in order to properly consume it.

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>

PR #737, opened 2022-04-27
Use new nettest simpleserver from Shipyard

dfarrell07:
The nginx image is giving us maintenance issues, use a simple netcat
process run from an existing container instead.


dfarrell07:
The upgrade tests are still correctly failing, see https://github.com/submariner-io/lighthouse/pull/736#issuecomment-1111213026 for details.

dfarrell07:
There are _lots_ of nginx references in Lighthouse. To properly move away from it, even in names, would be a major change.

```
[~/go/src/submariner-io/lighthouse]$ grep -riI --exclude-dir=vendor nginx . | wc -l
324
```

Also wondering if one of these usages is something the `nc` nifty hack doesn't support?

dfarrell07:
The E2E failures here are very similar to the ones on Mike's PR with only the Shipyard version bump.

> Failed to  service IP verification. expected execution result "10.2.96.4" to be empty

https://github.com/submariner-io/lighthouse/runs/6198528492?check_suite_focus=true#step:3:2239

> Failed to retrieve Endpoints. endpoints have 0 IPs when expected 3

https://github.com/submariner-io/lighthouse/runs/6198528492?check_suite_focus=true#step:3:2295

> Failed to  service IP verification. expected execution result "10.2.96.4\n10.1.0.6" to be empty

https://github.com/submariner-io/lighthouse/runs/6198528492?check_suite_focus=true#step:3:2584

mkolesnik (dismissed):
Can you rebase on top of #738 instead of #735 ?

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
 	github.com/submariner-io/admiral v0.12.0-m3.0.20220223161649-65232cddf5c0
-	github.com/submariner-io/shipyard v0.12.0-m3.0.20220421170648-adf7571e55ed
+	github.com/submariner-io/shipyard v0.12.0-m3.0.20220427131749-6b99f918e45e
```
mkolesnik: No need to do this commit, the image is available through Quay and you're just using it

test/e2e/framework/framework.go
```diff
@@ -394,15 +394,15 @@ func (f *Framework) NewNginxStatefulSet(cluster framework.ClusterIndex) *appsv1.
 					Containers: []v1.Container{
 						{
 							Name:            statefulServiceName,
-							Image:           "quay.io/testing-farm/nginx:latest",
+							Image:           "localhost:5000/nettest:local",
```
mkolesnik: This assumes that it was loaded to a local registry which might not be true when calling `subctl verify`, better use `quay.io/submariner/nettest:devel`

aswinsuryan (approved):

skitt (approved):

tpantelis (approved):

dfarrell07:
This is backported here: https://github.com/submariner-io/lighthouse/pull/748

PR #738, opened 2022-04-28
Move `deployments.go` to Lighthouse

mkolesnik:
These functions `NewNetShootDeployment` and `NewNginxDeployment` are
only used in LH E2E, so it makes sense this lives here.

This simplifies maintenance as any change to these functions will
necessitate only changing them in LH, instead of forcing a change in
Shipyard as well.

The `FindDeployment` function is unused so it wasn't copied over.

### DIFF
--- a/test/e2e/framework/deployments.go
+++ b/test/e2e/framework/deployments.go
@@ -1,0 +1,127 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"context"
+
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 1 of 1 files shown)


vthapar (commented):
Thanks for this. At the time idea was we will need these in submariner, but it never happened.

Changes look  fine but want to confirm that we don't want any changes in test code in `test/e2e/discovery`.  E2E will still pass on this patch if it is picking these functions off shipyard framework but will break when we remove from shipyard.

mkolesnik:
> Changes look fine but want to confirm that we don't want any changes in test code in `test/e2e/discovery`. E2E will still pass on this patch if it is picking these functions off shipyard framework but will break when we remove from shipyard.

Not sure why would it break, the `Framework` instance is the one defined in LH which promotes these methods from Shipayrd's `Framework`. Once we remove them from Shipyard then they won't be promoted, but it would still use the ones here.

If the ones copied here weren't being used instead of the promoted ones, then the E2E tests would still be failing, IIUC.

vthapar:
Yeah, that is what I thought too but wanted to confirm my understanding was correct.

vthapar (approved):

tpantelis (approved):

test/e2e/framework/deployments.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (f *Framework) NewNetShootDeployment(cluster framework.ClusterIndex) *corev1.PodList {
+	var replicaCount int32 = 1
```
sridhargaddam, line 32: The replicaCount is set to 1, so it looks like the requirement is just to spawn a Pod which can be used to run a command. We already have APIs in Shipyard to spawn customPods which can be used for this purpose - https://github.com/submariner-io/shipyard/blob/6b99f918e45e1dcae9e977cb294cf09065266c44/test/e2e/framework/network_pods.go#L124

and its used in Submariner e2e tests here - https://github.com/submariner-io/submariner/blob/c047698ccbf1fcc43b2dcd7c0773acc902914f61/test/e2e/framework/dataplane.go#L143

I think we can remove this API and directly use the CustomPod interface from Shipyard.

mkolesnik:
Backported to 0.12 as #739

PR #744, opened 2022-05-02
Move the CoreDNS plugin to its own module

skitt:
This simplifies dependencies in Lighthouse itself (and its own
dependencies, e.g. the operator), and allows us to dissociate the
dependency trees of the Lighthouse agent and CoreDNS plugins.

This allows us to drop all Kubernetes pinning, and let the plugin
upgrade to v0.23.4 (as pulled in by the CoreDNS dependencies). As a
result, the build requires Go 1.17 or later.

Depends on https://github.com/submariner-io/shipyard/pull/775


go.mod
```diff
@@ -3,22 +3,26 @@ module github.com/submariner-io/lighthouse
...
 	github.com/prometheus/client_golang v1.12.1
 	github.com/submariner-io/admiral v0.12.0-m3.0.20220223161649-65232cddf5c0
 	github.com/submariner-io/shipyard v0.12.0-m3.0.20220421170648-adf7571e55ed
 	github.com/uw-labs/lichen v0.1.5
+	golang.org/x/sys v0.0.0-20220209214540-3681064d5158 // indirect
+	golang.org/x/time v0.0.0-20211116232009-f0f3c7e86c11 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	k8s.io/api v0.23.4
 	k8s.io/apimachinery v0.23.4
 	k8s.io/client-go v11.0.0+incompatible
```
tpantelis: Can we now change these to v0.21.11?

go.mod
```diff
@@ -3,22 +3,26 @@ module github.com/submariner-io/lighthouse
...
 	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.40.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
 	k8s.io/utils v0.0.0-20211116205334-6203023598ed
 	sigs.k8s.io/controller-runtime v0.7.2
```
tpantelis, line 27: Might as well bump this.

```suggestion
	sigs.k8s.io/controller-runtime v0.9.7
```

tpantelis (changes requested):

go.mod
skitt: I’d rather not make changes unrelated to the plugin split.

go.mod
skitt: All the changes in the base `go.mod` were automated, I’d rather not add manual changes. `go mod tidy` always bumps `coredns/go.mod`’s `k8s.io` dependencies to v0.23.4, so I can’t specify v0.21.11 there either.

coredns/go.mod
```diff
@@ -0,0 +1,29 @@
...
+require (
+	github.com/caddyserver/caddy v1.0.5
+	github.com/coredns/caddy v1.1.1
+	github.com/coredns/coredns v1.9.1
+	github.com/miekg/dns v1.1.48
+	github.com/onsi/ginkgo v1.16.5
+	github.com/onsi/gomega v1.19.0
+	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.12.1
+	github.com/submariner-io/lighthouse v0.12.0-m3.0.20220502131803-864fecb47b2f
```
mkolesnik, line 14: What's the implication here?

We will have to update it during release?

Also will LH changes pertaining to both now have to be done in 2 steps?

Local testing of changes is now more problematic?

skitt: I’ve added a local replace so we don’t need to worry about any of this in practice.

go.mod
tpantelis: So **submariner-operator** had dependency issues b/c **lighthouse** was upgraded to v0.23.4 b/c of `coredns` bump. I thought by splitting out `coredns` as you've done would allow this `go.mod` to specify v0.21.11. Or at least there wouldn't be any dependency automatically pulling in v0.23.4. I mean if this `go.mod` specified v0.21.11 it would avoid the extra pinning in **submariner-operator**, ie `logr`, `klog` etc.

skitt: Ah right, good point. I was thinking in the other direction, that our Lighthouse dependencies were preventing us from upgrading dependencies in the CoreDNS plugin (I get regular emails about the version of etcd that we pull in). I’ll look into why we still pull in v0.23.4 here then.

skitt: Done, both `go.mod`s can be simplified quite a bit, but this should end up requiring Go 1.17 (see https://github.com/submariner-io/shipyard/pull/775).

tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
> Linting / Dependency Licenses (pull_request) Failing after 3m

Waiting on Dependabot to send a PR with the fix :watch:

tpantelis:
> > Linting / Dependency Licenses (pull_request) Failing after 3m
> 
> Waiting on Dependabot to send a PR with the fix ⌚

We don't need to wait 😉

PR #749, opened 2022-05-05
Bump markdown-link-check to the current tip

skitt:
This hasn't been released yet but fixes the safe directories git
errors.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -79,7 +79,7 @@ jobs:
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
       - name: Run markdown-link-check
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

dfarrell07:
Do you think we should ask them to cut a release with this fix? I see it's in the v1 tag, but that doesn't help projects using Dependabot to bump to latest-release SHAs.

dfarrell07:
> Do you think we should ask them to cut a release with this fix? I see it's in the v1 tag, but that doesn't help projects using Dependabot to bump to latest-release SHAs.

Well I see you've already sent PRs to many places, so I guess we don't need this.

dfarrell07 (approved):

PR #757, opened 2022-05-12
Return local cluster IP for headless services

### LINKED ISSUE #656: GlobalNet-Configured Exported Headless Service Pod's FQDN Not DNS-Resolvable to Itself
**What happened**:
GlobalNet-Configured exported headless service pod's FQDN was DNS-resolvable to other joined clusters but NOT resolvable to itself.

From within headless service pod, I expected to be able to curl the local cluster's service via `<pod>.<cluster>.<service>.<namespace>.clusterset.local` FQDN.

I deployed a MongoDB replica set using the [bitnami/mongodb helm charts](https://github.com/bitnami/charts/tree/master/bitnami/mongodb) across three Submariner clusters which is too long to describe here; however, a much simpler test could be performed using [Submariner User Guide's Test StatefulSet and Headless Service](https://submariner.io/operations/usage/#test-statefulset-and-headless-service) setup steps and then just curl the exported headless service using the FQDN (e.g. **_curl web-0.cluster3.nginx-ss.nginx-test.svc.clusterset.local_**) from within the pod itself.

My work-around was to add the `<pod>.<cluster>.<service>.<namespace>.clusterset.local` FQDN to the pod's **_/etc/hosts_** via [hostAliases](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/).
**Environment**: 
- Diagnose information (use `subctl diagnose all`):  unable to provide since clusters are on private air-gapped (no Internet) bare metal servers
- Gather information (use `subctl gather`):  unable to provide since clusters are on private air-gapped (no Internet) bare metal servers
- Cloud provider or hardware configuration:  private air-gapped (no Internet) bare metal servers (CentOS 7)
- Install tools:  RKE
- Others:  Submariner v0.10.0, Kubernetes v1.19.8

mikenorgate:
This fixes issue #656 where DNS resolution of a headless services 
does not return the local IP for pods within the local cluster

### DIFF
--- a/coredns/endpointslice/map.go
+++ b/coredns/endpointslice/map.go
@@ -19,12 +19,19 @@ limitations under the License.
 package endpointslice
 
 import (
+	"context"
 	"sync"
 
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/lighthouse/coredns/serviceimport"
 	"github.com/submariner-io/lighthouse/pkg/constants"
 	discovery "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog"
 	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
@@ -40,8 +47,10 @@ type clusterInfo struct {
 }
 
 type Map struct {
-	epMap map[string]*endpointInfo
-	mutex sync.RWMutex
+	epMap          map[string]*endpointInfo
+	mutex          sync.RWMutex
+	localClusterID string
+	kubeClient     kubernetes.Interface
 }
 
 func (m *Map) GetDNSRecords(hostname, cluster, namespace, name string, checkCluster func(string) bool) ([]serviceimport.DNSRecord, bool) {
@@ -86,9 +95,11 @@ func (m *Map) GetDNSRecords(hostname, cluster, namespace, name string, checkClus
 	}
 }
 
-func NewMap() *Map {
+func NewMap(localClusterID string, kubeClient kubernetes.Interface) *Map {
 	return &Map{
-		epMap: make(map[string]*endpointInfo),
+		epMap:          make(map[string]*endpointInfo),
+		localClusterID: localClusterID,
+		kubeClient:     kubeClient,
 	}
 }
 
@@ -109,6 +120,39 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 	if !ok {
 		klog.Warningf("Cluster label missing on %#v", es.ObjectMeta)
 		return
+	}
+
+	// TODO: Only do this look up if globalnet enabled to reduce unnessary call to Kube API
+	var localSlice *discovery.EndpointSlice
+	if cluster == m.localClusterID {
+		retryErr := retry.OnError(retry.DefaultBackoff, func(err error) bool {
+			return !apierrors.IsNotFound(err)
+		}, func() error {
+			localSlices, err := m.kubeClient.DiscoveryV1().EndpointSlices(es.Labels[constants.LabelSourceNamespace]).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: labels.Set(map[string]string{
+					constants.KubernetesServiceName: es.Labels[constants.MCSLabelServiceName],
+				}).String(),
+			})
+			if err != nil {
+				return err
+			}
+
+			if len(localSlices.Items) == 0 {
+				return apierrors.NewNotFound(schema.GroupResource{
+					Group:    discovery.SchemeGroupVersion.Group,
+					Resource: "endpointslices",
+				}, "")
+			}
+			localSlice = &localSlices.Items[0]
+			return nil
+		})
+		if retryErr != nil {
+			klog.Errorf("Error finding local endpoint slice for service (%s/%s): %+v", es.Labels[constants.LabelSourceNamespace], es.Labels[constants.MCSLabelServiceName], retryErr)
+			return
+		}
+		if localSlice == nil {
+			return
+		}
 	}
 
 	m.mutex.Lock()
@@ -142,7 +186,18 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 	for _, endpoint := range es.Endpoints {
 		var records []serviceimport.DNSRecord
 
-		for _, address := range endpoint.Addresses {
+		addresses := endpoint.Addresses
+		if localSlice != nil {
+			for _, localEndpoint := range localSlice.Endpoints {
+				if localEndpoint.NodeName != nil && endpoint.NodeName != nil && *localEndpoint.NodeName == *endpoint.NodeName &&
+					(endpoint.Hostname == nil || localEndpoint.TargetRef.Name == *endpoint.Hostname) {
+					addresses = localEndpoint.Addresses
+				}
+			}
+		}
+
+		for _, address := range addresses {
+
 			record := serviceimport.DNSRecord{
 				IP:          address,
 				Ports:       mcsPorts,
--- a/test/e2e/discovery/headless_services.go
+++ b/test/e2e/discovery/headless_services.go
@@ -53,6 +53,10 @@ var _ = Describe("[discovery] Test Headless Service Discovery Across Clusters", 
 		It("should only resolve the IPs from the active pods", func() {
 			RunHeadlessPodsAvailabilityTest(f)
 		})
+
+		It("should resolve the local pod IPs", func() {
+			RunHeadlessPodsAvailabilityTestLocal(f)
+		})
 	})
 
 	When("a pod tries to resolve a headless service in a specific remote cluster by its cluster name", func() {
@@ -80,10 +84,10 @@ func RunHeadlessDiscoveryTest(f *lhframework.Framework) {
 
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 
-	ipList, hostNameList := f.GetPodIPs(framework.ClusterB, nginxHeadlessClusterB)
-	verifyHeadlessIpsWithDig(f.Framework, framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
+	ipList, hostNameList := f.GetPodIPs(framework.ClusterB, nginxHeadlessClusterB, false)
+	f.VerifyIPsWithDig(framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
 		"", true)
-	verifyHeadlessIpsWithDig(f.Framework, framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
+	f.VerifyIPsWithDig(framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
 		clusterBName, true)
 
 	verifyHeadlessSRVRecordsWithDig(f.Framework, framework.ClusterA, nginxHeadlessClusterB, netshootPodList, hostNameList, checkedDomains,
@@ -94,7 +98,7 @@ func RunHeadlessDiscoveryTest(f *lhframework.Framework) {
 	f.DeleteServiceExport(framework.ClusterB, nginxHeadlessClusterB.Name, nginxHeadlessClusterB.Namespace)
 	f.AwaitServiceImportCount(framework.ClusterA, nginxHeadlessClusterB.Name, nginxHeadlessClusterB.Namespace, 0)
 
-	verifyHeadlessIpsWithDig(f.Framework, framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
+	f.VerifyIPsWithDig(framework.ClusterA, nginxHeadlessClusterB, netshootPodList, ipList, checkedDomains,
 		"", false)
 	verifyHeadlessSRVRecordsWithDig(f.Framework, framework.ClusterA, nginxHeadlessClusterB, netshootPodList, hostNameList, checkedDomains,
 		clusterBName, false, false, false)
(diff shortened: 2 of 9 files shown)


dfarrell07:
Thank you for the PR @mikenorgate!

vthapar:
@mikenorgate Thanks for working on this. Have you tested it locally? Just looking at code I'd expect it to work for local clusters, but break for remote.

mikenorgate:
> @mikenorgate Thanks for working on this. Have you tested it locally? Just looking at code I'd expect it to work for local clusters, but break for remote.

@vthapar I've been testing it in local KinD clusters at it looks to be behaving correctly. Can you explain a bit more of what you mean by break for remote clusters?

pkg/agent/controller/endpoint.go
```diff
@@ -274,7 +275,7 @@ func allAddressesIPv6(addresses []corev1.EndpointAddress) bool {
 }
 
 func (e *EndpointController) getIP(address *corev1.EndpointAddress) string {
-	if e.isHeadless && e.globalIngressIPCache != nil {
+	if e.clusterID != e.localClusterID && e.isHeadless && e.globalIngressIPCache != nil {
```
vthapar: This will work for local cluster but break remote clusters. This is part of code where we create EndpointSlices from service's endpoints to sync to remote clusters. ClusterId check here is irrelevant as we will only hit this code for local services on lighthouse agent.

The files you need to modify and handle clusterIDs are in pkg/endpointslice/

vthapar (changes requested):

pkg/agent/controller/endpoint.go
mikenorgate: I understand what you mean now, I'll go back and try to come up with another fix

mikenorgate:
@vthapar I've changed my approach and would appreciate your feedback. I still have a couple of tests to fix up but wanted to make sure I was on the correct path now

pkg/endpointslice/map.go
```diff
@@ -111,6 +117,20 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
...
 
+	var localSlice *discovery.EndpointSlice
+	if cluster == m.localClusterID {
+		localSlices, err := localClient.EndpointSlices(es.Labels[constants.LabelSourceNamespace]).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: labels.Set(map[string]string{
+				"kubernetes.io/service-name": es.Labels[constants.MCSLabelServiceName],
+			}).String(),
+		})
+		if err != nil || len(localSlices.Items) != 1 {
+			klog.Warningf("Unable to find local endpoint slice for service")
```
vthapar: This doesn't need to be warning but debug. Normal operation will have lots of cases where we only have remote services. This will make our logs fill up with warning logs which don't add any value. A debug level will work better as then it can be used, by increasing log level, to troubleshoot specific failure.

test/e2e/discovery/headless_services.go
```diff
@@ -127,7 +127,12 @@ func RunHeadlessDiscoveryLocalAndRemoteTest(f *lhframework.Framework) {
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 
 	ipListB, hostNameListB := f.GetPodIPs(framework.ClusterB, nginxHeadlessClusterB)
-	ipListA, hostNameListA := f.GetPodIPs(framework.ClusterA, nginxHeadlessClusterA)
+	_, hostNameListA := f.GetPodIPs(framework.ClusterA, nginxHeadlessClusterA)
+
+	var ipListA []string
```
vthapar: How about modifying GetPodIPs itself to return podIPs if cluster is local? Maybe add an `isLocal` flag? That way complete logic sits in GetPodIPs and test code stays the same and easy to read.

test/e2e/discovery/headless_services.go
```diff
@@ -217,9 +222,14 @@ func RunHeadlessDiscoveryClusterNameTest(f *lhframework.Framework) {
 
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 
-	ipListClusterA, hostNameListA := f.GetPodIPs(framework.ClusterA, nginxHeadlessClusterA)
+	_, hostNameListA := f.GetPodIPs(framework.ClusterA, nginxHeadlessClusterA)
 	ipListClusterB, hostNameListB := f.GetPodIPs(framework.ClusterB, nginxHeadlessClusterB)
 
+	var ipListClusterA []string
```
vthapar: same as above

pkg/endpointslice/map.go
```diff
@@ -111,6 +117,20 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
...
 	}
 
+	var localSlice *discovery.EndpointSlice
+	if cluster == m.localClusterID {
+		localSlices, err := localClient.EndpointSlices(es.Labels[constants.LabelSourceNamespace]).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: labels.Set(map[string]string{
+				"kubernetes.io/service-name": es.Labels[constants.MCSLabelServiceName],
+			}).String(),
+		})
+		if err != nil || len(localSlices.Items) != 1 {
```
vthapar: Need to add a retry in case error is due to timeout. Those are not that uncommon if Kube API server is under heavy traffic.

pkg/endpointslice/map.go
```diff
@@ -142,7 +162,18 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 	for _, endpoint := range es.Endpoints {
 		var records []serviceimport.DNSRecord
 
-		for _, address := range endpoint.Addresses {
+		addresses := endpoint.Addresses
+		if localSlice != nil {
+			for _, localEndpoint := range localSlice.Endpoints {
+				if *localEndpoint.NodeName == *endpoint.NodeName &&
```
vthapar: Not sure if we need this check. These would be same for non-globalnet while for globalnet localslice we fetched here will be more up-to-date and recent than the one we got from event notification.

pkg/endpointslice/map.go
```diff
@@ -111,6 +117,20 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 		return
 	}
 
+	var localSlice *discovery.EndpointSlice
+	if cluster == m.localClusterID {
```
vthapar: We should also add globalnet flag and do this only for globalnet. For non globalnet this is irrelevant and will just add extra calls to Kube API server.

This probably means adding another env to lighthouse-coredns deployment. It can come in a later patch and I'd recommend adding a TODO here and creating an issue to track it.

vthapar (changes requested):
Overall approach is good and exactly what I had in mind. Some comments regarding error handling and possibe optimizations.

Would be good to create a separate E2E test for HeadlessLocal only where we test change in number of endpoints by increasing and decreasing replica set size.

vthapar:
@mikenorgate I've pushed https://github.com/submariner-io/lighthouse/pull/762 It should make things simpler for you.

pkg/endpointslice/map.go
mikenorgate: I found that the `localSlice` didn't include the hostname which is why I attempt to map back to the globalnet endpoint slice

dfarrell07:
This needs a manual rebase before re-running CI.

dfarrell07:
Thanks for the rebase @mikenorgate, I ran CI and it's showing some minor issues.

Can you re-review @vthapar?

coredns/endpointslice/map.go
```diff
@@ -111,6 +119,35 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 		return
 	}
 
+	// TODO: Only do this look up if globalnet enabled to reduce unnessary call to Kube API
```
vthapar, line 125: Can you create an issue to track this TODO item?

vthapar (approved):
Thanks for all the effort on this, looks good. Tested locally and confirmed to be working.

coredns/endpointslice/map.go
mikenorgate: Created #775

dfarrell07:
This has a number of CI failures (go linting/unit, commit message/DCO) that need to be fixed before it can be merged.

I'd suggest going ahead and squashing all the commits together into (likely 1) logically-atomic commit. More details:

https://submariner.io/development/code-review/#squashamend-commits-into-discrete-steps

vthapar:
@mikenorgate Can you resolve the linter and UT failures?

mikenorgate:
@vthapar I've fixed the linter errors but I can't see what the reason for the unit test failure is, all the tests are passing and there doesn't seem to be an error message in the output

tpantelis:
> @vthapar I've fixed the linter errors but I can't see what the reason for the unit test failure is, all the tests are passing and there doesn't seem to be an error message in the output

The failure is due to [compilation errors](https://github.com/submariner-io/lighthouse/runs/6881666688?check_suite_focus=true#step:4:921) in _coredns/plugin/handler_test.go_.

coredns/endpointslice/map.go
```diff
@@ -86,13 +93,14 @@ func (m *Map) GetDNSRecords(hostname, cluster, namespace, name string, checkClus
...
+func NewMap(localClusterID string) *Map {
 	return &Map{
-		epMap: make(map[string]*endpointInfo),
+		epMap:          make(map[string]*endpointInfo),
+		localClusterID: localClusterID,
 	}
 }
 
-func (m *Map) Put(es *discovery.EndpointSlice) {
+func (m *Map) Put(es *discovery.EndpointSlice, localClient v1.DiscoveryV1Interface) {
```
tpantelis: I think it's cleaner to pass `localClient` into the `NewMap` ctor and store in `Map`.

tpantelis (changes requested):

dfarrell07:
While you're making wonderful contributions to Submariner @mikenorgate, maybe you'd like to add your company to the ADOPTERS.md file? We have a CNCF review coming up soon and it would be great to make it reflect reality a bit more.

https://github.com/submariner-io/submariner/blob/devel/ADOPTERS.md

mikenorgate:
I've made the requested changes and fixed the unit tests

coredns/endpointslice/map.go
```diff
@@ -86,9 +95,11 @@ func (m *Map) GetDNSRecords(hostname, cluster, namespace, name string, checkClus
 	}
 }
 
-func NewMap() *Map {
+func NewMap(localClusterID string, kubeConfig *rest.Config) *Map {
```
tpantelis: It's better to pass in a `kubernetes.Interface` - we avoid passing `rest.Config` b/c there's no mock/fake implementation for unit tests.

```suggestion
func NewMap(localClusterID string, kubeClient kubernetes.Interface) *Map {
```

tpantelis (changes requested):

coredns/endpointslice/map.go
```diff
@@ -142,7 +188,18 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 	for _, endpoint := range es.Endpoints {
 		var records []serviceimport.DNSRecord
 
-		for _, address := range endpoint.Addresses {
+		addresses := endpoint.Addresses
+		if localSlice != nil {
+			for _, localEndpoint := range localSlice.Endpoints {
+				if *localEndpoint.NodeName == *endpoint.NodeName &&
```
tpantelis: Is it certain that either `NodeName` won't be nil?

coredns/endpointslice/map.go
```diff
@@ -111,6 +122,41 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
...
+			return !apierrors.IsNotFound(err)
+		}, func() error {
+			localSlices, err := clientSet.DiscoveryV1().EndpointSlices(es.Labels[constants.LabelSourceNamespace]).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: labels.Set(map[string]string{
+					constants.KubernetesServiceName: es.Labels[constants.MCSLabelServiceName],
+				}).String(),
+			})
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			} else if err != nil || len(localSlices.Items) != 1 {
```
tpantelis: We don't really need to check for `IsNotFound` error - it would be reported on L152 below anyway. Also for the `List` operation, it doesn't return `IsNotFound` if there's no items found - it only returns `IsNotFound` if the resource type isn't found which shouldn't happen.

```suggestion
			if err != nil {
				return err
			} 
			
			if len(localSlices.Items) != 1 {
```

coredns/endpointslice/map.go
```diff
@@ -111,6 +122,41 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
...
+				klog.V(log.DEBUG).Infof("Unable to find local endpoint slice for service %s/%s", es.Labels[constants.LabelSourceNamespace], es.Labels[constants.MCSLabelServiceName])
+				return nil
```
tpantelis: For the case where there's more than one `EndpointSlice` found, this log message isn't actually  accurate. Also we log to debug here as opposed to error on L152. However, do we need to fail in this case, ie can't we just take the first one? What would be the case where there might be more than one?

For the case where the returned list is empty, we could return a `NotFound` error and let if fall through to L152, ie 

```Go
  apierrors.NewNotFound(schema.GroupResource{
		Group:    discovery.SchemeGroupVersion.Group,
		Resource: "endpointslices",
	}, "")
```

coredns/endpointslice/map.go
```diff
@@ -111,6 +122,41 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
...
 
+	// TODO: Only do this look up if globalnet enabled to reduce unnessary call to Kube API
+	var localSlice *discovery.EndpointSlice
+	if cluster == m.localClusterID {
+		clientSet, err := kubernetes.NewForConfig(m.kubeConfig)
+		if err != nil {
+			klog.Warningf("Error creating client set %#v", err)
+			return
+		}
+
```
tpantelis: It would be nice to add unit tests to cover this new block of code but we can do that in another PR.

tpantelis (changes requested):

coredns/endpointslice/map.go
mikenorgate: The node name of `endpoint` wont be nil as it is set here https://github.com/submariner-io/lighthouse/blob/4a48ab0ef6769397760c6f268c04108c008a4e20/pkg/agent/controller/endpoint.go#L242 
The local endpoint could potentially have a nil node name but if that happens we not going to be able to tell if that endpoint is the one needed so should probably ignore it

coredns/endpointslice/map.go
mikenorgate: I think taking the first one should be fine, if there ever is more than one they should all point to the same set of addresses

coredns/endpointslice/map.go
tpantelis: The `NodeName` pointer for `endpoint` is assigned from `address` which theoretically could be be nil. We just want to avoid a crash if somehow either is nil so, if we're not certain, then we should check for nil and perhaps log a warning.

mikenorgate: I've added a check to make sure they're not nil

tpantelis (approved):

dfarrell07:
It's weird there isn't a rebase button or a note about conflict on the PR, but it can't be merged because it's "out-of-date with the base branch".

vthapar:
@mikenorgate Can you rebase with latest devel so we can get this merged? Pulling this locally I can see it is missing a rebase.

maayanf24:
@mikenorgate - can you please fix the lint errors?

mikenorgate:
> @mikenorgate - can you please fix the lint errors?

I think it just needs to be re-run, looking at the error it seems to be a networking issue

PR #762, opened 2022-05-16
Move controllers used by plugin to coredns

vthapar:
Following packages in `pkg` are used only by coredns plugin.
 * endpointslice
 * gateway
 * service
 * serviceimport

Move these to `coredns` package as that is more logical package
for them. This will simplify any changes to these as will no longer
need changes to come in two different PRs or require updating
`coredns/go.mod`.


skitt (approved):

coredns/gateway/controller.go
```diff
@@ -46,8 +46,8 @@ type NewClientsetFunc func(c *rest.Config) (dynamic.Interface, error)
 var NewClientset NewClientsetFunc
 
 type Controller struct {
-	NewClientset     NewClientsetFunc
-	informer         cache.Controller
+	NewClientset NewClientsetFunc
+	informer     cache.Controller
```
tpantelis: This looks like an unintended change?

vthapar: Yeah. I'll fix this.

EDIT: Argh! Sorry @tpantelis I ended up doing commit ammend instead of new commit for this.

tpantelis (approved):

PR #797, opened 2022-07-08
Require golang.org/x/text 0.3.7

dfarrell07:

### DIFF
--- a/coredns/go.mod
+++ b/coredns/go.mod
@@ -2,6 +2,8 @@ module github.com/submariner-io/lighthouse/coredns
 
 go 1.16
 
+replace golang.org/x/text => golang.org/x/text v0.3.7
+
 require (
 	github.com/caddyserver/caddy v1.0.5
 	github.com/coredns/caddy v1.1.1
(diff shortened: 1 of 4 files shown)


dfarrell07:
Am I doing this right @skitt?

skitt (changes requested):
I don’t think this is the right way to go about this (and this reminds me I’ve been meaning to do a Tech Exchange session on Go module and package dependencies).

Replacing a module in this way ties the project to a specific version of `x/text`. This means that any future upgrade brought in by another dependency will be ignored, which is at best undesirable and at worst can cause problems (_e.g._ if said dependency relies on a new feature added in a later version of `x/text`).

In practice, the `coredns` module already relies on `x/text` version 0.3.7 or later; the simplest way to see this using our build infrastructure is to run `make licensecheck` — that lists all the modules which are effectively included in the binaries we build.

Of course it’s desirable to express the fact that we want to have `x/text` version 0.3.7 or later. Go 1.16 doesn’t give us any nice way of doing so; upgrading the module to require Go 1.17 would, except that

```
github.com/submariner-io/lighthouse/coredns imports
	github.com/coredns/coredns/plugin/k8s_external tested by
	github.com/coredns/coredns/plugin/k8s_external.test imports
	github.com/coredns/coredns/plugin/kubernetes imports
	k8s.io/client-go/plugin/pkg/client/auth/gcp imports
	golang.org/x/oauth2/google imports
	cloud.google.com/go/compute/metadata: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
	cloud.google.com/go v0.81.0 (/home/skitt/go/pkg/mod/cloud.google.com/go@v0.81.0/compute/metadata)
	cloud.google.com/go/compute v1.3.0 (/home/skitt/go/pkg/mod/cloud.google.com/go/compute@v1.3.0/metadata)
```

and I haven’t figured out how to fix that yet.

skitt:
See https://github.com/submariner-io/lighthouse/pull/799 for a different approach.

skitt:
However, we do need something like this on the 0.12 branch, and we can’t backport a fix from `devel` since the project structure has changed.

PR #828, opened 2022-07-20
Use minimal KIND clusters

mkolesnik:
Seems that we don't really need such large clusters to run E2E, as
Lighthouse doesn't care about specific nodes.

Using minimal cluster speeds up the deployment and tests, and frees up
system resources.

### DIFF
--- a/.shipyard.e2e.yml
+++ b/.shipyard.e2e.yml
@@ -1,9 +1,8 @@
 ---
 cni: weave
 submariner: true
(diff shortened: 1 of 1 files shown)


skitt (approved):

tpantelis (approved):

dfarrell07:
As discussed on the dev call, we should backport this to maybe-help timeout issues on release branches.

PR #860, opened 2022-08-23
If no service matches a DNS request, return empty

### LINKED ISSUE #859: query a not exist record too slow
** if we query a service name like nginx.default.svc.clusterset.local associate a service export, we can quickly get the result. but if a service name ends with svc.clusterset.local which not associate a service export, we will cost lots of time.**:

**query a service that not export ,we will get a empty response or NXDomain  quickly**:

attach into a pod ,then nslookup xxx.default.svc.clusterset.local .if we tcpdump the traffic between pod and coredns,we will see the log below:

```log
13:53:40.532548 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 148: 10.20.144.27.12626 > 10.243.38.189.53: 4345+ A? xxx.default.svc.clusterset.local. (106)
13:53:42.532924 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 59: 10.20.144.27.51308 > 10.243.38.189.53: 47653+ NS? . (17)
13:53:42.532925 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 148: 10.20.144.27.1798 > 10.243.38.189.53: 4345+ A? xxx.default.svc.clusterset.local. (106)
13:53:42.533094 c2:0b:32:92:ba:c2 > ee:ee:ee:ee:ee:ee, ethertype IPv4 (0x0800), length 59: 10.243.38.189.53 > 10.20.144.27.51308: 47653 Refused- 0/0/0 (17)
13:53:47.530687 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 148: 10.20.144.27.54031 > 10.243.38.189.53: 4345+ A? xxx.default.svc.clusterset.local. (106)
13:53:47.530713 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 59: 10.20.144.27.62748 > 10.243.38.189.53: 42075+ NS? . (17)
13:53:47.530882 c2:0b:32:92:ba:c2 > ee:ee:ee:ee:ee:ee, ethertype IPv4 (0x0800), length 59: 10.243.38.189.53 > 10.20.144.27.62748: 42075 Refused- 0/0/0 (17)
13:53:51.531340 ee:ee:ee:ee:ee:ee > c2:0b:32:92:ba:c2, ethertype IPv4 (0x0800), length 59: 10.20.144.27.53267 > 10.243.38.189.53: 15764+ NS? . (17)
13:53:51.531520 c2:0b:32:92:ba:c2 > ee:ee:ee:ee:ee:ee, ethertype IPv4 (0x0800), length 59: 10.243.38.189.53 > 10.20.144.27.53267: 15764 Refused- 0/0/0 (17)
```

the total time will cost about ten seconds.

**Environment**: submariber 0.12.1
- Install tools:
- Others:

blue-troy:
Write the dns message response body when it is not a ServerFailure

Coredns assumes ServeDNS has written the response body unless
the returned code indicates it hasn't been written as determined
by ClientWrite (which is only the codes RcodeServerFailure,
RcodeRefused, RcodeFormatError and RcodeNotImplemented),
in which case it writes the response.

If there's no next plugin, it simply returns the code but if
it isn't one of the ClientWrite codes, then no response is written.

If no response,client will try to resend dns request multiple times,
which causes seconds delay.

Fixes #859
Signed-off-by: blue-troy <12729455+blue-troy@users.noreply.github.com>

### DIFF
--- a/coredns/plugin/handler.go
+++ b/coredns/plugin/handler.go
@@ -44,14 +44,14 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 	zone := plugin.Zones(lh.Zones).Matches(qname)
 	if zone == "" {
 		log.Debugf("Request does not match configured zones %v", lh.Zones)
-		return lh.nextOrFailure(ctx, state.Name(), w, r, dns.RcodeNotZone, "No matching zone found")
+		return lh.nextOrFailure(ctx, state, r, dns.RcodeNotZone)
 	}
 
 	if state.QType() != dns.TypeA && state.QType() != dns.TypeAAAA && state.QType() != dns.TypeSRV {
 		msg := fmt.Sprintf("Query of type %d is not supported", state.QType())
 		log.Debugf(msg)
 
-		return lh.nextOrFailure(ctx, state.Name(), w, r, dns.RcodeNotImplemented, msg)
+		return lh.nextOrFailure(ctx, state, r, dns.RcodeNotImplemented)
 	}
 
 	zone = qname[len(qname)-len(zone):] // maintain case of original query
@@ -61,7 +61,7 @@ func (lh *Lighthouse) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns
 	if pErr != nil || pReq.podOrSvc != Svc {
 		// We only support svc type queries i.e. *.svc.*
 		log.Debugf("Request type %q is not a 'svc' type query - err was %v", pReq.podOrSvc, pErr)
-		return lh.nextOrFailure(ctx, state.Name(), w, r, dns.RcodeNameError, "Only services supported")
+		return lh.nextOrFailure(ctx, state, r, dns.RcodeNameError)
 	}
 
 	return lh.getDNSRecord(ctx, zone, state, w, r, pReq)
@@ -83,7 +83,7 @@ func (lh *Lighthouse) getDNSRecord(ctx context.Context, zone string, state *requ
 			pReq.service, lh.ClusterStatus.IsConnected)
 		if !found {
 			log.Debugf("No record found for %q", state.QName())
-			return lh.nextOrFailure(ctx, state.Name(), w, r, dns.RcodeNameError, "record not found")
+			return lh.nextOrFailure(ctx, state, r, dns.RcodeNameError)
 		}
 
 		isHeadless = true
@@ -141,11 +141,15 @@ func (lh *Lighthouse) getDNSRecord(ctx context.Context, zone string, state *requ
 func (lh *Lighthouse) emptyResponse(state *request.Request) (int, error) {
 	a := new(dns.Msg)
 	a.SetReply(state.Req)
+
+	return lh.writeResponse(state, a)
+}
+
+func (lh *Lighthouse) writeResponse(state *request.Request, a *dns.Msg) (int, error) {
 	a.Authoritative = true
 
 	wErr := state.W.WriteMsg(a)
 	if wErr != nil {
-		// Error writing reply msg
 		log.Errorf("Failed to write message %#v: %v", a, wErr)
 		return dns.RcodeServerFailure, lh.error("failed to write response")
 	}
@@ -162,10 +166,13 @@ func (lh *Lighthouse) error(str string) error {
 	return plugin.Error(lh.Name(), errors.New(str)) // nolint:wrapcheck // Let the caller wrap it.
 }
 
-func (lh *Lighthouse) nextOrFailure(ctx context.Context, name string, w dns.ResponseWriter, r *dns.Msg, code int, err string) (int, error) {
-	if lh.Fall.Through(name) {
-		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r) // nolint:wrapcheck // Let the caller wrap it.
+func (lh *Lighthouse) nextOrFailure(ctx context.Context, state *request.Request, r *dns.Msg, rcode int) (int, error) {
+	if lh.Fall.Through(state.Name()) {
+		return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, state.W, r) // nolint:wrapcheck // Let the caller wrap it.
 	}
 
-	return code, lh.error(err)
+	a := new(dns.Msg)
+	a.SetRcode(r, rcode)
+
+	return lh.writeResponse(state, a)
 }
--- a/coredns/plugin/handler_test.go
+++ b/coredns/plugin/handler_test.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/dnstest"
 	"github.com/coredns/coredns/plugin/pkg/fall"
 	"github.com/coredns/coredns/plugin/test"
@@ -328,7 +329,12 @@ func testWithFallback() {
 		t.mockCs.localClusterID = clusterID
 		t.mockEs.endpointStatusMap[clusterID] = true
 		t.lh.Fall = fall.F{Zones: []string{"clusterset.local."}}
-		t.lh.Next = test.NextHandler(dns.RcodeBadCookie, errors.New("dummy plugin"))
+		t.lh.Next = test.HandlerFunc(func(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
+			m := new(dns.Msg)
+			m.SetRcode(r, dns.RcodeBadCookie)
+			_ = w.WriteMsg(m)
+			return dns.RcodeBadCookie, nil
+		})
 
 		rec = dnstest.NewRecorder(&test.ResponseWriter{})
 	})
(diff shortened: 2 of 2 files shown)


skitt (changes requested):
In your commit message, please describe the change you’re making, not just what it’s supposed to fix. Something like

```
If no service matches a DNS request, return empty

Chaining to the next server results in long delays before returning failure.
Since we know we manage all of .clusterset.local, return empty instead of chaining when no matching service is found.
```

(You’d have to check whether the “we know …” assumption is correct.)

blue-troy:
> 

thank for you remind,i will edit it

blue-troy:
in [coredns code](https://github.com/coredns/coredns/blob/master/plugin/kubernetes/handler.go#L65) , it handle NameError like below it return the RcodeSuccess and write code RcodeNameError to the Msg


```go
	if k.IsNameError(err) {
		if k.Fall.Through(state.Name()) {
			return plugin.NextOrFailure(k.Name(), k.Next, ctx, w, r)
		}
		if !k.APIConn.HasSynced() {
			// If we haven't synchronized with the kubernetes cluster, return server failure
			return plugin.BackendError(ctx, &k, zone, dns.RcodeServerFailure, state, nil /* err */, plugin.Options{})
		}
		return plugin.BackendError(ctx, &k, zone, dns.RcodeNameError, state, nil /* err */, plugin.Options{})
	}

func BackendError(ctx context.Context, b ServiceBackend, zone string, rcode int, state request.Request, err error, opt Options) (int, error) {
	m := new(dns.Msg)
	m.SetRcode(state.Req, rcode)
	m.Authoritative = true
	m.Ns, _ = SOA(ctx, b, zone, state, opt)

	state.W.WriteMsg(m)
	// Return success as the rcode to signal we have written to the client.
	return dns.RcodeSuccess, err
}
```

tpantelis:
The [second commit](https://github.com/submariner-io/lighthouse/pull/860/commits/0ca3deaf37dfa636f4ee4e09fc784befbfcb926c) is missing the `Signed-off-by`. Also the second overwrites the first so you might as well squash the commits.

Also these changes introduce [unit test failures](https://github.com/submariner-io/lighthouse/runs/7986090476?check_suite_focus=true).

blue-troy:
> The [second commit](https://github.com/submariner-io/lighthouse/pull/860/commits/0ca3deaf37dfa636f4ee4e09fc784befbfcb926c) is missing the `Signed-off-by`. Also the second overwrites the first so you might as well squash the commits.
> 
> Also these changes introduce [unit test failures](https://github.com/submariner-io/lighthouse/runs/7986090476?check_suite_focus=true).

thanks，[our unit test test the ServeDNS return code](https://github.com/submariner-io/lighthouse/blob/devel/coredns/plugin/handler_test.go#L1009) :

```go
	code, err := t.lh.ServeDNS(context.TODO(), rec, tc.Msg())

	Expect(code).Should(Equal(tc.Rcode))

	if tc.Rcode == dns.RcodeSuccess {
		Expect(err).To(Succeed())
		Expect(test.SortAndCheck(rec.Msg, tc)).To(Succeed())
	} else {
		Expect(err).To(HaveOccurred())
	}
```

but the Msg.code is more important , maybe we shoud check mgs.code and  ServeDNS's return error like below which coredns [kubernetes](https://github.com/coredns/coredns/blob/master/plugin/kubernetes/handler_case_test.go#L62) plugin do ? : 

```go
	_, err := k.ServeDNS(ctx, w, r)
		if err != tc.Error {
			t.Errorf("Test %d expected no error, got %v", i, err)
			return
		}
		if tc.Error != nil {
			continue
		}

		resp := w.Msg
		if resp == nil {
			t.Fatalf("Test %d, got nil message and no error for %q", i, r.Question[0].Name)
		}

		if err := test.SortAndCheck(resp, tc); err != nil {
			t.Error(err)
		}
```

blue-troy:
> our unit test test the ServeDNS return code

or we can simply return the dns.RcodeSuccess with empty ？

coredns/plugin/handler.go
```diff
@@ -83,7 +83,10 @@ func (lh *Lighthouse) getDNSRecord(ctx context.Context, zone string, state *requ
 			pReq.service, lh.ClusterStatus.IsConnected)
 		if !found {
 			log.Debugf("No record found for %q", state.QName())
-			return lh.nextOrFailure(ctx, state.Name(), w, r, dns.RcodeNameError, "record not found")
+			if lh.Fall.Through(state.Name()) {
+				return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r) // nolint:wrapcheck // Let the caller wrap it.
+			}
+			return lh.emptyResponse(state, dns.RcodeNameError)
```
tpantelis: In looking at the [coredns server code](https://github.com/coredns/coredns/blob/fc7f3835d506960364b21cf62116a70eb8f4caf2/core/dnsserver/server.go#L263):

```Go
rcode, _ := h.pluginChain.ServeDNS(ctx, w, r)
if !plugin.ClientWrite(rcode) {
    errorFunc(s.Addr, w, r, rcode)
}
return
```

it assumes `ServeDNS` has written the response body unless the returned code indicates it hasn't been written as determined by `ClientWrite` (which is only the codes `RcodeServerFailure`, `RcodeRefused`, `RcodeFormatError` and `RcodeNotImplemented`), in which case it writes the response. 

So  in `lh.nextOrFailure`:

```Go
func (lh *Lighthouse) nextOrFailure(ctx context.Context, name string, w dns.ResponseWriter, r *dns.Msg, code int, err string) (int, error) {
    if lh.Fall.Through(name) {
        return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r) // nolint:wrapcheck // Let the caller wrap it.
    }

    return code, lh.error(err)
}
```

if there's no next plugin, it simply returns the code but if it isn't one of the `ClientWrite` codes, then no response is written, which is the case here with `RcodeNameError`. 

Rather than using `emptyResponse` here, I think we should modify `nextOrFailure` to return a response similarly. To reduce code duplication, we can define a general function:

```Go
func (lh *Lighthouse) writeResponse(state *request.Request, a *dns.Msg) (int, error) {
    a.Authoritative = true

    wErr := state.W.WriteMsg(a)
    if wErr != nil {
        log.Errorf("Failed to write message %#v: %v", a, wErr)
	return dns.RcodeServerFailure, lh.error("failed to write response")
    }

    return dns.RcodeSuccess, nil
}
```

Then `emptyResponse` becomes:

```Go
func (lh *Lighthouse) emptyResponse(state *request.Request) (int, error) {
    a := new(dns.Msg)
    a.SetReply(state.Req)

    return lh.writeResponse(state, a)
}
```

And `nextOrFailure` becomes:

```Go
func (lh *Lighthouse) nextOrFailure(ctx context.Context, name string, w dns.ResponseWriter, r *dns.Msg, code int) (int, error) {
    if lh.Fall.Through(name) {
        return plugin.NextOrFailure(lh.Name(), lh.Next, ctx, w, r) // nolint:wrapcheck // Let the caller wrap it.
    }

    a := new(dns.Msg)
    a.SetRcode(state.Req, rcode)

    return lh.writeResponse(state, a)
}
```

tpantelis (changes requested):

coredns/plugin/handler.go
blue-troy: good idea

blue-troy: [in coredns code ](https://github.com/coredns/coredns/blob/master/plugin/plugin.go#L20) 
```go
	// Handler is like dns.Handler except ServeDNS may return an rcode
	// and/or error.
	//
	// If ServeDNS writes to the response body, it should return a status
	// code. CoreDNS assumes *no* reply has yet been written if the status
	// code is one of the following:
	//
	// * SERVFAIL (dns.RcodeServerFailure)
	//
	// * REFUSED (dns.RecodeRefused)
	//
	// * FORMERR (dns.RcodeFormatError)
	//
	// * NOTIMP (dns.RcodeNotImplemented)
```

so only this 4 code can not use writeResponse method

tpantelis: No it doesn't mean we can't handle those codes - it just means we don't have to. But we might as well to simplify our code.

tpantelis (approved):

tpantelis:
@vthapar @aswinsuryan Can you review?

coredns/plugin/handler.go
```diff
@@ -162,10 +166,13 @@ func (lh *Lighthouse) error(str string) error {
 	return plugin.Error(lh.Name(), errors.New(str)) // nolint:wrapcheck // Let the caller wrap it.
 }
 
-func (lh *Lighthouse) nextOrFailure(ctx context.Context, name string, w dns.ResponseWriter, r *dns.Msg, code int, err string) (int, error) {
+func (lh *Lighthouse) nextOrFailure(ctx context.Context, name string, state *request.Request, r *dns.Msg, rcode int) (int, error) {
```
tpantelis: Since callers always pass `state.Name()`, we can eliminate the `name` parameter since we have the `state` instance.

tpantelis (changes requested):

coredns/plugin/handler.go
blue-troy: ok, it is solved

tpantelis (approved):

dfarrell07:
@aswinsuryan to verify if this is actually user-facing, needs release note. - PR scrub

skitt (approved):

aswinsuryan (approved):

aswinsuryan:
> @aswinsuryan to verify if this is actually user-facing, needs release note. - PR scrub

user should see improvements when querying a record not available in "clusterset.local" , so can be added in release note.

PR #879, opened 2022-09-20
Handle update of an exported service

### LINKED ISSUE #645: Updating an exported service is not handled
**What happened**:
In a working setup, when a service is exported, we can successfully access it from remote clusters.
If we update the service (like the port numbers) in the origin cluster, the corresponding serviceImport is not updated.

blue-troy:
When an exported service update, Service deletion
will trigger ServiceExport update.

Fixes #645
Signed-off-by: blue-troy <12729455+blue-troy@users.noreply.github.com>


pkg/agent/controller/agent.go
```diff
@@ -381,6 +385,20 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, numRequeue
...
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, true
+		}
+		// some other error. Log and requeue
+		klog.Errorf("Error delete ServiceImport for Service (%s/%s): %v", svc.Namespace, svc.Name, err)
+
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, corev1.ConditionFalse, serviceUpdate,
+			"Service update")
+
+		return nil, false
```
tpantelis: What specific modification of an exported service would warrant this action? It seems potentially disruptive to delete the `ServiceImport` - why not update the existing one?

tpantelis (changes requested):

pkg/agent/controller/agent.go
blue-troy: The service may change to an unsupported type service. service Import seem don't have a not avalible status

tpantelis: Then let's handle that specific case rather than blindly deleting the `ServiceImport`.

blue-troy: Good idea, i will try to solve it in the next few days

blue-troy: update `ServiceExport` status in Service deletion, let `ServiceExport -> ServiceImport` decide how to deal with it.

pkg/agent/controller/agent.go
```diff
@@ -381,6 +391,13 @@ func (a *Controller) serviceToRemoteServiceImport(obj runtime.Object, numRequeue
 	serviceImport := a.newServiceImport(svcExport.Name, svcExport.Namespace)
 
 	// Update the status and requeue
+	if op == syncer.Update {
+		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, corev1.ConditionFalse, serviceUpdate,
+			"Service update")
```
tpantelis: Rather than changing the status condition to an intermediate state indicating it's not valid and relying on a `ServiceExport` update to trigger `serviceExportToServiceImport`, I think it would be cleaner to call `serviceExportToServiceImport`  directly:

`return serviceExportToServiceImport(svcExport, numRequeues, syncer.Update)`

tpantelis (changes requested):

pkg/agent/controller/agent.go
blue-troy: When call `return serviceExportToServiceImport(svcExport, numRequeues, syncer.Update)` directly

[serviceExportToServiceImport code](https://github.com/submariner-io/lighthouse/blob/v0.14.0-m1/pkg/agent/controller/agent.go#L263)

```go
	if op == syncer.Update && getLastExportConditionReason(svcExport) != serviceUnavailable {
		return nil, false
	}
```

LastExportConditionReason "" may be trigger by `onSuccessfulServiceImportSync` or `serviceToRemoteServiceImport` directly call at least. will ConditionReason "" may be somthing else? If it's not,we can change onSuccessfulServiceImportSync reason with `ServiceImportSynced`.

If we do not want to change `onSuccessfulServiceImportSync` reason, change `ServiceExport` then directly call `serviceToRemoteServiceImport` will cause `serviceToRemoteServiceImport` triggered more than it's needs.

tpantelis: Sorry I'm not sure I follow. So if `serviceToRemoteServiceImport` is called directly then the above check will skip updating the `ServiceImport`? If so then perhaps instead of checking the last condition reason we check if the last condition Status is `ConditionTrue`, ie process the update only if the service was previously exported.

blue-troy: you mean [serviceExportToServiceImport](https://github.com/submariner-io/lighthouse/blob/v0.14.0-m1/pkg/agent/controller/agent.go#L263) like this ?：

```go
	if op == syncer.Update && getLastExportCondition(svcExport) != corev1.ConditionTrue {
		return nil, false
	}
```

onSuccessfulServiceImportSync will trigger this ConditionTrue too. how distinguish service update direct call and onSuccessfulServiceImportSync? service update direct call and onSuccessfulServiceImportSync has the same Status and Reason

tpantelis: https://github.com/submariner-io/lighthouse/pull/899 should alleviate your issues with calling `serviceToRemoteServiceImport` directly.

blue-troy: OK, i will have a look

blue-troy: #899 is useful in our case 👍, i will push direct call `serviceToRemoteServiceImport` version code after your  #899 PR merged

blue-troy:
should we update [code](https://github.com/submariner-io/lighthouse/blob/8a8c1fbfc8d34b98c15aaef670d1c3797cdb2fbf/pkg/agent/controller/agent.go#L269-L277) like this ?:

```go
	svcType, ok := getServiceImportType(svc)

	if !ok {
		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid, corev1.ConditionFalse,
			invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
		klog.Errorf("Service type %q not supported", svc.Spec.Type)

		//delete unsupported service type
		err := a.serviceImportSyncer.GetLocalFederator().Delete(serviceImport)
                 ....
}
```

to support case like an exported service type change from a supported type to an unsupported type?

vthapar:
> should we update [code](https://github.com/submariner-io/lighthouse/blob/8a8c1fbfc8d34b98c15aaef670d1c3797cdb2fbf/pkg/agent/controller/agent.go#L269-L277) like this ?:
> 
> ```go
> 	svcType, ok := getServiceImportType(svc)
> 
> 	if !ok {
> 		a.updateExportedServiceStatus(svcExport.Name, svcExport.Namespace, mcsv1a1.ServiceExportValid, corev1.ConditionFalse,
> 			invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
> 		klog.Errorf("Service type %q not supported", svc.Spec.Type)
> 
> 		//delete unsupported service type
> 		err := a.serviceImportSyncer.GetLocalFederator().Delete(serviceImport)
>                  ....
> }
> ```
> 
> to support case like an exported service type change from a supported type to an unsupported type?

It should be done becasue a serviceImport shouldn't exist for a service that isn't supported as per current logic. But would like to add a check for `op.Update`

@tpantelis In case of delete of something that doesn't exist yet, do we silently ignore, or log it?

pkg/agent/controller/agent.go
```diff
@@ -273,7 +273,7 @@ func (a *Controller) serviceExportToServiceImport(obj runtime.Object, numRequeue
 			invalidServiceType, fmt.Sprintf("Service of type %v not supported", svc.Spec.Type))
 		klog.Errorf("Service type %q not supported", svc.Spec.Type)
 
-		return nil, true
+		return nil, false
```
vthapar: Why changing this to false?

vthapar (changes requested):

pkg/agent/controller/agent.go
blue-troy: serviceToRemoteServiceImport can cover `service change from not supported type to supported type` case, by call a.serviceExportToServiceImport.

tpantelis: It should. We should add unit test cases to cover the new changes. I can do that in a follow-up PR.

tpantelis:
> It should be done becasue a serviceImport shouldn't exist for a service that isn't supported as per current logic. But would like to add a check for `op.Update`
> 
> @tpantelis In case of delete of something that doesn't exist yet, do we silently ignore, or log it?

It doesn't: https://github.com/submariner-io/admiral/blob/e07dc24fda439c217cf8c512161474b1038aa355/pkg/federate/base_federator.go#L70. It's up to the caller to check the error.

tpantelis:
OK I pushed 2 additional commits: the first deletes the `ServiceImport` when the type is unsupported and the second adds unit tests cases to cover the changes. I think we should get this in for 0.14. @vthapar @aswinsuryan 

**Edit**: also pushed commit for the release note.

tpantelis (approved):

dfarrell07 (approved):
LGTM, thanks @blue-troy!

skitt (approved):

skitt:
@vthapar this is waiting on your re-review.

vthapar (approved):
Thanks for all the work on this @blue-troy and @tpantelis . Looks good.

PR #893, opened 2022-10-07
Remove helm jobs

tpantelis:
The helm jobs have been failing recently due to mis-matched image versions, ie it's using the latest **submariner-operator** b/c it's specifically pre-loaded for helm but it's using the chart version set in **submariner-charts** for the **submariner-gateway** and **submariner-routeagent** images, which is currently set to _0.13.0-rc2_. So in order to use the latest _devel_ image versions, we'd need to pre-load them all.

Given that and the fact that we don't have helm jobs in the **submariner** and **submariner-operator** repos (or they were removed), it seems inconsistent to test helm in lighthouse so the jobs have been removed.

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        deploytool: ['operator', 'helm']
+        deploytool: ['operator']
         globalnet: ['', 'globalnet']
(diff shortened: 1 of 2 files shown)


sridhargaddam (approved):

dfarrell07:
See Slack for more discussion: https://kubernetes.slack.com/archives/C010RJV694M/p1665065349226939

dfarrell07 (approved):

PR #964, opened 2022-11-17
Include namespace in EndpointSlice name

### LINKED ISSUE #957: Synced endpointslice name in broker namespace missing original namespace name
**What happened**:
We have two services under different namespace with same name. When we export them using lighthouse, it overwrites information from the last updated resource. This conflict affects our ability to run same name services under different namespaces.

Namespace name attached to endpoint slice document name synced to broker namespace.

1. Setup a broker and agent cluster.
2. Setup nginx service in agent cluster under temp namespace using following commands
   1. kubectl create namespace temp
   2. kubectl create deployment nginx -n temp --image=nginx
   3. kubectl expose deployment nginx -n temp --port=80
   4. subctl export service nginx
3. Check synced document for nginx under broker namespace. It should show endpoints from **nginx under temp** namespace.
4. Follow all the steps from step 2 under namespace demo.
5. Check synced document for nginx under broker namespace. It should show endpoints from **nginx under demo** namespace.
7. Scale nginx deployment under temp namespace by executing "kubectl scale deployment nginx -n temp --replicas=2"
8. Check synced document for nginx under broker namespace. It should show endpoints from **nginx under temp** namespace

**Environment**:
- Install tools:
- Others:

tpantelis:
Fixes https://github.com/submariner-io/lighthouse/issues/957

Also remove prior `EndpointSlices` on migration.

Depends on https://github.com/submariner-io/subctl/pull/394

### DIFF
--- a/pkg/agent/controller/reconciliation_test.go
+++ b/pkg/agent/controller/reconciliation_test.go
@@ -21,7 +21,10 @@ package controller_test
 import (
 	. "github.com/onsi/ginkgo"
 	"github.com/submariner-io/admiral/pkg/syncer/test"
+	"github.com/submariner-io/lighthouse/pkg/constants"
 	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/agent/controller/reconciliation_test.go
```diff
@@ -156,4 +159,28 @@ var _ = Describe("Reconciliation", func() {
 			t.awaitNoEndpointSlice(t.cluster2.localEndpointSliceClient)
 		})
 	})
+
+	When("a local EndpointSlice with the old naming convention sans namespace exists on startup", func() {
```
skitt, line 163: How about a test with two slices with the same name in different namespaces?

skitt (changes requested):

pkg/agent/controller/reconciliation_test.go
tpantelis: So a service with the same name in 2 namespaces. I can look into that. But this test is still valid as is.

skitt: Yes, I’m not saying this test is invalid; just asking for an additional test that verifies the behaviour in the bug.

tpantelis: Added with https://github.com/submariner-io/lighthouse/pull/964/commits/3c961200cc28f9b68291cf4ec4e9a5aba642b075

skitt (approved):

vthapar (approved):
Changes look fine. Issue is with making the name even longer and closer to limit, but that needs to be addressed differently by using hash or uuid.

tpantelis:
> Changes look fine. Issue is with making the name even longer and closer to limit, but that needs to be addressed differently by using hash or uuid.

Yes that’s on my TODO list.

PR #981, opened 2022-11-28
Increase the dig timeout to 20s

skitt:
We're seeing e2e failures with dig exiting with code 9, "No reply from server". This increases the dig timeout to reduce the risk of prematurely aborting the test.

### DIFF
--- a/test/e2e/discovery/headless_services.go
+++ b/test/e2e/discovery/headless_services.go
@@ -332,7 +332,7 @@ func verifyHeadlessSRVRecordsWithDig(f *framework.Framework, cluster framework.C
 func createSRVQuery(f *framework.Framework, port *corev1.ServicePort, service *corev1.Service,
 	domain string, clusterName string, withPort, withcluster bool,
 ) (cmd []string, domainName string) {
-	cmd = []string{"dig", "+short", "SRV"}
+	cmd = []string{"dig", "+short", "+timeout=20", "SRV"}
 
 	domainName = service.Name + "." + f.Namespace + ".svc." + domain
 	clusterDNSName := domainName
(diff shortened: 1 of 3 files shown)


dfarrell07 (approved):

test/e2e/discovery/headless_services.go
```diff
@@ -332,7 +332,7 @@ func verifyHeadlessSRVRecordsWithDig(f *framework.Framework, cluster framework.C
 func createSRVQuery(f *framework.Framework, port *corev1.ServicePort, service *corev1.Service,
 	domain string, clusterName string, withPort, withcluster bool,
 ) (cmd []string, domainName string) {
-	cmd = []string{"dig", "+short", "SRV"}
+	cmd = []string{"dig", "+short", "+timeout=20", "SRV"}
```
tpantelis, line 335: Perhaps we should  make this a constant in _framework.go_ , ie `DIG_TIMEOUT = "+timeout=20"`?

test/e2e/framework/framework.go
```diff
@@ -570,7 +570,7 @@ func (f *Framework) VerifyServiceIPWithDig(srcCluster, targetCluster framework.C
 func (f *Framework) VerifyIPWithDig(srcCluster framework.ClusterIndex, service *v1.Service, targetPod *v1.PodList,
 	domains []string, clusterName, serviceIP string, shouldContain bool,
 ) {
-	cmd := []string{"dig", "+short"}
+	cmd := []string{"dig", "+short", "+timeout=20"}
```
vthapar, line 573: Agree with @tpantelis Would be good to use a const.

vthapar:
https://github.com/submariner-io/lighthouse/pull/984 Should help with `code 9` error in this particular case. It is testing a reply without error but coredns doesn't send back response to us and we timeout. All this while, coredns is bombarding lighthouse-coredns with queries. #984 will allow coredns to return error to dig and reduce pressure on lighthouse-coredns.

nyechiel:
> #984 Should help with `code 9` error in this particular case. It is testing a reply without error but coredns doesn't send back response to us and we timeout. All this while, coredns is bombarding lighthouse-coredns with queries. #984 will allow coredns to return error to dig and reduce pressure on lighthouse-coredns.

Do we still want to get this PR reviewed and merged at this point?

PR #1038, opened 2023-01-09
Remove legacy label checks for Endpointslices

tpantelis:
The proprietary `LighthouseLabelSourceCluster` et al labels were replaced with the MSC-standard labels in 0.12 but some checks were left in for migration.

### DIFF
--- a/coredns/endpointslice/map.go
+++ b/coredns/endpointslice/map.go
@@ -112,11 +112,6 @@ func (m *Map) Put(es *discovery.EndpointSlice) {
 
 	cluster, ok := es.Labels[constants.MCSLabelSourceCluster]
 
-	// Remove this after 0.12 (this handles old map entries with pre-MCS labels)
-	if !ok {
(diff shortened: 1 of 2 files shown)


skitt:
I wonder if https://github.com/submariner-io/lighthouse/blob/devel/coredns/serviceimport/map.go#L154 doesn’t need fixing before we can do this properly...

tpantelis:
> I wonder if https://github.com/submariner-io/lighthouse/blob/devel/coredns/serviceimport/map.go#L154 doesn’t need fixing before we can do this properly...

yeah that's on my list to convert `ServiceImport` to use the MCS labels as well but that separate from `Endpointslices`. Unless you were referring to something else?

skitt (approved):

skitt:
> yeah that's on my list to convert `ServiceImport` to use the MCS labels as well but that separate from `Endpointslices`. Unless you were referring to something else?

I was looking at all the uses of the constants, and didn’t pay attention to the objects that were being manipulated.

aswinsuryan (approved):

PR #1041, opened 2023-01-12
Bump mellium.im/sasl to fix CVE-2022-48195

dfarrell07:
I don't think we're actually impacted by this as we're not connecting to untrusted endpoints, but we shouldn't need to trust endpoints.

Decription of CVE:

An issue was discovered in Mellium mellium.im/sasl before 0.3.1. When performing SCRAM-based SASL authentication, if the remote end advertises support for channel binding, no random nonce is generated (instead, the nonce is empty). This causes authentication to fail in the best case, but (if paired with a remote end that does not validate the length of the nonce) could lead to insufficient randomness being used during authentication.

https://mellium.im/cve/cve-2022-48195/

Without the replace statement, running `go mod tidy` undoes this change.

### DIFF
--- a/coredns/go.mod
+++ b/coredns/go.mod
@@ -21,6 +21,10 @@ require (
 // Until coredns is updated for client-go 0.26.0
 replace k8s.io/client-go => k8s.io/client-go v0.25.4
 
+// Until dependencies update to at least v0.3.1 to fix CVE-2022-48195
+// To test if this is needed: remove the replace, run `go mod tidy`, and see if v0.3.1 remains
+replace mellium.im/sasl => mellium.im/sasl v0.3.1
+
 require (
 	cloud.google.com/go/compute v1.7.0 // indirect
 	github.com/DataDog/datadog-agent/pkg/obfuscate v0.0.0-20211129110424-6491aa3bf583 // indirect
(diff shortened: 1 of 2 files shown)


dfarrell07:
> Summarizing 1 Failure:
  [FAIL] EndpointSlice controller when a service has a valid endpoint when IsHealthy is called for the service with a valid cluster [It] should return true
  /go/src/github.com/submariner-io/lighthouse/coredns/endpointslice/controller_test.go:174

Not sure what's going on with the UTs.

dfarrell07:
This issue also exists for Submariner 0.14, but not older releases.

dfarrell07:
Three runs of the UTs and the same failure each time. Any idea @vthapar?

tpantelis:
> Three runs of the UTs and the same failure each time. Any idea @vthapar?

I can repro locally.  It looks like an intermittent timing issue related to fake clients - it goes away if I add a small delay in the test.

dfarrell07:
Two weird things about this:
* The mellium.im/sasl dep is only in go.sum, so I don't think it should actually be used. 
* The CVE was also flagged for the main lighthouse mod, which has no reference to mellium.im/sasl.

tpantelis:
> I can repro locally. It looks like an intermittent timing issue related to fake clients - it goes away if I add a small delay in the test.

Fixed by https://github.com/submariner-io/lighthouse/pull/1042.

skitt (changes requested):
This is a false positive, nothing in our code or in any of our dependencies’ packages we actually use, uses `mellium.im/sasl` (for proof of that, `make licensecheck` shows all the modules providing packages that end up in our binaries). [`go.sum`](https://go.dev/ref/mod#go-sum-files) is only a repository of digests, representing the full module dependency tree, and doesn’t reflect the modules which are really used; those are listed exhaustively in `go.mod`.

PR #1044, opened 2023-01-13
Consistently use the coredns logger in the plugin

tpantelis:
Modified the plugin helper modules (_endpointslice_, _serviceimport_ et al) to use the `coredns` logger (indirectly via the `logr` interface) instead of `klog`.


skitt (approved):
Perhaps we should enforce this (golangci-lint can be given a list of forbidden imports).

sridhargaddam (approved):

dfarrell07 (approved):

PR #1051, opened 2023-01-17
Use MCS source name/cluster labels for ServiceImport

tpantelis:
...`multicluster.kubernetes.io/service-name` and `multicluster.kubernetes.io/source-cluster` instead of `lighthouse.submariner.io/sourceName` and `lighthouse.submariner.io/sourceCluster`.

For migration, when the agent controller restarts after upgrade, it will automatically re-process all `ServiceExports` from the informer and update the labels on the corresponding `ServiceImports` and propagate them to the broker et al (I added a UT to cover this). There is a window where the `ServiceImport` broker syncer could observe the un-migrated version first in which case elides syncing it. Reconciliation on startup also looks for the legacy source name label.

The `coredns` module only needs to look for the new labels.


skitt (approved):

tpantelis:
@skitt I made some updates with https://github.com/submariner-io/lighthouse/pull/1051/commits/dae761278da20d4fb164fea55761de63ae338e50.

skitt (approved):

aswinsuryan (approved):

PR #1052, opened 2023-01-18
Resolve service to serviceimport IP

### LINKED ISSUE #993: Always return globalIP for an exported clusterIP service when using clusterset.local domain name
**What happened**:

In the initial implementation of Globalnet (i.e., v1) the globalIPs associated with exported `clusterIPServices` were not routable within the local cluster. So, to support the use-case, when a **client from the local cluster** makes a DNS query to `<svc-name>.<namespace>.svc.clusterset.local` domain, we are currently returning the `clusterIP` instead of the `globalIP` associated with the service. 

With the newer implementation of Globalnet (i.e.,v2), for every exported `clusterIPService`, we can support connectivity to the `globalIP` even from the same cluster as well. 

When a `clusterIPServices` is exported and a request is made to the `<svc-name>.<namespace>.svc.clusterset.local` domain from the local cluster, Lighthouse should return the `globalIP` associated with the exported service instead of returning the `clusterIP`.

yboaron:
Always resolve service to the IP appears in serviceimport resource.

Fixes: https://github.com/submariner-io/lighthouse/issues/993
Depend On: https://github.com/submariner-io/lighthouse/pull/1068

Signed-off-by: Yossi Boaron <yboaron@redhat.com>

### DIFF
--- a/coredns/plugin/record.go
+++ b/coredns/plugin/record.go
@@ -99,13 +99,8 @@ func (lh *Lighthouse) createSRVRecords(dnsrecords []serviceimport.DNSRecord, sta
 func (lh *Lighthouse) getClusterIPForSvc(pReq *recordRequest) (*serviceimport.DNSRecord, bool) {
 	localClusterID := lh.ClusterStatus.LocalClusterID()
 
-	record, found, isLocal := lh.ServiceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, localClusterID, lh.ClusterStatus.IsConnected,
+	record, found, _ := lh.ServiceImports.GetIP(pReq.namespace, pReq.service, pReq.cluster, localClusterID, lh.ClusterStatus.IsConnected,
 		lh.EndpointsStatus.IsHealthy)
-
-	getLocal := isLocal || pReq.cluster != "" && pReq.cluster == localClusterID
-	if found && getLocal {
-		record, found = lh.LocalServices.GetIP(pReq.service, pReq.namespace)
-	}
 
 	return record, found
 }
--- a/test/e2e/discovery/service_discovery.go
+++ b/test/e2e/discovery/service_discovery.go
@@ -189,8 +189,10 @@ func RunServiceDiscoveryLocalTest(f *lhframework.Framework) {
 	netshootPodList := f.NewNetShootDeployment(framework.ClusterA)
 	clusterADomain := getClusterDomain(f.Framework, framework.ClusterA, netshootPodList)
 
-	f.VerifyServiceIPWithDig(framework.ClusterA, framework.ClusterA, nginxServiceClusterA, netshootPodList,
-		[]string{clusterADomain}, "", true)
+	if !framework.TestContext.GlobalnetEnabled {
+		f.VerifyServiceIPWithDig(framework.ClusterA, framework.ClusterA, nginxServiceClusterA, netshootPodList,
+			[]string{clusterADomain}, "", true)
+	}
 
 	f.DeleteService(framework.ClusterA, nginxServiceClusterA.Name)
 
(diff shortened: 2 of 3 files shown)


yboaron:
cc @aswinsuryan @vthapar @sridhargaddam

vthapar:
You'll need to update the UTs and E2E as well.

yboaron:
> You'll need to update the UTs and E2E as well.

@vthapar 
Do you think that the proposed change is OK? 

Because I can see that tests related to headless service and DNS SRV records failed

yboaron:
Updated the PR , 
thanks @aswinsuryan for your help!

yboaron:
cc @aswinsuryan

coredns/plugin/record.go
```diff
@@ -102,10 +102,20 @@ func (lh *Lighthouse) getClusterIPForSvc(pReq *recordRequest) (*serviceimport.DN
...
 	getLocal := isLocal || pReq.cluster != "" && pReq.cluster == localClusterID
 	if found && getLocal {
 		record, found = lh.LocalServices.GetIP(pReq.service, pReq.namespace)
 	}
 
+	// for local request with empty cluster in request prefer the IP retrieved from serviceimport
+	if getLocal && pReq.cluster == "" && serviceImportIP != "" && record != nil {
+		record.IP = serviceImportIP
+	}
+
```
tpantelis: I"m not sure I follow the logic with the changes. But I think the `LocalServices` code was added for globalnet b/c the global IPs weren't locally routable, in which case we can remove it and none of L105-118 should be needed.  @vthapar @aswinsuryan

yboaron: > I"m not sure I follow the logic with the changes. But I think the `LocalServices` code was added for globalnet b/c the global IPs weren't locally routable, in which case we can remove it and none of L105-118 should be needed. @vthapar @aswinsuryan

Also thought like you at first, and that was also the first version of this PR, but all SRV and HEADLESS SRV tests failed.
Let's see what @aswinsuryan @vthapar think

yboaron:
I can confirm with current version of the PR the local service name resolved to GN IP and not to service clusterIP , check [1]



`$ oc --kubeconfig output/kubeconfigs/kind-config-cluster2 get serviceimport -n submariner-operator nginx-default-cluster2 -o yaml 
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceImport
metadata:
  annotations:
    cluster-ip: 242.1.255.253
    origin-name: nginx
    origin-namespace: default
  creationTimestamp: "2023-01-24T16:51:15Z"
  generation: 1
  labels:
    lighthouse.submariner.io/sourceCluster: cluster2
    lighthouse.submariner.io/sourceName: nginx
    lighthouse.submariner.io/sourceNamespace: default
  name: nginx-default-cluster2
  namespace: submariner-operator
  resourceVersion: "244223"
  uid: f3fe3f53-7d2f-4297-91b2-b81b442ae1a6
spec:
  ips:
  - 242.1.255.253
  ports:
  - port: 80
    protocol: TCP
  sessionAffinityConfig: {}
  type: ClusterSetIP
$ oc --kubeconfig output/kubeconfigs/kind-config-cluster2 get svc nginx 
NAME    TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
nginx   ClusterIP   100.67.114.77   <none>        80/TCP    2d19h
$ kubectl --kubeconfig output/kubeconfigs/kind-config-cluster2 -n default run tmp-shell --rm -i --tty --image quay.io/submariner/nettest -- /bin/bash 
If you don't see a command prompt, try pressing enter.
bash-5.0# dig nginx.default.svc.clusterset.local

; <<>> DiG 9.16.6 <<>> nginx.default.svc.clusterset.local
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 799
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: 226b52884c39ac43 (echoed)
;; QUESTION SECTION:
;nginx.default.svc.clusterset.local. IN A

;; ANSWER SECTION:
nginx.default.svc.clusterset.local. 5 IN A      242.1.255.253

;; Query time: 0 msec
;; SERVER: 100.67.0.10#53(100.67.0.10)
;; WHEN: Thu Jan 26 10:14:37 UTC 2023
;; MSG SIZE  rcvd: 125

bash-5.0# `

coredns/plugin/record.go
tpantelis: Hmm...we should investigate why the SRV tests fail in that case - perhaps there's assumptions/checks in the tests that need to be adjusted.  Maybe it's just me but I find L105-118 a bit confusing and hard to follow - I think it would be ideal if we could at least simplify this section if not remove it altogether.

aswinsuryan: It could be a test issue, if e2e is not failing and the use case is working when we removed those lines.  

if not, we do retrieve the port information as well in GetIP method and it could be a case where we depend on this behavior.

tpantelis: I tried it w/o the local service lookup and all the E2E tests pass but the headless service plugin handler tests fail. This is an issue with the test setup - it turns out they're  not actually testing with a headless service b/c the general test setup adds a non-headless service first. I pushed [this PR](https://github.com/submariner-io/lighthouse/pull/1068) to fix it. With that we can remove the `LocalServices` functionality.

yboaron: Ack, 10x

tpantelis (approved):

tpantelis:
We can remove `LocalServices` entirely but I can do that later.  @aswinsuryan

aswinsuryan (approved):

PR #1086, opened 2023-02-06
Properly disable gateway controller in SD-only install

### LINKED ISSUE #936: service discovery not working without gateways
**What happened**:

I deployed solely the lighthouse agents (no gateways) for service discovery by : 

`subctl deploy-broker --components service-discovery`

I followed the user-guide and deployed the nginx-test pods.
https://submariner.io/operations/usage/

The following connectivity test results in an `unknown host exception` when the deployment on the local cluster is scaled to zero: 

`curl nginx.nginx-test.svc.clusterset.local:8080`.

Since the local cluster is not responding, I expect a response from the remote cluster.

Enable debug logging on the lighthouse coredns pods for more specific logging.

```
[DEBUG] plugin/lighthouse: Request received for "nginx.nginx-test.svc.clusterset.local."
[DEBUG] plugin/lighthouse: Couldn't find a connected cluster or valid IPs for "nginx.nginx-test.svc.clusterset.local."
```

Just perform the userguide and scale down the deployment on of of the clusters to zero. Perform the curl command on the cluster with the scaled down deployment. Only install lighthouse agents and **no gateways!**

To get it working, apply this workaround on all clusters:

`kubectl delete crd gateways.submariner.io`

tpantelis:
Since the `gateways` CRD is always installed we need to also check for the existence of the `Submariner` resource to determine if the connectivity component is installed.

Fixes https://github.com/submariner-io/lighthouse/issues/936
Depends on https://github.com/submariner-io/subctl/pull/548


coredns/gateway/controller.go
```diff
@@ -263,17 +264,36 @@ func (c *Controller) getClusterStatusMap() map[string]bool {
 	return c.clusterStatusMap.Load().(map[string]bool)
 }
 
-func (c *Controller) getCheckedClientset(kubeConfig *rest.Config) (dynamic.ResourceInterface, error) {
+func (c *Controller) getCheckedClientset(kubeConfig *rest.Config) (dynamic.ResourceInterface, bool, error) {
```
skitt: We’ve said previously that we want to avoid adding boolean return values to represent information which is available elsewhere; in this instance, returning `false` implies returning a `nil` clientset.

skitt (changes requested):

coredns/gateway/controller.go
tpantelis: But that means returning `nil, nil` in some cases which is not idiomatic, ie typically a nil error means the first return value is valid. Plus we have he `nilnil` linter which is intended to flag that but, after changing the code, it didn't. The other solution is to return an error indicating "not found". I'll do that.

skitt (approved):

dfarrell07 (approved):

Issue #214, opened 2020-07-22
Aggregate ServiceImports


vthapar:
Currently ServiceImports are distributed as individual copy from each source cluster and exist that way. Aggregation is done in Plugin code, which is not optimal. Agent should aggregate ServiceImports into a single resource on destination clusters and plugin should use those. This will also make it easy to troubleshoot.

tpantelis:
bump

tpantelis:
bump

tpantelis:
bump

srampal:
This issue also causes Submariner to not support Multi-Cluster ingress load balancing via upstream K8s Gateway api which was being [tested here.](https://github.com/submariner-io/submariner/issues/1685). It would be good to escalate the priority of fixing this issue as it is also breaking the usage of Submariner  with implementations of K8s Gateway API for multi-cluster load balancing.

dfarrell07:
Wanted to make sure it's clear there was an EP merged for this here: https://github.com/submariner-io/enhancements/pull/111

dfarrell07:
Discussed on the PR scrub today. This seems important for folks that want to use Lighthouse service discovery with something else to provide connectivity, see https://github.com/submariner-io/lighthouse/issues/733. This also relates to namespaces, making an export in SubM makes the service available in all clusters vs only the clusters with a matching namespace. Also scale, having aggregate service imports would reduce the data to sync.

dfarrell07:
It seems a user is asking for this in https://github.com/submariner-io/lighthouse/issues/957.

tpantelis:
@vthapar @aswinsuryan I think we should do this for 0.15. I can work on it with your SME guidance. WDYT? cc @nyechiel

aswinsuryan:
ya, it is a long pending task, it would be good if we can get this done. Since lighthouse coredns is not going to consume this we may not have an impact in DNS resolution logic. Also to start with we can leave the IP field in the ServiceImport blank and focus on conflict resolution, like validating if all the ServiceImport with the same name and in the same namespace have the same service type.

tpantelis:
Aggregating would reduce the data to sync but we'd also lose the cluster-specific port information which is needed for per-cluster DNS requests. The spec doesn't cover that.

[...]

PR #1105, opened 2023-02-22
Introduce resolver module to the CoreDNS plugin

tpantelis:
Currently, `ServiceImport` and `EndpointSlice` processing are in separate components, with `ServiceImport`s used for resolving non-headless service queries and `EndpointSlice` used for headless services. However, the upcoming `ServiceImport` aggregation changes will require direct coordination between `ServiceImport`s and `EndpointSlice`s for resolving DNS records. Thus, it makes sense to combine the processing in [serviceimport.Map](https://github.com/submariner-io/lighthouse/blob/devel/coredns/serviceimport/map.go) and [endpointslice.Map](https://github.com/submariner-io/lighthouse/blob/devel/coredns/endpointslice/map.go) into a single component. A new _resolver_ module is introduced to contain a controller component that watches for `ServiceImport`s and `EndpointSlice`s and feeds them into a resolver component that builds internal structs used to provide `DNSRecord`s on request.
    
Subsequent commits will modify the plugin handler to user the resolver.

Related to https://github.com/submariner-io/lighthouse/issues/214


aswinsuryan (approved):

vthapar (approved):
Tested, looks good.

PR #1146, opened 2023-03-16
Create a new set instead of clearing

skitt:


tpantelis (approved):

skitt:
I’m following my own advice here ;-). https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/sets/set.go#L68

dfarrell07 (approved):

dfarrell07:
[E2E](https://github.com/submariner-io/lighthouse/actions/runs/4438151403/jobs/7789620526?pr=1146#logs) failed, re-running.

PR #1154, opened 2023-03-27
Implement aggregated ServiceImport port merging

tpantelis:
Compute the union of the constituent cluster's ports contained in their `EndpointSlices` as per the MCS spec. This is only done for ClusterIP services.

### DIFF
--- a/pkg/agent/controller/service_import_aggregator.go
+++ b/pkg/agent/controller/service_import_aggregator.go
@@ -27,8 +27,11 @@ import (
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/admiral/pkg/util"
+	"github.com/submariner-io/lighthouse/pkg/constants"
+	discovery "k8s.io/api/discovery/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/utils/pointer"
@@ -46,7 +49,7 @@ func newServiceImportAggregator(brokerClient dynamic.Interface, brokerNamespace,
 }
 
 func (a *ServiceImportAggregator) updateOnCreateOrUpdate(name, namespace string) error {
-	return a.update(name, namespace, func(existing *mcsv1a1.ServiceImport) {
+	return a.update(name, namespace, func(existing *mcsv1a1.ServiceImport) error {
 		var added bool
 
 		existing.Status.Clusters, added = slices.AppendIfNotPresent(existing.Status.Clusters,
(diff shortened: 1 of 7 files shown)


pkg/agent/controller/service_import_aggregator.go
```diff
@@ -57,28 +60,63 @@ func (a *ServiceImportAggregator) updateOnCreateOrUpdate(name, namespace string)
...
+	serviceNamespace := si.Annotations[constants.LabelSourceNamespace]
+
+	list, err := a.brokerClient.Resource(endpointSliceGVR).Namespace(a.brokerNamespace).List(context.Background(), metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			discovery.LabelManagedBy:       constants.LabelValueManagedBy,
+			constants.LabelSourceNamespace: serviceNamespace,
+			mcsv1a1.LabelServiceName:       serviceName,
+		}).String(),
+	})
+	if err != nil {
```
vthapar, line 83: Should we retry on error here, or requeue logic in caller will take care of it?

pkg/agent/controller/service_import_aggregator.go
```diff
@@ -57,28 +60,63 @@ func (a *ServiceImportAggregator) updateOnCreateOrUpdate(name, namespace string)
...
+		return errors.Wrapf(err, "error listing the EndpointSlices associated with service %s/%s",
+			serviceNamespace, serviceName)
+	}
+
+	si.Spec.Ports = make([]mcsv1a1.ServicePort, 0)
+
+	for i := range list.Items {
+		eps := a.converter.toEndpointSlice(&list.Items[i])
+		si.Spec.Ports = slices.Union(si.Spec.Ports, a.converter.toServicePorts(eps.Ports), func(p mcsv1a1.ServicePort) string {
+			return fmt.Sprintf("%s%s%d", p.Name, p.Protocol, p.Port)
```
vthapar, line 93: What about `p.AppProtocol`? It is another field in ServicePort but not seen it used much ouside of service without selectors. We can revisit this if run into any bugs due to services using appProtocol.

tpantelis: yeah I've wondering about that. We don't use it in the plugin - not sure what it's for.

pkg/agent/controller/service_import_aggregator.go
tpantelis: yes the latter.

aswinsuryan (approved):

pkg/agent/controller/service_import_aggregator.go
skitt: It’s used to define the application-level protocol used by the service, see https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol — I suspect it doesn’t get used all that often, but it’s supposed to be carried over to `EndpointSlice`s.

tpantelis: It's unclear to me how it's actually used in K8s other than propagating it to the `EndpointSlices`, which we're doing as well.  Would a user ever define two ports with matching `Name`, `Protocol` and `Port` but differing `AppProtocol`? Not sure...here we wouldn't differentiate them but the aggregated ports are only for compliance anyway, ie we don't use them.  But perhaps we should differentiate `AppProtocol` here in case someone else might use them.

Either way, like I said, we don't use `AppProtocol` in the plugin for resolution so it doesn't seem like we should include it for conflict detection and intersection in the plugin.

tpantelis: Oops correction - we're not propagating `AppProtocol` to the `EndpointSlices`. I can change that. 

Do you guys think we should differentiate `AppProtocol` when calculating the aggregated ports?

skitt: I think we should propagate it, but not use it to differentiate — it’s supposed to describe the traffic, not distinguish multiple services on the same protocol/port.

tpantelis: OK - makes sense. I'll push a separate commit to propagate it.

skitt (approved):

PR #1155, opened 2023-03-28
Implement ServiceImport conflict detection

tpantelis:
...for service type and ports. See individual commits for details.

Depends on https://github.com/submariner-io/lighthouse/pull/1154

### DIFF
--- a/pkg/agent/controller/endpoint_slice.go
+++ b/pkg/agent/controller/endpoint_slice.go
@@ -36,11 +36,11 @@ import (
 
 //nolint:gocritic // (hugeParam) This function modifies syncerConf so we don't want to pass by pointer.
 func newEndpointSliceController(spec *AgentSpecification, syncerConfig broker.SyncerConfig,
-	updateExportedServiceStatus updateExportedServiceStatusFn,
+	serviceExportClient *ServiceExportClient,
 ) (*EndpointSliceController, error) {
 	c := &EndpointSliceController{
-		clusterID:                   spec.ClusterID,
-		updateExportedServiceStatus: updateExportedServiceStatus,
+		clusterID:           spec.ClusterID,
+		serviceExportClient: serviceExportClient,
 	}
 
 	syncerConfig.LocalNamespace = metav1.NamespaceAll
@@ -122,10 +122,10 @@ func (c *EndpointSliceController) onLocalEndpointSliceSynced(obj runtime.Object,
 	} else {
 		err = c.serviceImportAggregator.updateOnCreateOrUpdate(serviceName, serviceNamespace)
 		if err != nil {
-			c.updateExportedServiceStatus(serviceName, serviceNamespace, newServiceExportCondition(constants.ServiceExportSynced,
+			c.serviceExportClient.updateStatusConditions(serviceName, serviceNamespace, newServiceExportCondition(constants.ServiceExportSynced,
 				corev1.ConditionFalse, exportFailedReason, fmt.Sprintf("Unable to export: %v", err)))
 		} else {
-			c.updateExportedServiceStatus(serviceName, serviceNamespace, newServiceExportCondition(constants.ServiceExportSynced,
+			c.serviceExportClient.updateStatusConditions(serviceName, serviceNamespace, newServiceExportCondition(constants.ServiceExportSynced,
 				corev1.ConditionTrue, "", "Service was successfully exported to the broker"))
 		}
 	}
--- a/pkg/agent/controller/endpoint_slice.go
+++ b/pkg/agent/controller/endpoint_slice.go
@@ -19,12 +19,16 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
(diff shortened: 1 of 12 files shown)


pkg/agent/controller/service_import.go
```diff
@@ -313,17 +316,31 @@ func (c *ServiceImportController) Distribute(obj runtime.Object) error {
...
 		c.converter.toUnstructured(aggregate),
 		func(obj runtime.Object) (runtime.Object, error) {
-			// TODO - check for service type conflict.
+			existing := c.converter.toServiceImport(obj)
+
+			if localServiceImport.Spec.Type != existing.Spec.Type {
+				conflict = true
+				conflictCondition := newServiceExportCondition(
+					mcsv1a1.ServiceExportConflict, corev1.ConditionTrue, typeConflictReason,
+					fmt.Sprintf("The service type %q does not match the type (%q) of the oldest service export",
```
vthapar: fmt.Sprintf("The service type %q does not match the type (%q) of the existing service export",

vthapar (changes requested):
What is the initial state set to, conflict true? We can't set conflict true before we know there is conflict for sure, nor can we set ready before we rule out conflict.

BTW, suggest edit function not working for me.

pkg/agent/controller/endpoint_slice.go
```diff
@@ -136,3 +154,70 @@ func (c *EndpointSliceController) onLocalEndpointSliceSynced(obj runtime.Object,
...
+	conflict := false
+
+	for i := range epsList.Items {
+		eps := c.serviceExportClient.toEndpointSlice(&epsList.Items[i])
+
+		servicePorts := c.serviceExportClient.toServicePorts(eps.Ports)
+		if prevServicePorts == nil {
+			prevServicePorts = servicePorts
+			intersectedServicePorts = servicePorts
+		} else if !slices.Equivalent(prevServicePorts, servicePorts, servicePortKey) {
```
vthapar, line 181: I believe this is based on assumption that service ports of aready exported service are not going to change, and we are only making sure newly exported service doesn't conflict.

What if someone changes service definition of alread exported service in a cluster?

tpantelis: It handles updates to the service ports. The workflow is

- `Service` is changed which updates the local `ServiceImport`
- `ServiceImportController.Distribute` restarts the `Endpoints` controller.
- `EndpointController` updates the `EndpointSlice`
- `EndpointSliceController` gets notified and `enqueueForConflictCheck` again

There's a unit test case for this scenario in _clusterip_service_test.go_:

`ClusterIP Service export Two clusters with differing ports initially and after updating the ports to match should correctly update the ports in the aggregated ServiceImport and clear the Conflict status condition`

skitt (approved):

tpantelis:
> What is the initial state set to, conflict true? We can't set conflict true before we know there is conflict for sure, 

Initial/normal state is no `Conflict` status condition. It only gets added if there's a conflict, thus it will only ever have a status of _true_.  If the conflict is resolved, the status condition is removed. I thought about setting it to _false_ but I think it makes sense  for the status condition to only be present if there's a conflict.

> nor can we set ready before we rule out conflict.

Actually both the `Synced` and `Conflict` conditions can be _true_, which is the case for ports, ie the service is still exported and usable even if the ports conflict since we resolve the conflict by exposing the intersection of the ports for discovery. A service type conflict is not resolvable in which case `Synced` is _false_ and `Conflict` is _true_.

BTW I intend to rename `Synced` to `Ready` after 0.15 and maybe propose the `Ready` condition type to the MCS sig.

vthapar:
> > What is the initial state set to, conflict true? We can't set conflict true before we know there is conflict for sure,
> 
> Initial/normal state is no `Conflict` status condition. It only gets added if there's a conflict, thus it will only ever have a status of _true_. If the conflict is resolved, the status condition is removed. I thought about setting it to _false_ but I think it makes sense for the status condition to only be present if there's a conflict.
> 
> > nor can we set ready before we rule out conflict.
> 
> Actually both the `Synced` and `Conflict` conditions can be _true_, which is the case for ports, ie the service is still exported and usable even if the ports conflict since we resolve the conflict by exposing the intersection of the ports for discovery. A service type conflict is not resolvable in which case `Synced` is _false_ and `Conflict` is _true_.
> 
> BTW I intend to rename `Synced` to `Ready` after 0.15 and maybe propose the `Ready` condition type to the MCS sig.

+1 to `Ready` state. `Conflict` false does seem superflous if we only add conflict when it is present. 

P.S. Sorry, closed by mistake when trying to post comment.

pkg/agent/controller/endpoint_slice.go
vthapar: My quetsion was specific for scenario where service changes to introduce a conflict, but from workflow looks like it is covered too. Do we have an E2E for this scenario, conflict created after export creation?

vthapar (approved):
Tested locally, works great. Needs more E2E for aggregation and conflict scenarios, but can add them later.

pkg/agent/controller/endpoint_slice.go
tpantelis: There's unit tests that cover the scenario but not E2E. We can add E2E later.

aswinsuryan (approved):

PR #1157, opened 2023-03-29
Implement legacy ServiceImport migration in the agent controller

tpantelis:
Legacy per-cluster `ServiceImports` are no longer created but, for rolling cluster upgrades, there may be a period of time where there is a mix of upgraded and non-upgraded clusters. The latter still need to observe the per-cluster `ServiceImports` so they need to remain on the broker during the transition period.
    
An upgraded cluster can remove its legacy per-cluster `ServiceImport` from the broker when it observes that all of the legacy `ServiceImport` cluster names are present in the aggregated `ServiceImport` status cluster name list. This indicates that all constituent clusters have been upgraded.
    
Introduced a `ServiceImportMigrator` that integrates with the `ServiceImportController` to handle syncing of legacy `ServiceImports` from the broker and removing the local legacy `ServiceImport` from the broker based on the criteria above. The latter is triggered whenever an aggregated `ServiceImport` is synced from the broker, either new or updated.

Depends on https://github.com/submariner-io/lighthouse/pull/1155

### DIFF
--- a/pkg/agent/controller/reconciliation_test.go
+++ b/pkg/agent/controller/reconciliation_test.go
@@ -20,7 +20,7 @@ package controller_test
 
 import (
 	"context"
-	"time"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


dfarrell07:
Needs a manual rebase

aswinsuryan (approved):

pkg/agent/controller/reconciliation_test.go
```diff
@@ -102,26 +111,33 @@ var _ = Describe("Reconciliation", func() {
...
-			testutil.EnsureNoActionsForResource(&brokerDynClient.Fake, "serviceimports", "delete")
 			testutil.EnsureNoActionsForResource(&brokerDynClient.Fake, "endpointslices", "delete")
+
+			// For migration cleanup, it may attempt to delete a local legacy ServiceImport from the broker so ignore it.
+			Consistently(func() bool {
+				siActions := brokerDynClient.Fake.Actions()
+				for i := range siActions {
+					if siActions[i].GetResource().Resource == "serviceimports" && siActions[i].GetVerb() == "delete" &&
+						!strings.Contains(siActions[i].(testing.DeleteAction).GetName(), t.cluster1.clusterID) {
+						fmt.Printf("****DEL: %s, %s, %s\n", siActions[i].GetResource().Resource, siActions[i].GetVerb(),
```
skitt: Do we want to keep this?

tpantelis: oh no - nice catch - not sure how I missed that.

skitt (approved):

Issue #1090, opened 2023-02-10
Service address is resolved from the local cluster while readiness probe is not healthy yet


MartijnStraatman:
**What happened**:

We have two exported services. As soon as the pod starts and it waits for the readiness probe to be healthy the local service address is resolved. Since the pod is not ready yet no traffic is routed to the pod. It will time out.

**What you expected to happen**:
I expected it to failover to the other cluster which has exported the service. That pod is healthy.

If we look at the endpoints

when all is `ok`

```bash
$ kubectl get endpoints
NAME                  ENDPOINTS          AGE
app                      10.0.0.1:8080   173m
```

when pod is `not ready`
```bash
$ kubectl get endpoints
NAME                  ENDPOINTS          AGE
app                                                     168m
```

when pod is `downscaled`
```bash
$ kubectl get endpoints
NAME                  ENDPOINTS          AGE
app                      <none>                  169m   
```

When the pod is not ready the endpoint has a list of ` notReadyAddresses`
```
subsets:
- notReadyAddresses:
  - ip: 10.0.0.1
  ```

Maybe you could check the state of the endpoint and do a failover?

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a deployement with a readiness probe
2. Export the service
3. Make sure one of the deployments is not ready
4. Try to resolve the service on the clusterset.local domain it fails or timesout.

dfarrell07:
Thanks for the report @MartijnStraatman! Discussing with @aswinsuryan, @vthapar, and @tpantelis, it seems like we need to adjust our handling of notReadyAddresses to avoid this.

It would help to know if this is headless or clusterIP service.

MartijnStraatman:
This was a clusterIP service. I have not tested headless services yet.

PR #1241, opened 2023-05-24
Honor publishNotReadyAddresses flag for headless services

tpantelis:
For headless services, K8s does not publish DNS records for pods that are not ready unless `publishNotReadyAddresses` is set to true on the `Service`. The documentation for this flags states the "_the primary use case for setting this field is for a StatefulSet's Headless Service to propagate SRV DNS records for its Pods for the purpose of peer discovery. The Kubernetes controllers that generate Endpoints and EndpointSlice resources for Services interpret this to mean that all endpoints are considered "ready" even if the Pods themselves are not_".

Submariner previously did publish DNS records for pods that are not ready but was changed not do to so by https://github.com/submariner-io/lighthouse/issues/1090. However that issue really pertained to Cluster IP services where the K8s behavior is to resolve the service even if no backing pod is ready. This PR honors the `publishNotReadyAddresses` flag for headless services.

For the implementation, a new annotation, `lighthouse.submariner.io/publish-not-ready-addresses`, is introduced to propagate the `Service`'s `publishNotReadyAddresses` flag to the `EndpointSlice` via the local `ServiceImport`. See individual commits for more details.


skitt (approved):

coredns/resolver/endpoint_slice.go
```diff
@@ -132,12 +133,14 @@ func (i *Interface) putHeadlessEndpointSlice(key, clusterID string, endpointSlic
 
 	mcsPorts := mcsServicePortsFrom(endpointSlice.Ports)
 
+	publishNotReadyAddresses := endpointSlice.Annotations[constants.PublishNotReadyAddresses] == strconv.FormatBool(true)
+
 	for i := range endpointSlice.Endpoints {
 		endpoint := &endpointSlice.Endpoints[i]
 
 		// Skip if not ready. Note: we're treating nil as ready to be on the safe side as the EndpointConditions doc
```
vthapar: nit: Update the comment to reflect the new check.

vthapar (changes requested):
Changes look good, testing locally. Just a small nit regarding a comment.

aswinsuryan (approved):

vthapar (approved):

PR #1272, opened 2023-06-07
Use the retrieved local K8s EndpointSlice in the CoreDNS plugin

### LINKED ISSUE #1107: Use the retrieved local K8s EndpointSlice in the CoreDNS plugin
https://github.com/submariner-io/lighthouse/pull/757 made changes to retrieve the local K8s `EndpointSlice` to replace the local cluster's `EndpointSlice` created by LH the due to **globalnet** IPs in the latter not being routable locally. The implementation loops through the LH `EndpointSlice`'s endpoints and searches for the equivalent endpoint in the K8s `EndpointSlice` as follows:

```
addresses := endpoint.Addresses
if localSlice != nil {
    for _, localEndpoint := range localSlice.Endpoints {
        if localEndpoint.NodeName != nil && endpoint.NodeName != nil && *localEndpoint.NodeName == *endpoint.NodeName &&
	        (endpoint.Hostname == nil || localEndpoint.TargetRef.Name == *endpoint.Hostname) {
	    addresses = localEndpoint.Addresses
	}
    }
}
```

It's unclear as to the purpose of the specific Nodename/Hostname checks. If no matching local endpoint is found and **globalnet** is enabled then falling back to the LH endpoint  addresses is problematic and would yield the same issue that https://github.com/submariner-io/lighthouse/pull/757 was intended to solve.

It's also unclear why it doesn't just use the K8s `EndpointSlice` as is - it seems we should be able to do that, ie the LH `EndpointSlice` derived from the K8s `Endpoints` resource should be equivalent to the K8s `EndpointSlice` (except for the global IPs replacing the local IPs).

tpantelis:
Fixes https://github.com/submariner-io/lighthouse/issues/1107

### DIFF
--- a/coredns/resolver/endpoint_slice.go
+++ b/coredns/resolver/endpoint_slice.go
@@ -48,7 +48,6 @@ func (i *Interface) PutEndpointSlice(endpointSlice *discovery.EndpointSlice) boo
 	logger.Infof("Put EndpointSlice %q on cluster %q", key, clusterID)
 
 	var localEndpointSliceErr error
-	var localEndpointSlice *discovery.EndpointSlice
 
 	globalnetEnabled := endpointSlice.Annotations[constants.GlobalnetEnabled] == strconv.FormatBool(true)
 	localClusterID := i.clusterStatus.GetLocalClusterID()
@@ -57,7 +56,13 @@ func (i *Interface) PutEndpointSlice(endpointSlice *discovery.EndpointSlice) boo
 		// The EndpointSlice is from the local cluster. With globalnet enabled, the local global endpoint IPs aren't
 		// routable in the local cluster so we retrieve the K8s EndpointSlice and use those endpoints. Note that this
 		// only applies to headless services.
+		var localEndpointSlice *discovery.EndpointSlice
 		localEndpointSlice, localEndpointSliceErr = i.getLocalEndpointSlice(endpointSlice)
+
+		if localEndpointSliceErr == nil {
+			endpointSlice.Endpoints = localEndpointSlice.Endpoints
+			endpointSlice.Ports = localEndpointSlice.Ports
+		}
 	}
 
 	i.mutex.Lock()
@@ -81,7 +86,7 @@ func (i *Interface) PutEndpointSlice(endpointSlice *discovery.EndpointSlice) boo
 		return true
 	}
 
-	i.putHeadlessEndpointSlice(key, clusterID, endpointSlice, serviceInfo, localEndpointSlice)
+	i.putHeadlessEndpointSlice(key, clusterID, endpointSlice, serviceInfo)
 
 	return false
 }
@@ -127,8 +132,7 @@ func (i *Interface) putClusterIPEndpointSlice(key, clusterID string, endpointSli
 	return false
 }
 
-func (i *Interface) putHeadlessEndpointSlice(key, clusterID string, endpointSlice *discovery.EndpointSlice, serviceInfo *serviceInfo,
-	localEndpointSlice *discovery.EndpointSlice) {
+func (i *Interface) putHeadlessEndpointSlice(key, clusterID string, endpointSlice *discovery.EndpointSlice, serviceInfo *serviceInfo) {
 	clusterInfo := &clusterInfo{
 		endpointRecordsByHost: make(map[string][]DNSRecord),
 	}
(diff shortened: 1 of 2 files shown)


skitt (approved):

vthapar:
https://github.com/submariner-io/lighthouse/pull/757#discussion_r873595715 gives some context to it. Not 100% sure if we need this check, in agent code these were added due to some issues with StatefulSets where some of the fields were empty or hostname/podname had a mismatch [don't recall exactly]. Also, issue was seen only in particular statefulset setup created by Raphael which, iirc, depended on how statefulset was defined. I do think we got it covered by E2E, so if E2e works, we should be fine.

So, while okay with this change, will need to keep an eye out for any SS issues that may crop up later.

vthapar (approved):
Tentatively approved as changes look fine, but would be good to make sure StatefulSets +Globalnet with local service works fine.

tpantelis:
> https://github.com/submariner-io/lighthouse/pull/757#discussion_r873595715 gives some context to it. Not 100% sure if we need this check, in agent code these were added due to some issues with StatefulSets where some of the fields were empty or hostname/podname had a mismatch [don't recall exactly].

So it seems there's a difference between how we derive our `EndpointSlice` addresses from the `Endpoints` and how K8s does it wrt the `Hostname` field. Notice [here](https://github.com/submariner-io/lighthouse/blob/6360c8a6f93db5cb6e46d180d13031380f980618/pkg/agent/controller/endpoint.go#L298) that we set `Hostname` from the `TargetRef` if `Hostname` isn't specifically set. I'm not clear on why we do this (shouldn't we just set the fields as is from the `EndpointAddress`?). I suspect K8s doesn't do this so `Hostname` is empty in this case, hence the mismatch. 

I'm not sure of the ramifications if this discrepancy but to preserve the behavior we could mimic our agent behavior here when replacing our `EndpointSlice` with K8s's.

aswinsuryan:
Is this going to break https://github.com/submariner-io/lighthouse/issues/993?

tpantelis:
> Is this going to break https://github.com/submariner-io/lighthouse/issues/993?

In what way? That PR was related to non-headless services.

coredns/resolver/endpoint_slice.go
```diff
@@ -153,7 +154,17 @@ func (i *Interface) putHeadlessEndpointSlice(key, clusterID string, endpointSlic
...
-		var records []DNSRecord
+		var (
+			records  []DNSRecord
+			hostname string
+		)
+
+		switch {
+		case endpoint.Hostname != nil && *endpoint.Hostname != "":
+			hostname = *endpoint.Hostname
+		case endpoint.TargetRef != nil && strings.ToLower((*endpoint.TargetRef).Kind) == "pod":
```
tpantelis, line 165: @vthapar @aswinsuryan I assume we only want to interpret the TargetRef as a host name if it's a Pod?

tpantelis:
> I'm not sure of the ramifications if this discrepancy but to preserve the behavior we could mimic our agent behavior here when replacing our `EndpointSlice` with K8s's.

Pushed commit https://github.com/submariner-io/lighthouse/pull/1272/commits/d6beaad9c441453b4da2350b79826826e650a26b

coredns/resolver/endpoint_slice.go
vthapar: Yes

vthapar:
> > [#757 (comment)](https://github.com/submariner-io/lighthouse/pull/757#discussion_r873595715) gives some context to it. Not 100% sure if we need this check, in agent code these were added due to some issues with StatefulSets where some of the fields were empty or hostname/podname had a mismatch [don't recall exactly].
> 
> So it seems there's a difference between how we derive our `EndpointSlice` addresses from the `Endpoints` and how K8s does it wrt the `Hostname` field. Notice [here](https://github.com/submariner-io/lighthouse/blob/6360c8a6f93db5cb6e46d180d13031380f980618/pkg/agent/controller/endpoint.go#L298) that we set `Hostname` from the `TargetRef` if `Hostname` isn't specifically set. I'm not clear on why we do this (shouldn't we just set the fields as is from the `EndpointAddress`?). I suspect K8s doesn't do this so `Hostname` is empty in this case, hence the mismatch.
> 
> I'm not sure of the ramifications if this discrepancy but to preserve the behavior we could mimic our agent behavior here when replacing our `EndpointSlice` with K8s's.

When we create and sync Endpointslice from agent, we don't add TargetRef field to it. This is to reduce the size of objects being synced for scale considerations. That is why we move information from TargetRef into hostname for Coredns plugin to consume on remote clusters. Similarly, in plugin we need to get hostname from TargetRef if hostname is empty as during testing, esp with StatefulSets I'd encountered situations where hostname was empty.

vthapar (approved):

aswinsuryan:
> > Is this going to break #993?
> 
> In what way? That PR was related to non-headless services.

Oh ok I think I go it wrong, thanks for clarifying.

aswinsuryan (approved):

PR #1287, opened 2023-06-20
Log version in lighthouse pods

### LINKED ISSUE #1279: Lighthouse components should log their version on startup

Each component (each `main.go` file) should log its version on startup, [in the same way as the operator](https://github.com/submariner-io/submariner-operator/blob/8fb881cb70d8ffe7eb04545c7a4071af895e7bb5/main.go#L69) (with a version defined at build-time).

**Why is this needed**:

This will allow deployed versions to be determined precisely.

vthapar:
Log version for Lighthoue agent and coredns pods on startup.

Fixes #1279

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ TARGETS := $(shell ls -p scripts | grep -v -e / -e deploy)
 override E2E_ARGS += cluster1 cluster2 cluster3
 override UNIT_TEST_ARGS += test/e2e
 export LIGHTHOUSE = true
+export LDFLAGS = -X main.version=$(VERSION)
 
 # Targets to make
 
--- a/coredns/main.go
+++ b/coredns/main.go
@@ -61,7 +61,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/tls"
 	_ "github.com/coredns/coredns/plugin/trace"
 	_ "github.com/coredns/coredns/plugin/whoami"
-	_ "github.com/submariner-io/lighthouse/coredns/plugin"
+	lighthouse "github.com/submariner-io/lighthouse/coredns/plugin"
 )
 
 var directives = []string{
@@ -105,7 +105,10 @@ var directives = []string{
 	"on",
 }
 
+var version = "not-compiled-properly"
+
 func init() {
+	lighthouse.Version = version
 	dnsserver.Directives = directives
 }
 
--- a/coredns/plugin/setup.go
+++ b/coredns/plugin/setup.go
@@ -42,6 +42,7 @@ import (
 var (
 	masterURL  string
 	kubeconfig string
+	Version    string
 )
 
 // Hooks for unit tests.
@@ -70,7 +71,7 @@ func init() {
 // setup is the function that gets called when the config parser see the token "lighthouse". Setup is responsible
 // for parsing any extra options the this plugin may have. The first token this function sees is "lighthouse".
 func setupLighthouse(c *caddy.Controller) error {
-	log.Debug("In setupLighthouse")
+	log.Infof("submariner-lighthouse-coredns version: %v", Version)
 
 	l, err := lighthouseParse(c)
 	if err != nil {
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -48,6 +48,7 @@ var (
 	kubeConfig string
 	help       = false
 	logger     = log.Logger{Logger: logf.Log.WithName("main")}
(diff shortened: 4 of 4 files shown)


Makefile
```diff
@@ -25,6 +25,7 @@ TARGETS := $(shell ls -p scripts | grep -v -e / -e deploy)
 override E2E_ARGS += cluster1 cluster2 cluster3
 override UNIT_TEST_ARGS += test/e2e
 export LIGHTHOUSE = true
+export LDFLAGS = -X main.VERSION=$(VERSION)
```
skitt:
```suggestion
export LDFLAGS = -X main.version=$(VERSION)
```

coredns/main.go
```diff
@@ -105,7 +105,10 @@ var directives = []string{
 	"on",
 }
 
+var VERSION = "not-compiled-properly"
```
skitt:
```suggestion
var version = "not-compiled-properly"
```

coredns/main.go
```diff
@@ -105,7 +105,10 @@ var directives = []string{
...
+
 func init() {
+	lighthouse.VERSION = VERSION
```
skitt:
```suggestion
	lighthouse.Version = version
```

coredns/plugin/setup.go
```diff
@@ -42,6 +42,7 @@ import (
 var (
 	masterURL  string
 	kubeconfig string
+	VERSION    string
```
skitt:
```suggestion
	Version    string
```

coredns/plugin/setup.go
```diff
@@ -70,7 +71,7 @@ func init() {
 // setup is the function that gets called when the config parser see the token "lighthouse". Setup is responsible
 // for parsing any extra options the this plugin may have. The first token this function sees is "lighthouse".
 func setupLighthouse(c *caddy.Controller) error {
-	log.Debug("In setupLighthouse")
+	log.Infof("submariner-lighthouse-coredns version: %v", VERSION)
```
skitt:
```suggestion
	log.Infof("submariner-lighthouse-coredns version: %v", Version)
```

pkg/agent/main.go
```diff
@@ -48,6 +48,7 @@ var (
 	kubeConfig string
 	help       = false
 	logger     = log.Logger{Logger: logf.Log.WithName("main")}
+	VERSION    = "not-compiled-properly"
```
skitt:
```suggestion
	version    = "not-compiled-properly"
```

pkg/agent/main.go
```diff
@@ -111,7 +112,7 @@ func main() {
 	localClient, err := dynamic.NewForConfig(cfg)
 	exitOnError(err, "Error creating dynamic client")
 
-	logger.Info("Starting submariner-lighthouse-agent")
+	logger.Infof("Starting submariner-lighthouse-agent version: %v", VERSION)
```
skitt:
```suggestion
	logger.Infof("Starting submariner-lighthouse-agent version: %v", version)
```

skitt (changes requested):

coredns/main.go
tpantelis: Why don't we just do `log.Infof("submariner-lighthouse-coredns version: %v", version)` in `main` before L116 below?

pkg/agent/main.go
```diff
@@ -111,7 +112,7 @@ func main() {
 	localClient, err := dynamic.NewForConfig(cfg)
 	exitOnError(err, "Error creating dynamic client")
 
-	logger.Info("Starting submariner-lighthouse-agent")
+	logger.Infof("Starting submariner-lighthouse-agent version: %v", version)
```
skitt: Grammatically, this should be

```suggestion
	logger.Infof("Starting submariner-lighthouse-agent version %v", version)
```

but to keep it consistent with other version output, and perhaps simplify parsing should we want to extract this in `subctl`, let’s make it

```suggestion
	logger.Infof("submariner-lighthouse-agent version: %v", version)
```

skitt (dismissed):
Sorry, something else I didn’t notice earlier.

skitt (approved):
In future, please don’t dismiss “stale” reviews when you address the changes requested — you can mark the changes resolved, but dismissing reviews should only be done if necessary to unblock a PR.

aswinsuryan (approved):

tpantelis (approved):

PR #1289, opened 2023-06-20
Add E2E test for nonexistent imported service NS

### LINKED ISSUE #605: Services are accessible even if their namespace doesn’t exist locally
**What happened**:

In a two-cluster setup running Lighthouse (devel), I set up a namespace and service in cluster2, and exported it. The namespace didn’t exist in cluster1, but even so, the exported service was accessible.

Given that the namespace containing the service doesn’t exist in cluster1, the service shouldn’t be imported there.

1. `make deploy` from Lighthouse or the operator
2. create a namespace in one cluster only
3. create a service there
4. export it
5. check that it can be resolved and accessed from the other cluster

tpantelis:
...to verify an imported service is not accessible if the service namespace doesn't exist locally.

Since the service is created in the framework's generated namespace, the test deletes this namespace from the source cluster and creates another namespace to contain the netshoot deployment.

The test was placed in a new folder named "_internal_" so it isn't run with `subctl verify`. We should only need to verify this functionality in CI plus it has the overhead of an additional namespace.

Fixes https://github.com/submariner-io/lighthouse/issues/605

### DIFF
--- a/test/e2e/internal/service_discovery.go
+++ b/test/e2e/internal/service_discovery.go
@@ -1,0 +1,99 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	lhframework "github.com/submariner-io/lighthouse/test/e2e/framework"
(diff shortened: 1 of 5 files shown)


test/e2e/internal/service_discovery.go
```diff
@@ -0,0 +1,85 @@
...
+
+	netshootPodList := f.NewNetShootDeploymentInNS(framework.ClusterA, netshootNS.Name)
+
+	svc, err := f.GetService(framework.ClusterB, nginxServiceClusterB.Name, nginxServiceClusterB.Namespace)
+	Expect(err).NotTo(HaveOccurred())
+
+	nginxServiceClusterB = svc
+
+	for i := 1; i <= 5; i++ {
+		time.Sleep(500 * time.Millisecond)
```
vthapar, line 83: Why need this sleep? Way `verifyIP` is written, it will retry till it works or till it fails. We avoid using sleep to avoid adding fixed delays in testing.

vthapar (changes requested):

test/e2e/internal/service_discovery.go
tpantelis: I'm verifying that it consistently doesn't resolve the IP otherwise, if done too quickly, it may falsely succeed b/c the service wasn't yet imported into the cluster. So trying over a couple of seconds should avoid that. It's the equivalent of using Ginkgo's `Consistently` instead of `Eventually`.

tpantelis:
@vthapar I expanded the test to also recreate the namespace and verify the service is discovered.

skitt (approved):

vthapar (approved):

PR #1308, opened 2023-06-30
Remote service is not available if namespace is deleted and recreated

### LINKED ISSUE #685: Remote service is not available if namespace is deleted and recreated
**What happened**:

1. Export service `my-svc` on namespace `my-ns` on `cluster1`
2. Create namespace `my-ns` on `cluster2`
3. LH agent creates `endpointslice` on `my-ns` at `cluster2` 
4. `my-svc` is now accessible on cluster2 at endpoint `my-svc.my-ns.svc.clusterset.local`
5. Delete namespace `my-ns` on `cluster2`
6. Create namespace `my-ns` on `cluster2`

`endpointslice` shall be recreated once namespace is recreated.
In practice, it is not created again and remote service is not available

If LH agent is restarted, `endpointslice` is downloaded from broker and situation is fixed.

tpantelis:
Rather than watching namespaces we can simply specify a resync period for the broker syncers so resources from the broker are periodically re-evaluated.

Fixes #685


pkg/agent/controller/endpoint_slice.go
```diff
@@ -36,6 +34,7 @@ import (
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+	"strings"
```
skitt: This shouldn’t move.

skitt (changes requested):
Please update the commit title to describe the change, rather than the bug being fixed.

pkg/agent/controller/types.go
```diff
@@ -38,6 +39,8 @@ const (
 	portConflictReason = "ConflictingPorts"
 )
 
+var brokerResyncePeriod = time.Minute * 2
```
vthapar, line 42: Is this fixed duration or exponential backoff with max of 2 minutes? Do we log resync and at what severity?

Since user will need to create namespace for it to succeed, it may take long enough that it fills up the log.

If it is fixed at 2 minutes, then we need to consider it in E2E as it test will take up to 2 minutes to pass this requirement.

tpantelis: > Is this fixed duration or exponential backoff with max of 2 minutes? 

Fixed. This isn't a retry mechanism but a periodic resync.

> Do we log resync and at what severity?
> 
> Since user will need to create namespace for it to succeed, it may take long enough that it fills up the log.
> 

No debug or info level is logged here by default. If a namespace doesn't exist, an error is logged by admiral but that would also happen on periodic retries on initial failure.  So theoretically the log could fill up but it would take a long time at a 2 min interval.

> If it is fixed at 2 minutes, then we need to consider it in E2E as it test will take up to 2 minutes to pass this requirement.

I didn't write an E2E for this edge case for that reason. Ideally, we'd have to somehow reduce the duration for E2E.  I did recently add a test for when the namespace doesn't initially exist which is handled by the retry which essentially follows the same path in our code.

pkg/agent/controller/endpoint_slice.go
tpantelis: Fixed

skitt (approved):

vthapar (approved):

PR #1332, opened 2023-08-01
Add version flag to LH binaries

### LINKED ISSUE #1333: Add version flag to LH binaries
Add `--version` flag to lighthouse binaries so it can be used to get actual runtime version and used in `subctl show version`

**Why is this needed**:
Refer https://github.com/submariner-io/enhancements/pull/204 for details

vthapar:
Adds `--version` flag to Lighthouse Agent and `--subm-version` flag to Lighthouse Coredns. CoreDNS already has `--version` which prints CoreDNS version, so we can't use the same flag for Lighthouse version.

Refer https://github.com/submariner-io/enhancements/pull/204

Fixes #1333

### DIFF
--- a/coredns/main.go
+++ b/coredns/main.go
@@ -22,6 +22,8 @@ package main
 // https://coredns.io/2017/07/25/compile-time-enabling-or-disabling-plugins/#build-with-external-golang-source-code
 
 import (
+	"flag"
+
 	_ "github.com/coredns/caddy/onevent"
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/coremain"
@@ -61,7 +63,9 @@ import (
 	_ "github.com/coredns/coredns/plugin/tls"
 	_ "github.com/coredns/coredns/plugin/trace"
 	_ "github.com/coredns/coredns/plugin/whoami"
-	lighthouse "github.com/submariner-io/lighthouse/coredns/plugin"
+	"github.com/submariner-io/admiral/pkg/names"
+	admversion "github.com/submariner-io/admiral/pkg/version"
+	_ "github.com/submariner-io/lighthouse/coredns/plugin"
 )
 
 var directives = []string{
@@ -105,13 +109,23 @@ var directives = []string{
 	"on",
 }
 
-var version = "not-compiled-properly"
+var (
+	version     = "not-compiled-properly"
+	showVersion = false
+)
 
 func init() {
-	lighthouse.Version = version
+	flag.BoolVar(&showVersion, "subm-version", showVersion, "Show version")
 	dnsserver.Directives = directives
 }
 
 func main() {
+	flag.Parse()
+	admversion.Print(names.LighthouseCoreDNSComponent, version)
+
+	if showVersion {
+		return
+	}
+
 	coremain.Run()
 }
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -31,8 +31,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/log/kzerolog"
+	"github.com/submariner-io/admiral/pkg/names"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	"github.com/submariner-io/admiral/pkg/util"
+	admversion "github.com/submariner-io/admiral/pkg/version"
 	"github.com/submariner-io/lighthouse/pkg/agent/controller"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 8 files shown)


coredns/main.go
```diff
@@ -105,13 +107,23 @@ var directives = []string{
 	"on",
 }
 
-var version = "not-compiled-properly"
+var (
+	version     = "not-compiled-properly"
+	submVersion = false
```
tpantelis:
```suggestion
	showVersion = false
```

pkg/agent/main.go
```diff
@@ -43,16 +43,20 @@ import (
...
 var (
-	masterURL  string
-	kubeConfig string
-	help       = false
-	logger     = log.Logger{Logger: logf.Log.WithName("main")}
-	version    = "not-compiled-properly"
+	masterURL   string
+	kubeConfig  string
+	help        = false
+	submVersion = false
```
tpantelis:
```suggestion
	showVersion = false
```

coredns/main.go
```diff
@@ -105,13 +107,23 @@ var directives = []string{
...
 	lighthouse.Version = version
 	dnsserver.Directives = directives
 }
 
 func main() {
+	flag.Parse()
+	if submVersion {
+		fmt.Printf("submariner-lighthouse-coredns version: %s", version)
+		return
+	}
```
tpantelis: We also log the version in [setupLighthouse](https://github.com/submariner-io/lighthouse/blob/6cab076201b9d77adeb8714d34d54d66324d8e56/coredns/plugin/setup.go#L74). How about we just print the version here (and remove the log in `setupLighthouse`) and exit if the version was requested? Same with the agent.

```suggestion
        fmt.Printf("submariner-lighthouse-coredns version: %s", version)
        
	if showVersion {
		return
	}
```

tpantelis (changes requested):

coredns/main.go
vthapar: Yeah, I thought about printing and not using logger, but what if someone is using an external logger? printf will not show up in that, right? I like idea of using common print for both coz it ensures version message will be same.

tpantelis: How would someone use/setup an external logger? But regardless, wouldn't `fmt.Printf` go to the pod log no matter what? We'll scrape from that.

vthapar: Tried it out, `fmt.Printf` didn't show up in logs, only if we use logger. So will continue with current approach. Changed `submVersion` to `showVersion`.

tpantelis: That’s odd. So `fmt.Printf` shows up when you run it with `--version` but not when run normally? What’s the diff? Aren’t you scraping the logs in the same manner?

tpantelis: I looked at **klog** code and it writes to stderr by default, ie the `logtostderr` flag defaults to true. In fact there’s no option to write to stdout. **admiral**'s zerolog also writes to [stderr](https://github.com/submariner-io/admiral/blob/6febc35a01d2fe109fa46d8432edc55dc5a33332/pkg/log/kzerolog/kzerolog.go#L78). I think that's b/c `kubectl` scrapes the log from stderr. So if we do `fmt.Fprintf(os.Stderr, …)`, it should show up in the pod logs.

vthapar: @tpantelis Okay, print to stderr worked. But two issues with it:

1. Doesn't work for `coredns` For some reason any output to stdout or stderr before plugin directive is loaded results in CrashLoopBackoff on coredns. So we can't print before plugin is loaded and if we do that `--version` will not work coz coredns is already running so it will fail to bring it up again.
2. Using print to stderr shows up in logs but without timestamp or line of code information:

```
submariner-lighthouse-agent version: devel-e7cead11fe84
2023-08-03T05:42:07.414Z INF ..e/pkg/agent/main.go:105 main                 Arguments: [/usr/local/bin/lighthouse-a
gent -alsologtostderr -v=2]
```

Am okay either ways but do prefer to use logger for log file so that version information shows up with timestamp et al in the logs. Thoughts?

tpantelis: > Doesn't work for coredns For some reason any output to stdout or stderr before plugin directive is loaded results in CrashLoopBackoff on coredns. 

That's weird. Do you get a stack trace? We can log it where we do now.

> Am okay either ways but do prefer to use logger for log file so that version information shows up with timestamp et al in the logs. Thoughts?

I think it's fine w/o timestamp. It's not really needed since it's the first log entry and the pertinent info is the version.  But if you want it then set up the logger beforehand.

vthapar: > That's weird. Do you get a stack trace? We can log it where we do now.

No stack trace. Only error about missing `dns` directive.

tpantelis (commented):
I think to be consistent we should use `fmt.Fprintf` in `setupLighthouse` as well.

vthapar:
> I think to be consistent we should use `fmt.Fprintf` in `setupLighthouse` as well.

There are quite a few log statements printed from CoreDNS before setupLighthouse. It works for agent coz it is the first statement so no timestamp is fine. For setupLighthouse I prefer having the timestamp as it gives indication of when lighthouse was loaded.

Also, I think I tried it that way and got some issues, but it was probably coz I was doing printf in main.go

tpantelis (approved):

vthapar:
Managed to make it work without needing a logger in `setupLighthouse`.

As discussed on slack, will use `Fprintf(os.Stderr...)` in `main.go`, remove log entry from `setupLighthouse` and `metrics.go:init()` so that version is first line in logs.

tpantelis (approved):

coredns/main.go
```diff
@@ -105,13 +111,24 @@ var directives = []string{
...
 func init() {
-	lighthouse.Version = version
+	flag.BoolVar(&showVersion, "subm-version", showVersion, "Show version")
 	dnsserver.Directives = directives
 }
 
 func main() {
+	flag.Parse()
+	admversion.Print(names.LighthouseCoreDNSComponent, version)
+	fmt.Fprintf(os.Stderr, "submariner-lighthouse-coredns version: %s\n", version)
```
tpantelis:
```suggestion
```

tpantelis (changes requested):

tpantelis (approved):

sridhargaddam (approved):

PR #1400, opened 2023-10-03
Alleviate excessive periodic logging during conflict check

tpantelis:
If a service is imported without a local service export, the conflict checking tries to update the status of a non-existent `ServiceExport`, specifically to remove the Conflict status condition. This results in a log message: _ServiceExport (...) not found - unable to update status_
    
Due to the 2 minute resync intervaL for the `EndpointSlice` syncer, this can exhaust the pod disk storage over time.
    
([first commit](https://github.com/submariner-io/lighthouse/pull/1400/commits/16a68a73b66fd69f0747e3efc160bfab8a45bde8)) To alleviate this issue, check the local cache for a ServiceExport and, if non-existent, skip the conflict checking. If it does exist but the Conflict status condition doesn't exist in the local ServiceExport then elide trying to remove the status condition. Also, change the log level to TRACE for the offending message.
    
([second commit](https://github.com/submariner-io/lighthouse/pull/1400/commits/05d99653edeee932fa71c78fc50048ca5a11c88e)) Optimize conflict checking using the local cache to list the `EndpointSlices` instead of the API server client.

### DIFF
--- a/pkg/agent/controller/types.go
+++ b/pkg/agent/controller/types.go
@@ -39,7 +39,7 @@ const (
 	portConflictReason = "ConflictingPorts"
 )
 
-var brokerResyncePeriod = time.Minute * 2
+var BrokerResyncPeriod = time.Minute * 2
 
 type converter struct {
 	scheme *runtime.Scheme
@@ -118,6 +118,7 @@ type EndpointSliceController struct {
 type ServiceExportClient struct {
 	dynamic.NamespaceableResourceInterface
 	converter
+	localSyncer syncer.Interface
 }
 
 type globalIngressIPCache struct {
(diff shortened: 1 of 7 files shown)


pkg/agent/controller/types.go
```diff
@@ -39,7 +39,7 @@ const (
 	portConflictReason = "ConflictingPorts"
 )
 
-var brokerResyncePeriod = time.Minute * 2
+var BrokerResyncePeriod = time.Minute * 2
```
skitt: While you’re at it, you could fix the typo ;-)

```suggestion
var BrokerResyncPeriod = time.Minute * 2
```

skitt (approved):

aswinsuryan (approved):

sridhargaddam (approved):

sridhargaddam:
Do we want to backport this PR?

skitt:
> Do we want to backport this PR?

I think so, I’ve taken care of it — if there are objections feel free to raise them on the backport PR.

Issue #1322, opened 2023-07-29
De-duplicate EndpointSlice endpoints


tpantelis:
As outlined [here](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints), duplicate endpoints are possible with K8s EndpointSlices. We should de-duplicate them in the coredns plugin.

PR #1445, opened 2023-11-20
Update K8s versions tested in E2E

dfarrell07:
Add Kubernetes 1.26 and 1.28, remove end-of-life 1.22-1.25.

Add K8s 1.21, the oldest working version for Service Discovery.

Per discussions, only test ends of supported K8s version range. Don't
test versions in the middle of the range to avoid GHA runs that don't
add useful coverage.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -46,11 +46,15 @@ jobs:
       matrix:
         deploytool: ['operator']
         globalnet: ['', 'globalnet']
-        k8s_version: ['1.25']
+        # Run most tests against the latest K8s version
+        k8s_version: ['1.28']
         include:
-          - k8s_version: '1.22'
-          - k8s_version: '1.23'
-          - k8s_version: '1.24'
+          # Oldest Kubernetes version thought to work with SubM's Service Discovery.
+          # If this breaks, we may advance the oldest-working K8s version instead of fixing it. See:
+          # https://submariner.io/development/building-testing/ci-maintenance/
+          - k8s_version: '1.21'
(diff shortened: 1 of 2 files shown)


dfarrell07:
[Go Unit Tests](https://github.com/submariner-io/lighthouse/actions/runs/6933208128/job/18858865106?pr=1445#logs) failed, re-running.

dfarrell07:
> [Go Unit Tests](https://github.com/submariner-io/lighthouse/actions/runs/6933208128/job/18858865106?pr=1445#logs) failed, re-running.

Passing now

skitt (approved):

aswinsuryan (approved):

tpantelis (approved):

tpantelis:
The E2E for K8s 1.21 where we change the backing pods fails. `dig` returns duplicate IPs which doesn't affect verification of discovered IPs b/c we check for string containment but it does affect the case where we expect no IPs. I think doing https://github.com/submariner-io/lighthouse/issues/1322 should alleviate this issue. It seems the behavior of `EndpointSlices` in 1.21 is a bit different that later versions.

tpantelis:
> The E2E for K8s 1.21 where we change the backing pods fails. `dig` returns duplicate IPs which doesn't affect verification of discovered IPs b/c we check for string containment but it does affect the case where we expect no IPs. I think doing #1322 should alleviate this issue. It seems the behavior of `EndpointSlices` in 1.21 is a bit different that later versions.

It actually turned out to be a bug in the production code. https://github.com/submariner-io/lighthouse/pull/1457 fixes it.

PR #1463, opened 2023-12-20
Drop release-0.13 from dependabot config

skitt:
It's obsolete.


tpantelis:
Included with https://github.com/submariner-io/lighthouse/pull/1460

PR #1498, opened 2024-02-01
Use the Admiral metrics/profile helper and enable profiling

skitt:
The CoreDNS plugin uses the built-in profiling support.

Depends on https://github.com/submariner-io/admiral/pull/841

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -19,15 +19,12 @@ limitations under the License.
 package main
 
 import (
-	"errors"
 	"flag"
 	"fmt"
-	"net/http"
 	"os"
-	"time"
 
 	"github.com/kelseyhightower/envconfig"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/submariner-io/admiral/pkg/http"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/log/kzerolog"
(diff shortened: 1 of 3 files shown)


pkg/agent/main.go
```diff
@@ -147,33 +144,15 @@ func main() {
 	err = lightHouseAgent.Start(ctx.Done())
 	exitOnError(err, "Failed to start lighthouse agent")
 
-	httpServer := startHTTPServer()
+	defer http.StartServer(http.Metrics|http.Profile, 8082)()
```
tpantelis, line 147: Previously it just started the metrics handler (L170).

skitt: Yes, I forgot to mention that in the commit message — the goal of all this is to enable the profiling endpoints everywhere.

tpantelis: So it could be

```suggestion
	defer http.StartServer(http.All, 8082)()
```

but perhaps `All` is not safe if we add another endpoint and it's not desired by default. Maybe we should remove `All`.

skitt: That’s exactly why I didn’t use `All` here. I’ll change the Admiral PR to make `all` private.

vthapar (approved):

tpantelis (approved):

dfarrell07:
Do we want to backport this?

PR #1538, opened 2024-04-02
Use shipyard framework By instead of ginkgo By

tpantelis:
This alows flexibility to format or omit the output.

Related to https://github.com/submariner-io/subctl/issues/131

### DIFF
--- a/test/e2e/discovery/headless_services.go
+++ b/test/e2e/discovery/headless_services.go
@@ -72,17 +72,17 @@ func RunHeadlessDiscoveryTest(f *lhframework.Framework) {
 	clusterAName := framework.TestContext.ClusterIDs[framework.ClusterA]
 	clusterBName := framework.TestContext.ClusterIDs[framework.ClusterB]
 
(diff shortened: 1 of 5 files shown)


dfarrell07 (approved):

skitt (approved):
It might be worth noting that by default, tests run using `RunE2ETests` set `By` up so that it calls ginkgo’s `By`, so this doesn’t change existing behaviour.

tpantelis:
> It might be worth noting that by default, tests run using `RunE2ETests` set `By` up so that it calls ginkgo’s `By`, so this doesn’t change existing behaviour.

No  - this is for `subctl verify`.

skitt:
> > It might be worth noting that by default, tests run using `RunE2ETests` set `By` up so that it calls ginkgo’s `By`, so this doesn’t change existing behaviour.
> 
> No - this is for `subctl verify`.

Right, I get that the context of this change is `subctl verify`; but in each project where changes are being made, it also affects the E2E output. All I’m saying is that as far as E2E is concerned the change is innocuous because the framework sets `By` and `Fail` up to call the ginkgo functions in that scenario.

PR #1541, opened 2024-04-15
Update quic-go to 0.42.0 for CVE-2024-22189

dfarrell07:
Generated by:

go get github.com/quic-go/quic-go@v0.42.0
go get toolchain@none
go mod tidy


dfarrell07:
I'm not sure about the best way to handle the new-ish toolchain directive. It would be added by this bump, as this quic-go update requires golang 1.21 which adds the feature by default. However, the toolchain would be set to my local golang version, which seems pretty arbitrary. That would force go commands for this repo to use my version. It should happen seamlessly behind the scenes, but it still seems undesirable to set an arbitrary constraint.

https://go.dev/blog/toolchain
https://go.dev/doc/toolchain

skitt (approved):

tpantelis (approved):

PR #1551, opened 2024-04-30
Requeue ServiceExport when the global IP becomes available

tpantelis:
...rather than continuously re-queueing/retrying when it isn't available yet. This reduces the amount of re-queueing which can accumulate significant delays over time at scale due to the queue's bucket rate limiter. The `globalIngressIPCache` was adjusted to take a callback to be notified when the associated `GlobalIngressIP` is created or updated.


dfarrell07 (approved):

tpantelis:
@aswinsuryan

PR #1621, opened 2024-08-20
Propagate exported Service session affinity to ServiceImport

### LINKED ISSUE #1610: ServiceImports should inherit the SessionAffinity and SessionAffinityConfig from the exported Service
Referencing the MSC spec: https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#session-affinity.  We're not setting these fields on the ServiceImport.

tpantelis:
...to align with the MCS spec. If multiple clusters export a service, the fields are merged, ie for `SessionAffinity`, if any exported service has `ClientIP` then the `ServiceImport` will as well; for `SessionAffinityConfig`, the `ServiceImport` will inherit the first exported non-nil `SessionAffinityConfig`.

Fixes https://github.com/submariner-io/lighthouse/issues/1610


skitt (approved):

tpantelis:
@vthapar @skitt @aswinsuryan I wonder if we should report a conflict condition if all constituent cluster services don't specify the same `SessionAffinity`, ie if they don't all agree on `ClientIP`. Currently it doesn't really matter b/c we don't implement `SessionAffinity` at the clusterset service level (although Istio might). Which begs the next question, should we implement it in the core DNS plugin? 

For `SessionAffinityConfig`, we could be smarter and take the maximum ClientIP timeout value amongst the constituent services. I don't think we need to report a conflict condition if they don't all match.

vthapar (approved):

aswinsuryan (approved):

PR #1622, opened 2024-08-20
Fix sporadic failure in resolver migration unit test

tpantelis:
The round-robin test may not see cluster1's IP yet which causes it to fail, instead just wait until the resolver eventually returns the IP.


vthapar (approved):

skitt (approved):

dfarrell07:
@tpantelis thinks this was related to a version bump that may not have happened on other branches. Not sure if we need to backport, will keep an eye out for the failure on other branches.

PR #1624, opened 2024-08-22
Aggregate ServiceExport conflict conditions

tpantelis:
It's possible there could be more than one concurrent conflict however the MCS API only defines one conflict type so a second conflict would overwrite the first. We could define our own conflict types but the MCS spec does specifically reference the `ServiceExportConflict` type, defined as "_Conflict_", in the [conflict resolution](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#global-properties) section. It doesn't mention what the behavior should be if there's more than one conflict.
    
To represent multiple conflicts, we can still use the `ServiceExportConflict` type but aggregate the differing condition reasons and messages. For the `Reason` field, join multiple reasons separated by a comma and, for the `Message` field, join with a new line. When a particular conflict is resolved, remove its reason and message. When there's no more conflicts, ie `Reason` is empty, set the condition `Status` to `False`.

See individual commits for details.

### DIFF
--- a/pkg/agent/controller/service_export_client.go
+++ b/pkg/agent/controller/service_export_client.go
@@ -21,17 +21,21 @@ package controller
 import (
 	"context"
 	"reflect"
+	goslices "slices"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/lighthouse/pkg/constants"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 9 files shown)


pkg/agent/controller/service_export_client.go
```diff
@@ -104,6 +112,55 @@ func (c *ServiceExportClient) tryUpdateStatusConditions(ctx context.Context, nam
...
+	}
+
+	if ptr.Deref(to.Message, "") != "" {
+		messages = strings.Split(ptr.Deref(to.Message, ""), "\n")
+	}
+
+	index := goslices.Index(reasons, *from.Reason)
+	if index >= 0 {
+		if from.Status == corev1.ConditionTrue {
+			if index < len(messages) {
```
vthapar, line 129: When will this be true? Message is already present but not at end? Means this code is updating the message for this reason in case it changed?

tpantelis: `index` should always be < `len(messages)` b/c it's size mirrors `reason`. This is just a sanity check.

If you're referring to `if from.Status == corev1.ConditionTrue`, yes, if the `reason` already exists but the `message` changed.

dfarrell07 (approved):

vthapar (approved):

PR #1631, opened 2024-08-27
Address golangci-lint v1.60.3 errors

tpantelis:
Specifically, gosec "integer overflow conversion" issues.

Depends on https://github.com/submariner-io/shipyard/pull/1715

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -43,6 +43,7 @@ linters:
     - bidichk
     - bodyclose
     - contextcheck
+    - copyloopvar
     # - cyclop # This is equivalent to gocyclo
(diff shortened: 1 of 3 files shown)


skitt (changes requested):
There’s no Shipyard bump in the affected module (it doesn’t depend on Shipyard).

tpantelis:
> There’s no Shipyard bump in the affected module (it doesn’t depend on Shipyard).

yeah as it turned out. I had thought it would be affected by the Shipyard E2E changes but it wasn't. Still it doesn't hurt to bump it but I'll remove it.

dfarrell07 (approved):

skitt (approved):

PR #1632, opened 2024-08-30
Implement conflict checking for SessionAffinity

tpantelis:
The MCS spec's conflict resolution policy states that a "_conflict will be resolved by assigning precedence based on each ServiceExport's creationTimestamp, from oldest to newest_". We don't have a central MCS controller with access to all cluster's `ServiceExports` but we can store each cluster's `ServiceExport` `creationTimestamp` as annotations in the aggregated `ServiceImport` and use them to resolve conflicts. The `SessionAffinity` and `SessionAffinityConfig` fields on the aggregated `ServiceImport` will be set by the cluster with the oldest timestamp. The other clusters will set a `ServiceExportConflict` condition if their corresponding local service's fields do not match those on the aggregated `ServiceImport`. If a local service is updated in any cluster, each cluster re-evaluates the updated aggregated `ServiceImport` and either clears or sets the conflict condition. Also if the service from the precedent cluster is un-exported, the next precedent cluster will set the fields.

### DIFF
--- a/pkg/agent/controller/service_import.go
+++ b/pkg/agent/controller/service_import.go
@@ -21,6 +21,10 @@ package controller
 import (
 	"context"
 	"fmt"
+	"math"
+	"reflect"
+	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -36,12 +40,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/set"
 	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
+
+const timestampAnnotationPrefix = "timestamp.submariner.io/"
 
 //nolint:gocritic // (hugeParam) This function modifies syncerConf so we don't want to pass by pointer.
 func newServiceImportController(spec *AgentSpecification, syncerMetricNames AgentConfig, syncerConfig broker.SyncerConfig,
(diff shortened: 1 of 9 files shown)


pkg/agent/controller/service_import.go
```diff
@@ -442,3 +564,48 @@ func (c *ServiceImportController) localServiceImportLister(transform func(si *mc
 
 	return retList
 }
+
+func findClusterWithOldestTimestamp(from map[string]string) string {
```
vthapar, line 568: Am wondering if we can improve this logic. In a scale setup there might be too many clusters exporting a service and may cause issues loopin through all of them.

How about we also add an annotation like `timestamp.submariner.io/oldest` and use that to check if we're older or not. When the oldest one is exported, only then we will need to cycle through all to find the new oldest.

Also, this may have some issues when we're usin VIP. In case two different clusters export service at same time, VIP will end up being allocated by whosoever was first to successfully create Aggregated SI. It may happen that its timestamp is newer than the one that failed to create, coz it depends on timing of when the create attempt is made on broker.

vthapar (changes requested):
Tested and reviewed code locally. Other than thoughts on how to improve logic for oldest export, looks good.

pkg/agent/controller/service_import.go
tpantelis: > Am wondering if we can improve this logic. In a scale setup there might be too many clusters exporting a service and may cause issues loopin through all of them.

I think the time would be negligible with a single loop even with thousands. If it was a nested loop, then we could have an issue. With a single loop, it would take millions of iterations to start to become significant.

> 
> How about we also add an annotation like `timestamp.submariner.io/oldest` and use that to check if we're older or not. When the oldest one is exported, only then we will need to cycle through all to find the new oldest.

yeah we could do that, ie also if the oldest cluster is un-exported then remove the annotation to trigger other clusters to re-evaluate. But it adds more complexity that I'm not sure we really need. I can try it out....

> 
> Also, this may have some issues when we're usin VIP. In case two different clusters export service at same time, VIP will end up being allocated by whosoever was first to successfully create Aggregated SI. It may happen that its timestamp is newer than the one that failed to create, coz it depends on timing of when the create attempt is made on broker.

I wasn't planning on using the oldest timestamp for the VIP b/c we won't ever re-allocate it once set. The first cluster to successfully create the aggregated SI sets the VIP.

tpantelis: I did some testing of `findClusterWithOldestTimestamp` with varying number of entries in the annotations map:

- 1,000 entries took 38.793 µs
- 10,000 took 244.597 µs
- 1,000,000 took 46.399 ms
- 10,000,000 took 711.979 ms

I also tried calling `findClusterWithOldestTimestamp` in a nested loop. So calling it 1,000 times with 1,000 map entries took 21.924 ms (simulates 1000 services on 1000 clusters). With 10,000 and 10,000 it took 2.16 s, which is still pretty negligible.

So I think we should be fine the way it is.

skitt: Would it be worth adding those benchmarks as a test?

vthapar (approved):

pkg/agent/controller/service_import.go
tpantelis: What would we actually verify as a test?

This is really just normal looping of an array or map (with some fairly simple string manipulation/parsing) that I'm sure we do similarly in other places throughout the code base. These operations are fast with modern computers at the scales we'd be dealing with (in the thousands) so I don't think we need to be concerned. Only when you start getting into the millions is when you might need to be concerned.

tpantelis:
@aswinsuryan

aswinsuryan (approved):

dfarrell07:
The release notes are here: https://github.com/submariner-io/submariner-website/pull/1166

Related: https://github.com/submariner-io/lighthouse/pull/1621

PR #1641, opened 2024-09-17
Implement clusterset IP in the CoredDNS plugin

tpantelis:
Modify the resolver to return the `ServiceImport'`s IP if it's set and a specific cluster name wasn't requested. If the latter then return the cluster's DNS record info as it normally does.

### DIFF
--- a/coredns/resolver/service_info.go
+++ b/coredns/resolver/service_info.go
@@ -88,3 +88,7 @@ func (si *serviceInfo) selectIP(checkCluster func(string) bool) *DNSRecord {
 
 	return nil
 }
+
+func (si *serviceInfo) isHeadless() bool {
+	return si.spec.Type == mcsv1a1.Headless
+}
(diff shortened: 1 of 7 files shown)


dfarrell07 (approved):

coredns/resolver/service_info.go
```diff
@@ -88,3 +88,7 @@ func (si *serviceInfo) selectIP(checkCluster func(string) bool) *DNSRecord {
 
 	return nil
 }
+
+func (si *serviceInfo) isHeadless() bool {
```
yboaron, line 92: QQ: I know it's a small function, but performance-wise -  isn't it better to keep the current code ? I mean calculate it once when serviceInfo is created and not every time DNS resolver is called.

yboaron (approved):

coredns/resolver/service_info.go
tpantelis: I made it a method for readability and to avoid duplicate storage. It might be a little faster as a variable but the difference would be extremely negligible (probably in nanosec). Plus compilers are pretty smart about optimizing and caching frequently accessed code.

aswinsuryan (approved):

PR #1642, opened 2024-09-19
Add GHA to run MCS conformance tests

### LINKED ISSUE #1630: Add GHA to run the MCS conformance tests
This would identify and keep us aware of which aspects of the MCS spec where we may be non-compliant. Initially, we could just have the job run on-demand and eventually run on every PR once we're fully compliant.

tpantelis:
The GHA clones the **sigs.k8s.io/mcs-api** repo with the desired commit and runs `go test` directly in the directory. _v0.1.0_  doesn't have the conformance tests so we can't run them via Go import and trying to import a later commit is problematic due  incompatible changes to the MCS APIs. Also the conformance tests are really tied to the MCS spec and not specifically to the API version so they'll likely evolve separately anyway.

Fixes https://github.com/submariner-io/lighthouse/issues/1630


tpantelis:
The `ServiceImport` VIP test fails right now - we'll need to enable clusterset IP to pass it.

skitt (approved):

dfarrell07 (approved):

PR #1653, opened 2024-10-01
E2E: use ServiceDiscovery CR to determine if clusterset IP is enabled

vthapar:
...in E2E tests coz lighthouse agent may not be present if service-discovery is disabled.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -22,12 +22,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strconv"
 	"strings"
+	"sync/atomic"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/submariner-io/admiral/pkg/names"
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/lighthouse/pkg/constants"
@@ -67,7 +66,7 @@ var (
 	MCSClients          []*mcsClientset.Clientset
 	EndpointClients     []dynamic.ResourceInterface
 	SubmarinerClients   []dynamic.ResourceInterface
-	ClusterSetIPEnabled = false
+	clusterSetIPEnabled = atomic.Pointer[bool]{}
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


aswinsuryan (approved):

test/e2e/framework/framework.go
```diff
@@ -102,17 +100,20 @@ func beforeSuite() {
 
 	framework.DetectGlobalnet()
 
-	for _, k8sClient := range framework.KubeClients {
-		framework.AwaitUntil("find lighthouse agent deployment", func() (interface{}, error) {
-			return k8sClient.AppsV1().Deployments(framework.TestContext.SubmarinerNamespace).Get(context.TODO(),
-				names.ServiceDiscoveryComponent, metav1.GetOptions{})
+	for _, submClient := range SubmarinerClients {
+		framework.AwaitUntil("Get clustersetIP enabled Configuration", func() (interface{}, error) {
+			unstructuredSubmarinerConfig, err := submClient.Get(context.TODO(),
```
tpantelis: If only SD is installed than the `Submariner` CR isn't created so I think we should look for the `ServiceDiscovery` CR. Of course if SD isn't installed then that CR won't exist. So we can either Get the `ServiceDiscovery` CR without using use `AwaitUntil` and set `ClusterSetIPEnabled = if ServiceDiscovery found && ServiceDiscovery.Spec.ClusterSetIPEnabled` OR, preferably, lazily do this check from `BeforeEach`:

```
    var clusterSetIPEnabled = atomic.Pointer[bool]{}
    ...
    func IsClusterSetIPEnabled() bool {
	enabledPtr := clusterSetIPEnabled.Load()
	if enabledPtr != nil {
		return *enabledPtr
	}

	gvr, _ := schema.ParseResourceArg("servicediscoveries.v1alpha1.submariner.io")
	sdClient := framework.DynClients[0].Resource(*gvr).Namespace(framework.TestContext.SubmarinerNamespace)

	framework.AwaitUntil("find service-discovery resource", func() (interface{}, error) {
		return sdClient.Get(context.TODO(), "service-discovery", metav1.GetOptions{})
		enabled, _, err := unstructured.NestedBool(result.(*unstructured.Unstructured).Object, "spec", "clustersetIPEnabled")
		if err != nil {
			return false, "", err
		}

		clusterSetIPEnabled.Store(ptr.To(enabled))

		return true, "", nil
	})

	return *clusterSetIPEnabled.Load()
    }
```

tpantelis (changes requested):

test/e2e/framework/framework.go
tpantelis: I pushed this change.

vthapar: > set `ClusterSetIPEnabled = if ServiceDiscovery found && ServiceDiscovery.Spec.ClusterSetIPEnabled`

This works better.

vthapar (commented):
looks good, but I can't approve the review.

tpantelis (approved):

skitt (approved):

PR #1681, opened 2024-11-22
Fix SRV records not returned for clusterset IP service

tpantelis:
With clusterset IP enabled, we extract the ports from the aggregated `ServiceImport` instead of the `EndpointSlices`. However the ports are populated in the `ServiceImport` sometime after creation so `PutServiceImport` needs to update the cached spec if it already exists.

### DIFF
--- a/coredns/resolver/clusterip_service_test.go
+++ b/coredns/resolver/clusterip_service_test.go
@@ -450,9 +450,13 @@ func testClusterSetIP() {
 
 	BeforeEach(func() {
 		si := newAggregatedServiceImport(namespace1, service1)
-		si.Spec.IPs = []string{clusterSetIP}
-		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
 
+		si.Spec.IPs = []string{clusterSetIP}
+		t.resolver.PutServiceImport(si)
+
+		// We call PutServiceImport separately with the ports to ensure it updates the cache. This simulates the
+		// real behavior where the port information is added/updated in the ServiceImport some time after creation.
+		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
 		t.resolver.PutServiceImport(si)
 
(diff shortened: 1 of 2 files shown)


coredns/plugin/handler_test.go
```diff
@@ -975,10 +975,12 @@ func testClusterSetIP() {
 
 		si := newServiceImport(namespace1, service1, mcsv1a1.ClusterSetIP)
 		si.Spec.IPs = []string{clusterSetIP}
-		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
 
 		t.lh.Resolver.PutServiceImport(si)
 
+		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
+		t.lh.Resolver.PutServiceImport(si)
```
skitt: Could you add a comment here explaining why `PutServiceImport` needs to be called twice? I’m concerned that in a few months’ time someone might see this and remove the earlier `PutServiceImport` call since it appears redundant.

coredns/resolver/clusterip_service_test.go
```diff
@@ -450,9 +450,11 @@ func testClusterSetIP() {
 
 	BeforeEach(func() {
 		si := newAggregatedServiceImport(namespace1, service1)
+
 		si.Spec.IPs = []string{clusterSetIP}
-		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
+		t.resolver.PutServiceImport(si)
 
+		si.Spec.Ports = []mcsv1a1.ServicePort{port1, port2}
 		t.resolver.PutServiceImport(si)
```
skitt, line 460: Same comment here.

skitt (changes requested):

coredns/plugin/handler_test.go
tpantelis: I removed the two calls from here as the scenario is handled in _clusterip_service_test.go_.

skitt (approved):

vthapar (approved):

PR #1700, opened 2025-01-07
Adjust golangci-lint linters

tpantelis:
Added new linters (some commented out if not needed) and removed deprecated linters. Also addressed violations reported by new linters (in separate commits).
    
Related to https://github.com/submariner-io/enhancements/issues/231

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,29 +39,40 @@ linters-settings:
 linters:
   disable-all: true
   enable:
+    - asasalint
     - asciicheck
     - bidichk
     - bodyclose
+    # - canonicalheader # This is a slow linter and we don't use the net/http.Header API
+    - containedctx
     - contextcheck
     - copyloopvar
     # - cyclop # This is equivalent to gocyclo
+    - decorder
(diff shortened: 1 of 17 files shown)


skitt (approved):
I still would like to see “linter” commit descriptions that focus on the improvement to the project rather than the linter that flags the issues — these changes are worth doing because they are an improvement, not to fix a linter violation.

In this case the commit series would look like this:

```
Remove duplicate words ("on")

This was flagged by dupword.
```

```
Use resource.toJSON instead of json.MarshalIndent

This avoids having to check errors everywhere (flagged by errchkjson).
```

```
Use integer ranges in for loops

This was flagged by intrange.
```

```
Ignore maintidx on purposefully "unmaintainable" functions
```

```
Stop using named returns

This is now enforced by nonamedreturns.
```

```
Remove unnecessary formatting functions

* string concatenation instead of fmt.Sprintf
* errors.New instead of fmt.Errorf

This is now enforced by perfsprint.
```

Hopefully that helps illustrate why I think this approach is better. When reviewing the code history later, it’s more immediately apparent what the change is about.

PR #1757, opened 2025-04-07
Prevent duplicate LH EndpointSlices resources

tpantelis:
We observed a case in QE where there were two LH EPS resources that were created less than a second apart for the same ClusterIP service. The intention is that there's a single EPS created using the GenerateName functionality. As such, when updating the resource, we don't know the generated name so we look it up by its identifying labels via CreateOrUpdate. This expects that there's only one resource with the identifying labels and fails otherwise.

Since there's a single ServiceEndpointSliceController per service and it has a single-threaded work queue, there should never be two EPS resources. The only possible scenario that I can see is if there were two ServiceEndpointSliceController instances running concurrently in the small window between stopping the current controller and starting a new one when a service is updated. When stopping, we drain the work queue which will wait for all queued items to complete processing so we shouldn't have two instances running concurrently. However, the drain has a 5 sec deadline so, if it times out, there may still be a task being processed. So it's possible there was a current create-or-update operation that was delayed for more than 5 sec and remained running long enough when the new controller instance was started, resulting in both creating an EPS resource. I developed a unit test that reproduces this scenario.

To close this window, we need the previous controller instance to be completely stopped before we start a new instance. So when calling AwaitStopped on the resource syncer, pass in a timed Context and propagate the returned error so the operation is retried.

Depends on https://github.com/submariner-io/admiral/pull/1101


aswinsuryan (approved):

pkg/agent/controller/service_import.go
```diff
@@ -267,18 +271,24 @@ func (c *ServiceImportController) reconcileLocalAggregatedServiceImports() {
...
 
-	if obj, found := c.endpointControllers.LoadAndDelete(key); found {
-		logger.V(log.DEBUG).Infof("Stopping previous endpoints controller for %q", key)
-		obj.(*ServiceEndpointSliceController).stop()
+	if obj, found := c.endpointControllers.Load(key); found {
+		logger.V(log.DEBUG).Infof("Stopping previous EndpointSlice controller for %q", key)
+
+		err := obj.(*ServiceEndpointSliceController).stop(ctx)
+		if err != nil {
+			return errors.Wrapf(err, "failed to stop previous EndpointSlice controller for %q", key)
```
vthapar, line 282: In what scenarios this can fail? Is it safe to delete if failed to stop previous? Any cleanup missed?

coredns/gateway/controller.go
```diff
@@ -112,7 +112,7 @@ func (c *Controller) Start(client dynamic.Interface) error {
 		return errors.New("failed to wait for informer cache to sync")
 	}
 
-	go c.queue.Run(c.stopCh, c.processNextGateway)
+	go c.queue.Run(c.processNextGateway)
```
vthapar, line 115: Would've preferred this to be different commit as it is a bit unintutive why we have change in gateway controller for endpointslices. Not an issue, just was confused a bit by this.

vthapar (commented):
Look good, just one question about changes.

coredns/gateway/controller.go
tpantelis: yeah sorry this was due to the change in the admiral PR to remove the `stopCh` param. This was another API change made by that PR.

pkg/agent/controller/service_import.go
tpantelis: > In what scenarios this can fail?

If `stop` times out while awaiting the syncer queue to drain. It may mean it's stuck trying to create an EPS. We propagate the error so it is retried and give the stuck process a chance to complete, successfully or not.

> Is it safe to delete if failed to stop previous?

I'm not clear - safe to delete what?

> Any cleanup missed?

There's no more cleanup here.

PR #1758, opened 2025-04-07
[0.20] Bump oauth2 for CVE-2025-22868

dfarrell07:
Both module bumps generated by

  go get golang.org/x/oauth2@v0.27.0
  go mod tidy


tpantelis:
We really could just do the bump in admiral and let it trickle to the other projects when we do the next release.

tpantelis:
This will get picked up from shipyard on the next release.

PR #1805, opened 2025-06-09
Bump sigs.k8s.io/mcs-api version and adjust to changes

tpantelis:
...to match the version in https://github.com/coredns/coredns v1.12.2 so we can bump to the latter fo fix a CVE.

The breaking change in the MCS API is to use the standard `metav1.Condition` in `ServiceExport` Status instead of `ServiceExportCondition`. The unstructured serialized state doesn't change as the fields are compatible so no migration is needed. However `metav1.Condition` does make some fields required in the Go struct but that won't affect us since the deployed CRD doesn't list them as required.


dfarrell07 (approved):

dfarrell07:
```
coredns/plugin/setup.go:62:19: SA1019: mcsv1a1.AddToScheme is deprecated: use Install instead (staticcheck)
	utilruntime.Must(mcsv1a1.AddToScheme(scheme.Scheme))
	                 ^
coredns/resolver/resolver_suite_test.go:123:9: SA1019: mcsv1a1.AddToScheme is deprecated: use Install instead (staticcheck)
	Expect(mcsv1a1.AddToScheme(scheme.Scheme)).To(Succeed())
	       ^
```

And the [MCS Conformance](https://github.com/submariner-io/lighthouse/actions/runs/15539624366/job/43747283419?pr=1805#logs) job is failing.

dfarrell07:
The vuln scanning job is currently incorrectly failing due to a GH vuln DB bug. Discussion: https://kubernetes.slack.com/archives/C010RJV694M/p1749485243351409

PR #1807, opened 2025-06-10
Bump github.com/coredns/coredns to 1.12.2

tpantelis:
...to address Vulnerability GHSA-cvx7-x8pj-x2gw.


skitt (approved):

coredns/go.mod
```diff
@@ -4,85 +4,101 @@ go 1.23.0
...
 	k8s.io/client-go v0.32.4
 	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	sigs.k8s.io/controller-runtime v0.20.1
-	sigs.k8s.io/mcs-api v0.1.0
+	sigs.k8s.io/mcs-api v0.1.1-0.20250224121229-6c631f4730d0
 )
 
+// This is pulled in by github.com/coredns/coredns but we pin to original version
+// since the newer version is incompatible with lighthouse code.
+replace sigs.k8s.io/mcs-api v0.1.1-0.20250224121229-6c631f4730d0 => sigs.k8s.io/mcs-api v0.1.0
```
skitt, line 24: Nice idea but it seems like CoreDNS really needs the newer version :-/.

tpantelis: yup 😦

tpantelis: We can backport https://github.com/submariner-io/lighthouse/pull/1805 or just ignore the CVE.

skitt: If we make sure that QUIC isn’t enabled in the Lighthouse CoreDNS deployment then the CVE can be ignored. Backporting #1805 would be OK too IMO.

tpantelis: According to https://github.com/submariner-io/lighthouse/security/code-scanning/68, the core DNS CVE was detected in v1.12.1 but 0.20 uses version v1.12.0 so I don’t think we need to bump the version.

tpantelis: We don't enable QUIC in our plugin.

PR #1812, opened 2025-06-11
Ignore github.com/coredns/coredns CVE-2025-47950

tpantelis:

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -10,3 +10,7 @@ ignore:
   - vulnerability: CVE-2021-22570
     package:
       name: google.golang.org/protobuf
+  # We can ignore this CVE since it's related to QUIC which we don't use.
+  - vulnerability: GHSA-cvx7-x8pj-x2gw
+    package:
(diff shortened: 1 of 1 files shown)


.grype.yaml
```diff
@@ -10,3 +10,7 @@ ignore:
   - vulnerability: CVE-2021-22570
     package:
       name: google.golang.org/protobuf
+  # We can ignore this CVE since it's related to QUIC which we don't use.
+  - vulnerability: GHSA-cvx7-x8pj-x2gw
```
skitt, line 14:
```suggestion
  - vulnerability: CVE-2025-47950
```

(it’s strange to have a comment mentioning “CVE” and then a GHSA)

tpantelis: Using the CVE number doesn't work. I'm not sure of the difference between the two.

skitt: Ah, OK!

skitt (approved):

vthapar (approved):

PR #1829, opened 2025-07-01
Bump the CoreDNS dependency to address CVEs

skitt:
This fixes CVE-2024-0874 and CVE-2025-47950. This involves bumping k8s dependencies too, but only in a submodule, so this won't have an impact on the rest of the project.


tpantelis:
I didn't think we were planning to do another 0.16 release...

PR #1880, opened 2025-08-06
Update Tekton rpms-signature-scan, fixes EC

dfarrell07:


dfarrell07:
This still shows failures (very confusingly) because each job doesn't see the fixes in the other and the results are overall.

tpantelis (approved):

PR #1888, opened 2025-08-18
Handle empty string in stripUnderscore to avoid panic

tpantelis:
...by using `strings.TrimPrefix`.

This was identified by the Cursor AI tool.

### DIFF
--- a/coredns/plugin/parse.go
+++ b/coredns/plugin/parse.go
@@ -19,6 +19,8 @@ limitations under the License.
 package lighthouse
 
 import (
+	"strings"
+
 	"github.com/coredns/coredns/plugin/pkg/dnsutil"
 	"github.com/coredns/coredns/request"
 	"github.com/miekg/dns"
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

coredns/plugin/parse.go
```diff
@@ -151,9 +153,5 @@ func parseSegments(segs []string, count int, r *recordRequest, qType uint16) (*r
 
 // stripUnderscore removes a prefixed underscore from s.
 func stripUnderscore(s string) string {
-	if s[0] != '_' {
-		return s
-	}
-
-	return s[1:]
+	return strings.TrimLeft(s, "_")
```
skitt:
```suggestion
	return strings.TrimPrefix(s, "_")
```

to match the existing behaviour (except with empty strings).

skitt (changes requested):

skitt (changes requested):
The commit message needs to be updated too ;-). (Perhaps with an indication that Cursor’s suggestion was incorrect…)

skitt (approved):

PR #1945, opened 2025-11-04
Use strings.Builder instead of concatenation in loop

skitt:
This was flagged by the perfsprint linter.

See https://github.com/submariner-io/shipyard/pull/2187 and the comment there — this can wait until after v0.22 rc0.


tpantelis:
I fixed this locally as well. I've also updated the linters (eg `wsl` is deprecated in favor of `wsl_v5`).

dfarrell07 (approved):

tpantelis (approved):

PR #1947, opened 2025-11-05
Make TTL configurable via global ConfigMap

tpantelis:


coredns/plugin/lighthouse.go
```diff
@@ -92,5 +94,11 @@ func (lh *Lighthouse) configure(c *caddy.Controller) error {
 		}
 	}
 
+	lh.TTL = global.Get(TTLConfigKey, lh.TTL)
```
vthapar, line 97: What will this return if nothing configured? Don't want `lh.TTL` being overwritten with 0 if not configured in global ConfigMap?

coredns/plugin/setup_internal_test.go
```diff
@@ -159,7 +159,17 @@ func testCorrectConfig() {
 		})
 	})
 
-	It("Should handle missing optional fields", func() {
+	When("and the global config ttl setting is specified", func() {
```
vthapar, line 162: Add a test to confirm if global config tll is not specified, it uses default of 5.

vthapar (changes requested):

coredns/plugin/setup_internal_test.go
tpantelis: There already is: https://github.com/submariner-io/lighthouse/blob/91b85c7d820717a6db71af28ad89b47e11bb4a47/coredns/plugin/setup_internal_test.go#L172

coredns/plugin/lighthouse.go
tpantelis: It returns the specified default value `lh.TTL`. So if global setting not configured, it reduces to `lh.TTL = lh.TTL `.

vthapar (approved):

coredns/plugin/setup_internal_test.go
vthapar: Ah, should've expanded the file.

skitt (approved):

Issue #2197, opened 2026-06-04
Opt-in reflection of imported services into cluster.local (for clusters with non-configurable DNS, e.g. EKS Auto Mode)


kekkker:
## Problem

Lighthouse resolves `*.clusterset.local` by injecting a forward stanza into the cluster's CoreDNS `ConfigMap`:

```
clusterset.local:53 {
    forward . <lighthouse-coredns-cluster-ip>
}
```

This requires that (a) the cluster DNS config is editable and (b) the resolver the pods actually use loads it. A growing class of managed platforms breaks this:

- **EKS Auto Mode** bakes a **node-local CoreDNS** into every node, bound to the kube-dns Service VIP, with an Auto-Mode-managed **immutable Corefile that does not read the `kube-system/coredns` ConfigMap**. The lighthouse stanza is written but never reaches the resolver answering the VIP, so `clusterset.local` queries get forwarded upstream and `NXDOMAIN`.
  - Confirmed in the field: the in-cluster CoreDNS *pods* resolve `clusterset.local` correctly when queried directly by pod IP; only the node-local VIP path (what pods actually use) is broken.
- Similar lockdown is appearing on other managed/auto node runtimes.

On these platforms Lighthouse has **no working integration point**, and operators fall back to side-channels (a second CoreDNS Service + per-pod `dnsConfig` pointing at it) — fragile and per-workload.

## Proposal

An **opt-in** agent mode (`SUBMARINER_REFLECT_CLUSTER_LOCAL`, default `false`) that reflects each imported service into a native `cluster.local` `Service` + `EndpointSlice` in the consuming cluster, named **`<svc>-<sourceCluster>`** (e.g. `mysvc-clusterB`), backed by the addresses Submariner's data plane already routes.

Workloads then resolve `<svc>-<cluster>.<ns>.svc.cluster.local` through the **standard resolver** — no DNS config required, no second CoreDNS, no per-pod `dnsConfig`. `clusterset.local` stays the default and spec-conformant path; this runs alongside, only where enabled.

## Why it fits cleanly

The agent **already** suffixes imported EndpointSlices with the source cluster (`endpoint_slice.go` `isLegacyEndpointSlice` keys on `-<sourceCluster>`) and labels them `LabelSourceCluster` / `LabelServiceName` / `lighthouse.submariner.io/sourceNamespace`. A reflector just consumes objects the agent already produces — a small, idiomatic controller, not a new subsystem.

[...]

PR #2198, opened 2026-06-04
Add opt-in cluster.local reflection of imported services

kekkker:
Implements #2197.

## Problem
Lighthouse integrates DNS by injecting a `clusterset.local` forward stanza into the cluster's CoreDNS `ConfigMap`. That fails where DNS isn't configurable — notably **EKS Auto Mode**, which bakes an immutable node-local CoreDNS into every node that ignores the `kube-system/coredns` ConfigMap, so `clusterset.local` names `NXDOMAIN`. (Verified in the field: the in-cluster CoreDNS *pods* resolve fine by pod IP; only the node-local VIP path is broken.) The only workaround today is a side-channel CoreDNS + per-pod `dnsConfig`.

## Approach
Opt-in agent mode `SUBMARINER_REFLECT_CLUSTER_LOCAL` (default `false`). When enabled, the agent reflects each imported service into a native `cluster.local` Service + EndpointSlice named **`<service>-<sourceCluster>`** in the service namespace, copied from the imported EndpointSlice the agent already syncs. Workloads then resolve `<svc>-<cluster>.<ns>.svc.cluster.local` via the standard resolver — no second CoreDNS, no per-pod `dnsConfig`. `clusterset.local` stays the default, spec-conformant path; reflection runs alongside it only where enabled.

The `<svc>-<cluster>` naming reuses the agent's existing per-cluster slice convention, keeps it collision-safe (distinct names per exporting cluster; never squats a bare local name) and conformant-by-construction. The reflector is collision-guarded (won't overwrite a Service it doesn't own) and prunes the reflected Service when the last imported slice for a (service, cluster) pair is removed.

## Changes
- `ReflectClusterLocal` flag on `AgentSpecification`, wired into the agent (gated, default-off).
- `ClusterLocalReflector` controller — watches imported EndpointSlices, reflects only those whose source cluster differs from the local cluster.
- Unit tests for the reflection transforms + two-cluster integration specs (reflect / prune / disabled-noop).
- Design proposal: `docs/proposals/clusterlocal-reflection.md`.

## Testing
- `go build`, `go vet`, and the full `pkg/agent/controller` suite (incl. the new specs) pass locally.
- **Validated end-to-end on a live EKS Auto Mode cluster** (peered with a Hetzner cluster over Submariner): with the flag on, all imports reflected to `<svc>-<cluster>.<ns>.svc.cluster.local` and resolved via the node-local resolver — the exact path that `NXDOMAIN`s `clusterset.local`.

## Follow-up (separate submariner-operator PR)
- A `serviceDiscovery.reflectClusterLocal` value to set the env var.
- Extend the lighthouse-agent ServiceAccount RBAC with `services` and `endpointslices` create/update/delete (the reflector creates those).

A standalone controller implementing the same logic has been running in production on an EKS Auto Mode ↔ Hetzner clusterset, fully replacing the side-channel; this upstreams it as a supported switch. Design rationale and trade-offs (incl. why this isn't the MCS default) are in the proposal doc.

## Summary by CodeRabbit

* **New Features**
  * Optional cluster-local service reflection: when enabled, imported services from other clusters are reflected as native cluster.local headless Services plus corresponding EndpointSlices, with ownership and name validation to avoid collisions.
  * Controller startup now respects the cluster-local reflector setting and surfaces initialization/start failures.

* **Tests**
  * Added tests validating enabled/disabled behavior, reflected object contents, and eventual pruning of reflected resources.

### DIFF
--- a/pkg/agent/controller/clusterlocal_reflector.go
+++ b/pkg/agent/controller/clusterlocal_reflector.go
@@ -1,0 +1,338 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/syncer/broker"
+	"github.com/submariner-io/admiral/pkg/watcher"
+	"github.com/submariner-io/admiral/pkg/workqueue"
+	"github.com/submariner-io/lighthouse/pkg/constants"
+	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	validation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/ptr"
+	mcsv1b1 "sigs.k8s.io/mcs-api/pkg/apis/v1beta1"
+)
+
+const (
+	// ReflectorManagedByLabel/Value mark the cluster.local Service and EndpointSlice
+	// objects the reflector owns. Only objects carrying this label are ever created,
+	// updated or deleted by it.
+	ReflectorManagedByLabel = "app.kubernetes.io/managed-by"
+	ReflectorManagedByValue = "lighthouse-clusterlocal-reflector"
+
+	// reflectedFromLabel records the imported EndpointSlice a reflected EndpointSlice
+	// was derived from (1:1). It is also the federator's identifying label, so the
+	// reflected slice (created with GenerateName) is matched on update and delete.
+	reflectedFromLabel = "lighthouse.submariner.io/reflected-from"
+)
+
+// ServiceGVR is the GroupVersionResource of the Services the reflector creates.
+// (endpointSliceGVR is declared in cleanup.go.)
+var ServiceGVR = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
+
+// ClusterLocalReflector reflects each service imported from another cluster into a
+// native cluster.local Service + EndpointSlice named "<service>-<sourceCluster>" in
+// the service's namespace, so workloads can resolve cross-cluster services through
+// the standard cluster DNS path. This is for clusters whose DNS cannot host the
+// clusterset.local forward stanza (e.g. EKS Auto Mode's immutable node-local
+// CoreDNS). It is instantiated only when AgentSpecification.ReflectClusterLocal is
+// set; default behaviour (clusterset.local only) is unchanged.
+//
+// It consumes the imported EndpointSlices the agent already syncs to the local
+// service namespace (labelled discovery.LabelManagedBy=LabelValueManagedBy, with
+// LabelServiceName/LabelSourceCluster/LabelSourceNamespace), reflecting only those
+// whose source cluster is not this cluster.
+type ClusterLocalReflector struct {
+	clusterID           string
+	syncer              syncer.Interface
+	serviceWatcher      watcher.Interface
+	federator           federate.Federator
+	serviceClient       dynamic.NamespaceableResourceInterface
+	endpointSliceClient dynamic.NamespaceableResourceInterface
+}
+
+//nolint:gocritic // (hugeParam) matches the other controller constructors.
+func newClusterLocalReflector(spec *AgentSpecification, syncerConfig broker.SyncerConfig) (*ClusterLocalReflector, error) {
+	c := &ClusterLocalReflector{
+		clusterID:           spec.ClusterID,
+		serviceClient:       syncerConfig.LocalClient.Resource(ServiceGVR),
+		endpointSliceClient: syncerConfig.LocalClient.Resource(endpointSliceGVR),
+	}
+
+	// Empty TargetNamespace => the federator uses each object's own namespace.
+	// IdentifyingLabels let the federator find the reflected EndpointSlice (which
+	// uses GenerateName) on update and delete; the deterministically-named Service
+	// is matched by name and ignores them.
+	c.federator = federate.NewCreateOrUpdateFederator(federate.CreateOrUpdateOptions{
+		Client:            syncerConfig.LocalClient,
+		RestMapper:        syncerConfig.RestMapper,
+		TargetNamespace:   metav1.NamespaceAll,
+		IdentifyingLabels: []string{ReflectorManagedByLabel, reflectedFromLabel},
+	})
+
+	var err error
+
+	c.syncer, err = syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
+		Name:            "ClusterLocal reflector",
+		SourceClient:    syncerConfig.LocalClient,
+		SourceNamespace: metav1.NamespaceAll,
+		SourceLabelSelector: k8slabels.SelectorFromSet(map[string]string{
+			discovery.LabelManagedBy: constants.LabelValueManagedBy,
+		}).String(),
+		RestMapper:      syncerConfig.RestMapper,
+		Federator:       c.federator,
+		ResourceType:    &discovery.EndpointSlice{},
+		Transform:       c.reflect,
+		Scheme:          syncerConfig.Scheme,
+		WorkQueueConfig: workqueue.ConfigFromGlobal("clusterlocal-reflector", nil),
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating ClusterLocal reflector syncer")
+	}
+
+	// Watch the reflected Services so that, when one is deleted (or edited so it no
+	// longer carries our managed-by label and thus drops out of this watch), the
+	// now-orphaned reflected EndpointSlices are cleaned up. The EndpointSlice syncer
+	// alone wouldn't notice this, as it only watches imported EndpointSlices.
+	c.serviceWatcher, err = watcher.New(&watcher.Config{
+		Client:     syncerConfig.LocalClient,
+		RestMapper: syncerConfig.RestMapper,
+		Scheme:     syncerConfig.Scheme,
+		ResourceConfigs: []watcher.ResourceConfig{
+			{
+				Name:            "ClusterLocal reflected Service",
+				ResourceType:    &corev1.Service{},
+				SourceNamespace: metav1.NamespaceAll,
+				SourceLabelSelector: k8slabels.SelectorFromSet(map[string]string{
+					ReflectorManagedByLabel: ReflectorManagedByValue,
+				}).String(),
+				Handler: watcher.EventHandlerFuncs{
+					OnDeleteFunc: c.onReflectedServiceDeleted,
+				},
+			},
+		},
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating ClusterLocal reflected Service watcher")
+	}
+
+	return c, nil
+}
+
+func (c *ClusterLocalReflector) start(stopCh <-chan struct{}) error {
+	if err := c.syncer.Start(stopCh); err != nil {
+		return errors.Wrap(err, "error starting ClusterLocal reflector syncer")
+	}
+
+	if err := c.serviceWatcher.Start(stopCh); err != nil {
+		return errors.Wrap(err, "error starting ClusterLocal reflected Service watcher")
+	}
+
+	return nil
+}
+
+// onReflectedServiceDeleted deletes the reflected EndpointSlices belonging to a
+// reflected Service that was deleted or edited out of reflector ownership.
+func (c *ClusterLocalReflector) onReflectedServiceDeleted(obj runtime.Object, _ int) bool {
+	svc := obj.(*corev1.Service)
+
+	selector := k8slabels.SelectorFromSet(map[string]string{
+		discovery.LabelServiceName: svc.Name,
+		ReflectorManagedByLabel:    ReflectorManagedByValue,
+	}).String()
+
+	err := c.endpointSliceClient.Namespace(svc.Namespace).DeleteCollection(context.TODO(),
+		metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: selector})
+	if err != nil && !apierrors.IsNotFound(err) {
+		logger.Errorf(err, "Error deleting reflected EndpointSlices for Service %s/%s", svc.Namespace, svc.Name)
+		return true
+	}
+
+	logger.Infof("Cleaned up reflected EndpointSlices for removed Service %s/%s", svc.Namespace, svc.Name)
+
+	return false
+}
+
+// reflect is the syncer Transform. The returned EndpointSlice is created/updated/
+// deleted by the syncer's federator; the matching Service is reconciled here.
+func (c *ClusterLocalReflector) reflect(obj runtime.Object, _ int, op syncer.Operation) (runtime.Object, bool) {
+	imported := obj.(*discovery.EndpointSlice)
+
+	serviceName := imported.Labels[mcsv1b1.LabelServiceName]
+	sourceCluster := imported.Labels[mcsv1b1.LabelSourceCluster]
+	serviceNamespace := imported.Labels[constants.LabelSourceNamespace]
+
+	// Reflect only services imported from *other* clusters. Our own exported
+	// services already exist locally under their real cluster.local name.
+	if serviceName == "" || serviceNamespace == "" || sourceCluster == "" || sourceCluster == c.clusterID {
+		return nil, false
+	}
+
+	reflectedName := serviceName + "-" + sourceCluster
+
+	// The reflected name must be a valid Service name (DNS-1123 label, <=63 chars).
+	// An over-long service/cluster combination can't be reflected; skip and log
+	// rather than failing on every reconcile.
+	if errs := validation.IsDNS1123Label(reflectedName); len(errs) > 0 {
+		logger.Warningf("Cannot reflect imported service %q from cluster %q: %q is not a valid Service name: %v",
+			serviceName, sourceCluster, reflectedName, errs)
+
+		return nil, false
+	}
+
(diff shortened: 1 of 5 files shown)


docs/proposals/clusterlocal-reflection.md
```diff
@@ -0,0 +1,146 @@
+# Proposal: Opt-in reflection of imported services into `cluster.local`
```
tpantelis: Please push a PR to put this file in https://github.com/submariner-io/enhancements.

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,289 @@
...
+		if p.Protocol != nil {
+			sp.Protocol = *p.Protocol
+		}
+
+		if p.Port != nil {
+			sp.Port = *p.Port
+			sp.TargetPort = intstr.FromInt32(*p.Port)
+		}
+
+		ports = append(ports, sp)
```
tpantelis: You can simplify this to:


```suggestion
		p := &in[i]
		ports = append(ports, corev1.ServicePort{
		    Protocol:    ptr.Deref(p.Protocol, corev1.ProtocolTCP),
		    Name:        ptr.Deref(p.Name, ""),
		    Port:           ptr.Deref(p.Port, 0),
		    TargetPort: intstr.FromInt32(ptr.Deref(p.Port, 0)), })

```

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,289 @@
...
+
+func reflectEndpoints(in []discovery.Endpoint) []discovery.Endpoint {
+	out := make([]discovery.Endpoint, 0, len(in))
+	ready := true
+
+	for i := range in {
+		e := in[i]
+		if e.Conditions.Ready == nil {
+			e.Conditions.Ready = &ready
+		}
```
tpantelis: Why are you converting a nil `Ready` to true? The docs for `Ready` state that "_A nil value should be interpreted as "true"_" so this seems unnecessary. You should just use `in` directly and eliminate this function.

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+	discovery "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
+	mcsv1b1 "sigs.k8s.io/mcs-api/pkg/apis/v1beta1"
+)
+
+const reflectorManagedByValue = "lighthouse-clusterlocal-reflector"
```
tpantelis: Make the controller constant public instead of duplicating it.

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,289 @@
...
+	// reflectedFromLabel records the imported EndpointSlice a reflected EndpointSlice
+	// was derived from (1:1, for lifecycle and debugging).
+	reflectedFromLabel = "lighthouse.submariner.io/reflected-from"
+
+	// reflectedSliceSuffix is appended to the imported slice name to form the
+	// reflected slice name (kept in the same namespace, so it must differ).
+	reflectedSliceSuffix = "-cl"
+)
+
+var serviceGVR = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
```
tpantelis: Make this public so unit tests can use it.


```suggestion
var ServiceGVR = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+		return svc
+	}
+
+	When("reflection is enabled on the consuming cluster", func() {
+		BeforeEach(func() {
+			t.cluster2.agentSpec.ReflectClusterLocal = true
+		})
+
+		It("creates a native cluster.local Service and EndpointSlice for the import", func(ctx context.Context) {
```
tpantelis:
```suggestion
		It("should create a native cluster.local Service and EndpointSlice for the import", func(ctx context.Context) {
```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+			exportFromCluster1(ctx)
+
+			By("Verifying the reflected headless Service")
+
+			svc := awaitReflectedService(ctx)
+			Expect(svc.Spec.ClusterIP).To(Equal(corev1.ClusterIPNone))
+			Expect(svc.Labels).To(HaveKeyWithValue("app.kubernetes.io/managed-by", reflectorManagedByValue))
```
tpantelis: Use the "app.kubernetes.io/managed-by" constant from `controller` (in all places).

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+				continue
+			}
+
+			eps := &discovery.EndpointSlice{}
+			Expect(scheme.Scheme.Convert(&list.Items[i], eps, nil)).To(Succeed())
+			out = append(out, eps)
+		}
+
+		return out
+	}
```
tpantelis: `findReflectedSlices` and `findImportedSlices` are rather large for function literals. However there is an existing named method, `findEndpointSlices`, that we can reuse, with some mods:

```Go
func findEndpointSlices(ctx context.Context, client dynamic.ResourceInterface, namespace, name, clusterID string, ) []*discovery.EndpointSlice { 	
    ...
    for i := range list.Items { 		
        ...
        nameMatch := name == "" || labels[mcsv1b1.LabelServiceName] ==    name || labels[discovery.LabelServiceName] == name 
        namespaceMatch := namespace == "" || labels[constants.LabelSourceNamespace] == namespace 
        clusterMatch := clusterID == "" || labels[mcsv1b1.LabelSourceCluster] == clusterID
        if nameMatch && namespaceMatch && clusterMatch { 		
             ...
         } 
    }
  
    return endpointSlices 
}
```

Then:

`findEndpointSlices(ctx, t.cluster2.localEndpointSliceClient, "", reflectedName, "")`

`findEndpointSlices(ctx, t.cluster2.localEndpointSliceClient, "", "", clusterID1)`

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+			for _, s := range slices {
+				for i := range s.Endpoints {
+					addrs = append(addrs, s.Endpoints[i].Addresses...)
+				}
+			}
+
+			// For a ClusterSetIP import, the reflected slice exposes the clusterset
+			// VIP (which Submariner routes), not the remote pod IPs.
+			Expect(addrs).To(ContainElement(ipV4ServiceIP1))
```
tpantelis: You can simplify this:


```suggestion
			Eventually(func(g Gomega, ctx context.Context) {
			    slices := findEndpointSlices(ctx, t.cluster2.localEndpointSliceClient, "", reflectedName, "") 	
			    g.Expect(slices).NotTo(BeEmpty()) 	  
			    g.Expect(slices[0].Labels).To(HaveKeyWithValue(
			          "app.kubernetes.io/managed-by", reflectorManagedByValue))
			    g.Expect(slices[0].Endpoints).To(
			         ContainElement(HaveField("Addresses", 
			              ContainElement(ipV4ServiceIP1)))) }).Within(
			                     5 * time.Second).WithContext(ctx).Should(Succeed())

```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+		})
+
+		It("prunes the reflected Service and EndpointSlice when the export is removed", func(ctx context.Context) {
```
tpantelis:
```suggestion
		It("should prune the reflected Service and EndpointSlice when the export is removed", func(ctx context.Context) {
```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+			t.cluster1.deleteEndpointSlice(ctx, t.cluster1.serviceEndpointSlices[0].Name)
+
+			By("Verifying the reflected Service is removed")
+
+			Eventually(func(ctx context.Context) bool {
+				_, err := svcClient.Get(ctx, reflectedName, metav1.GetOptions{})
+				return apierrors.IsNotFound(err)
+			}).WithContext(ctx).Should(BeTrue())
+
+			Expect(findReflectedSlices(ctx)).To(BeEmpty())
```
tpantelis:
```suggestion
			Eventually(func(g Gomega, ctx context.Context) bool {
				_, err := svcClient.Get(ctx, reflectedName, metav1.GetOptions{})
				g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
				g.Expect(findEndpointSlices(ctx,
				     t.cluster2.localEndpointSliceClient, "", 
				     reflectedName, "")).To(BeEmpty())
			}).WithContext(ctx).Should(Succeed())
```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+		})
+	})
+
+	When("reflection is disabled (default)", func() {
+		It("does not create any cluster.local Service for the import", func(ctx context.Context) {
```
tpantelis:
```suggestion
		It("should not create any cluster.local Service for the import", func(ctx context.Context) {
```

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+			By("Ensuring no reflected Service is created")
+
+			Consistently(func(ctx context.Context) bool {
+				_, err := svcClient.Get(ctx, reflectedName, metav1.GetOptions{})
+				return apierrors.IsNotFound(err)
+			}).WithContext(ctx).WithTimeout(2 * time.Second).Should(BeTrue())
+
+			Expect(findReflectedSlices(ctx)).To(BeEmpty())
+		})
```
tpantelis: Simplify, also 1 sec is enough time. 

```suggestion
			Consistently(func(g Gomega, ctx context.Context) bool {
				_, err := svcClient.Get(ctx, reflectedName, metav1.GetOptions{})
				g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
				g.Expect(findReflectedSlices(ctx)).To(BeEmpty())
			}).WithContext(ctx).WithTimeout(time.Second).Should(Succeed())
		})
```

go.work
```diff
@@ -0,0 +1,6 @@
+go 1.26.3
```
tpantelis: We don't submit `go.work` and `go.sum` files.

tpantelis (changes requested):
It isn't necessary to push a new commit when addressing review comments. Plus we don't want to merge review commits as their context is only relevant to the PR. Please squash into one commit.

kekkker:
Thanks for the review @tpantelis! Addressed everything and squashed to a single commit:

- **Squashed** the 4 commits into one.
- Removed the accidentally-committed `go.work`/`go.work.sum`.
- `servicePortsFrom`: simplified with `ptr.Deref`.
- Dropped `reflectEndpoints` — nil `Ready` already means true, so the imported endpoints are used directly.
- Made `ServiceGVR` and the managed-by label/value constants public; the test now uses them instead of duplicating.
- Reused (and extended) `findEndpointSlices` instead of the bespoke closures; `It("should …")` naming; rewrote the assertions with the `Eventually(func(g Gomega, …))` pattern and a 1s `Consistently` timeout.

The design doc moved out to a SEP as you asked: **submariner-io/enhancements#270**.

Locally: `go build`, `go vet`, `make golangci-lint` (0 issues) and the full `pkg/agent/controller` suite all pass. Could you re-approve the workflows when you get a chance? 🙏

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,277 @@
...
+	if serviceName == "" || serviceNamespace == "" || sourceCluster == "" || sourceCluster == c.clusterID {
+		return nil, false
+	}
+
+	reflectedName := serviceName + "-" + sourceCluster
+
+	// The reflected name must be a valid Service name (DNS-1123 label, <=63 chars).
+	// An over-long service/cluster combination can't be reflected; skip and log
+	// rather than failing on every reconcile.
+	if errs := validation.IsDNS1123Label(reflectedName); len(errs) > 0 {
```
tpantelis, line 205: To (virtually) avoid this possibility altogether, you could follow the pattern used in _service_endpoint_slices.go_ which uses the `GenerateName` functionality. This also would eliminate the possibility for collisions. Instead of using `c.federator.Distribute` to create/update the service, you can use `util.CreateOrUpdateWithOptions` directly with `IdentifyingLabels`.  This returns the created/updated obj so you can obtain the name for the `EndpointSlice`.

You can also use `GenerateName` for the `EndpointSlice` and a federator with `IdentifyingLabels` as is done in _service_endpoint_slices.go_.

kekkker: Good call on GenerateName + IdentifyingLabels. I'll adopt that for the EndpointSlice, which has no external name contract.

One constraint on the Service side though. Its name is the cluster.local DNS name consumers resolve, `<service>-<sourceCluster>.<ns>.svc.cluster.local`, and that predictable name is the whole point of the feature. With GenerateName it would become `<service>-<sourceCluster>-xxxxx`, which workloads can't resolve, so I think the Service name has to stay deterministic.

Proposed split:
* Service: keep the deterministic `<service>-<sourceCluster>` name, but create/update it via `util.CreateOrUpdateWithOptions` with `IdentifyingLabels` (drops the manual GET based collision check). Keep the DNS-1123 label guard so an over-long name gets skipped and logged rather than erroring on every reconcile.
* EndpointSlice: GenerateName plus a federator with `IdentifyingLabels`, named off the created Service, exactly like `service_endpoint_slices.go`.

Does that work for you? If you'd rather drop the length guard entirely, the other option is to hash or bound the Service name, but that loses the predictable DNS name, so I lean toward keeping it deterministic and guarded. Happy to go either way.

tpantelis: Your proposal sounds good.

kekkker: Done in the latest push (force-pushed to keep it a single commit). EndpointSlice now uses GenerateName with an IdentifyingLabels federator; the Service keeps its deterministic name and DNS-1123 guard. `make golangci-lint` and the unit/integration tests pass.

pkg/agent/controller/clusterlocal_reflector_test.go
```diff
@@ -0,0 +1,139 @@
...
+			t.cluster1.deleteServiceExport(ctx)
+			t.cluster1.deleteEndpointSlice(ctx, t.cluster1.serviceEndpointSlices[0].Name)
+
+			By("Verifying the reflected Service and EndpointSlice are removed")
+
+			Eventually(func(g Gomega, ctx context.Context) {
+				_, err := svcClient.Get(ctx, reflectedName, metav1.GetOptions{})
+				g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+				g.Expect(findEndpointSlices(ctx, t.cluster2.localEndpointSliceClient, "", reflectedName, "")).To(BeEmpty())
+			}).WithContext(ctx).Should(Succeed())
```
tpantelis: If an assertion persistently fails, `Eventually` will hang waiting for the `ctx` to be cancelled. When `WithContext` is used, to avoid this, you either have to explicitly specify a timeout via `Within` or call `EnforceDefaultTimeoutsWhenUsingContexts()` in `BeforeSuite` so it uses the default timeout. Same on line 126.

```suggestion
			}).Within(5 * time.Second)
.WithContext(ctx).Should(Succeed())
```

tpantelis (changes requested):

pkg/agent/controller/clusterlocal_reflector_test.go
kekkker: Done, added `Within(5 * time.Second)` to both (matches the existing usage elsewhere in the suite). Force-pushed to keep it one commit.

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		logger.Errorf(err, "Error checking reflected Service %s/%s", serviceNamespace, reflectedName)
+		return nil, true
+	}
+
+	if !owned {
+		logger.Warningf("Service %s/%s already exists and is not reflector-owned; skipping reflection of imported "+
+			"service %q from cluster %q", serviceNamespace, reflectedName, serviceName, sourceCluster)
+
+		return nil, false
```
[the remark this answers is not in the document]
tpantelis: The suggested fix for when a service is replaced or edited into non-reflector-owned is incomplete - it would need to delete all reflected `EndpointSlices`. 

Also, the controller only watches `EndpointSlices` so if a user did edit a reflected service, reconciliation would not be triggered and the reflected `EndpointSlices` would remain indefinitely, unless all originating `EndpointSlices` eventually happen to be updated.

To address this, we'd have to add a Service watcher to detect when reflected Services are modified/deleted and ensure all reflected `EndpointSlices` are deleted (use admiral's `watcher.New` and override `ResourcesEquivalent` to check if `old` was reflected).   @kekkker

pkg/agent/controller/clusterlocal_reflector.go
```diff
@@ -0,0 +1,277 @@
...
+func servicePortsFrom(in []discovery.EndpointPort) []corev1.ServicePort {
+	ports := make([]corev1.ServicePort, 0, len(in))
+
+	for i := range in {
+		p := &in[i]
+		ports = append(ports, corev1.ServicePort{
+			Protocol:   ptr.Deref(p.Protocol, corev1.ProtocolTCP),
+			Name:       ptr.Deref(p.Name, ""),
+			Port:       ptr.Deref(p.Port, 0),
+			TargetPort: intstr.FromInt32(ptr.Deref(p.Port, 0)),
```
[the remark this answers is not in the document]
tpantelis: @kekkker  ☝️

pkg/agent/controller/clusterlocal_reflector.go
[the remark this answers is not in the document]
kekkker: Added a Service watcher (admiral `watcher.New` over reflector-owned Services). When a reflected Service is deleted, or edited so it no longer carries the managed-by label (which drops it from the watch), `OnDelete` deletes the orphaned reflected EndpointSlices via `DeleteCollection`. Since the label selector turns the edited-out-of-ownership case into a delete event, I didn't need ResourcesEquivalent. Added an integration spec for it. Force-pushed to keep one commit.

kekkker:
Heads up: I rebased onto the latest `devel` (one feature commit) rather than carrying the merge commit, to keep it a single commit as you'd asked. So my push superseded the `Merge branch 'devel'` commit, but the branch is now current with `devel` all the same. Also pushed the Service-watcher change from the thread above. `make golangci-lint` + unit/integration tests pass locally.

pkg/agent/controller/clusterlocal_reflector.go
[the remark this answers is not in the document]
tpantelis: @kekkker Please address this comment (either implement the suggested fix or, if not, add a comment why).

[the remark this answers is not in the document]
kekkker: Done — `servicePortsFrom` now sets `AppProtocol: p.AppProtocol` so the reflected Service ports preserve it.

tpantelis:
@coderabbitai approve

tpantelis (approved):

PR #2278, opened 2026-09-09
Add namespace allow list configuration

tpantelis:
Adds `import-namespace-allow-list` configuration to allow specific namespaces that would otherwise be denied by the deny list. The allow list acts as an exception mechanism, overriding deny list matches.

With default deny list (_kube-_, _openshift-_, _openshift_) and new default allow list (_openshift-storage_), _openshift-storage_ is now permitted by default while other openshift-* namespaces remain denied.

Both exact match and prefix match (ending with -) are supported in the allow list, consistent with deny list behavior.

## Summary by CodeRabbit

* **New Features**
  * Added configurable namespace allowlists alongside denylists.
  * Allowlist entries can override matching denylist rules, including supported prefix matches.
  * Added `openshift-storage` as a permitted default namespace.
* **Bug Fixes**
  * Improved configuration parsing by trimming whitespace and ignoring empty entries.
  * Preserved validation for empty namespaces and denied imports.

### DIFF
--- a/pkg/agent/controller/namespace_validator.go
+++ b/pkg/agent/controller/namespace_validator.go
@@ -26,33 +26,29 @@ import (
 )
 
 const (
-	ConfigKeyImportNamespaceDenyList = "import-namespace-deny-list"
-	DefaultImportNamespaceDenyList   = "kube-,openshift-,openshift"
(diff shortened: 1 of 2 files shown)


tpantelis:
@coderabbitai approve

vthapar (approved):

